PR #5, opened 2024-04-29
database setup

daniel-burghardt:
### What

Adds the `db` package for handling database connections and migrations.

### Why

Core component of the application.

### Known limitations

N/A

N/A

#### PR Structure

#### Thoroughness

#### Release


internal/data/models.go
```diff
@@ -0,0 +1,17 @@
...
+
+type Models struct {
+}
+
+func NewModels(db db.DBConnectionPool) (*Models, error) {
+	if db == nil {
+		return nil, errors.New("DBConnectionPool must be initialized")
+	}
+
+	return &Models{}, nil
```
daniel-burghardt, line 16: See https://github.com/stellar/wallet-backend/pull/6/files for intended usage.

howardtw:
👀

internal/db/db.go
```diff
@@ -0,0 +1,48 @@
...
+type DBConnectionPool interface {
+	Close() error
+	Ping() error
+	DriverName() string
+	sqlx.ExecerContext
+	sqlx.QueryerContext
+	sqlx.PreparerContext
+	GetContext(ctx context.Context, dest interface{}, query string, args ...interface{}) error
+	SelectContext(ctx context.Context, dest interface{}, query string, args ...interface{}) error
+}
```
howardtw, line 20: ❓ Is it not clear to me why we need to create an interface here? Can you explain a little bit more? Is it for testing purposes? Do we plan on swapping to another implementation in the future?

daniel-burghardt: The idea with the interface is that it would make it very trivial to add query metrics in the future.
As discussed on standup, we will look into what instruments AWS RDS provides in that respect to make a decision about keeping the interface vs. using any existing RDS tools.

daniel-burghardt: https://cheesecake.slack.com/archives/CJ55HPNR4/p1714421061072399

howardtw: Keep in mind that this is a public repo. People wouldn't be able to see what that link is about.

What would be the thing that you'd like to track on top of query execution time?

Another thing to consider is whether it will be easy for people to configure their own query metric-tracking system.

internal/data/models.go
```diff
@@ -0,0 +1,17 @@
...
+	"errors"
+	"wallet-backend/internal/db"
+)
+
+type Models struct {
+}
+
+func NewModels(db db.DBConnectionPool) (*Models, error) {
+	if db == nil {
+		return nil, errors.New("DBConnectionPool must be initialized")
```
howardtw, line 13: Looks like `errors.New("DBConnectionPool must be initialized")` will be a common error for different models. Can we make it a global variable in the data package?

internal/serve/serve.go
```diff
@@ -4,22 +4,31 @@ import (
...
 	supportlog "github.com/stellar/go/support/log"
 	"github.com/stellar/go/support/render/health"
 )
 
 type Configs struct {
-	Logger *supportlog.Entry
-	Port   int
+	Logger       *supportlog.Entry
+	Port         int
+	DBConnection string
```
howardtw, line 19: ❓ what is it going to look like? A database URL?

internal/db/dbtest/dbtest.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/stellar/go/support/db/schema"
+)
+
+func Open(t *testing.T) *dbtest.DB {
+	db := dbtest.Postgres(t)
+	conn := db.Open()
+	defer conn.Close()
+
+	migrateDirection := schema.MigrateUp
+	m := migrate.HttpFileSystemMigrationSource{FileSystem: http.FS(migrations.FS)}
```
howardtw, line 19: ❓ Can you explain how we are going to use this? Is it not going to use files in a certain directory?

daniel-burghardt: It is, `migrations.FS` is a reference to the file system (directory) where the migrations live. It is defined in `migrations/main.go`. It becomes an absolute reference so that we no longer need to worry about providing relative paths whenever opening the DB connection. Even if the migrations folder moves, the reference won't be broken.

internal/serve/serve.go
daniel-burghardt: Yes, I've renamed it: [b8f9b6c](https://github.com/stellar/wallet-backend/pull/5/commits/b8f9b6c74e681f3efa3a5be00689cb9eea27db12)

internal/db/db.go
daniel-burghardt: I've removed the Slack thread link. No links should live in the open repo indeed, not so much for lack of access but for security reasons I would say.

I don't have a particular array of metrics I would like to implement off the bat, but the idea would be to keep the implementation flexible for whenever ideas and requirements for this arise, making it easier to integrate them without major refactors.

But query time (planning and execution) seems like a good first candidate. One thing the folks from the OPS team mentioned in the Slack thread was the possibility of hooking up alerts here as well, which seems like a great option to have.

internal/serve/serve.go
```diff
@@ -4,22 +4,31 @@ import (
...
 	supportlog "github.com/stellar/go/support/log"
 	"github.com/stellar/go/support/render/health"
 )
 
 type Configs struct {
-	Logger *supportlog.Entry
-	Port   int
+	Logger          *supportlog.Entry
+	Port            int
+	DBConnectionURL string
```
howardtw, line 19: Can we simply name this DatabaseURL? I don't see where we read the env and assign the value to this variable. Should we add it to cmd/serve.go?

daniel-burghardt: Added: [fc7fee0](https://github.com/stellar/wallet-backend/pull/5/commits/fc7fee097ada767a932d5ec638cad99d78a3ce49)

howardtw (approved):
LGTM

PR #6, opened 2024-04-29
payments: subscribe

daniel-burghardt:
### What

Adds a new `POST /payments/subscribe` endpoint.

### Why

So that a Proprietary wallet backend can call to register its user accounts for payments tracking.

### Known limitations

N/A

[[OSB Ingestion] Implement user management](https://app.asana.com/0/1202672669313222/1207149132580452/f)

#### PR Structure

#### Thoroughness

#### Release


howardtw:
👀

internal/data/models.go
```diff
@@ -6,12 +6,15 @@ import (
 )
 
 type Models struct {
+	Payments *PaymentModel
 }
 
 func NewModels(db db.DBConnectionPool) (*Models, error) {
```
howardtw, line 12: nit: just realized we might be able to rename DBConnectionPool to ConnectionPool since it's in the db package. Otherwise, we have three `db`s in one line. (nits are non-blocking)

internal/serve/httperror/errors.go
howardtw, line 1: I wonder whether we should be this package outside the `serve` directory. One of the ideas is to put it under `net`

internal/serve/httphandler/payments_handler.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"net/http"
+	"wallet-backend/internal/data"
+	"wallet-backend/internal/serve/httperror"
+
+	"github.com/stellar/go/support/http/httpdecode"
+)
+
+type PaymentsHandler struct {
+	*data.Models
```
howardtw, line 12: ❓ does this mean will be put all the data models in the payments handler including ones that it doesn't use?

internal/serve/httphandler/payments_handler.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func (h PaymentsHandler) SubscribeAddress(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	var reqBody PaymentsSubscribeRequest
+	if err := httpdecode.DecodeJSON(r, &reqBody); err != nil {
+		httperror.BadRequest("Invalid request body").Render(w)
+		return
+	}
```
howardtw, line 26: nit: it would be great if we can have a consistent error handling pattern in a single function
```suggestion
	err := httpdecode.DecodeJSON(r, &reqBody)
	if err != nil {
		httperror.BadRequest("Invalid request body").Render(w)
		return
	}
```

internal/serve/httphandler/payments_handler.go
howardtw: Can we just include `data.PaymentMode`l here?

internal/serve/serve.go
```diff
@@ -63,10 +63,21 @@ func getHandlerDeps(cfg Configs) (handlerDeps, error) {
...
+	mux.MethodNotAllowed(httperror.ErrorHandler{Error: httperror.MethodNotAllowed}.ServeHTTP)
 
 	mux.Get("/health", health.PassHandler{}.ServeHTTP)
 
+	// Authenticated routes
+	mux.Group(func(r chi.Router) {
+		// r.Use(...authMiddleware...)
+
+		r.Route("/payments", func(r chi.Router) {
+			handler := &httphandler.PaymentsHandler{}
```
howardtw, line 76: Do we not have to put models in the handler?

internal/serve/httperror/errors.go
daniel-burghardt: My idea is that the `serve` package would bundle everything API and HTTP related. Since `errors.go` only deals with HTTP errors, it makes sense to me that it would live there.

internal/serve/httphandler/payments_handler.go
daniel-burghardt: We could, but a handler or a service is likely to need access to different data models, so this reduces significantly the hassle to implement and maintain them.

Since models only contain a pointer to the DB, I don't see a problem with that, as it isn't resource wasteful.

internal/data/models.go
daniel-burghardt: Agreed: [fd204fa](https://github.com/stellar/wallet-backend/pull/6/commits/fd204fa7dcbdef92d6369ac6f84d376c2fb341ea)

internal/serve/serve.go
daniel-burghardt: We do indeed, good catch! 🚀

internal/serve/httphandler/payments_handler.go
howardtw: To me, it's less about wasting the resource. It's more about making it clear what the PaymentsHandler is going to need. For example, if we include many models that are not used by the PaymentsHandler, it could be confusing to some readers.

internal/serve/httperror/errors.go
howardtw: I'm trying to think of the potential of other services using those errors. Since I don't have one in mind, let's leave it like this and refactor if necessary.

howardtw (approved):
Approving. There's one unresolved comment that I feel like we can make the responsibility of each handler more clear.

internal/serve/httphandler/payments_handler.go
leighmcculloch: The dependencies do become harder to understand when everything has a link to all dependencies. And maybe be harder to write tests if we need to provide all models to payments tests that otherwise don't need to use them.

CaioTeixeira95 (approved):
LGTM

internal/serve/httphandler/payments_handler.go
daniel-burghardt: @leighmcculloch writing tests shouldn't get harder I believe, because they can be left `nil` if not necessary for the test.
But I don't have a strong preference here, although I do lean towards the all-in-one approach because of its easier maintainability imo.

daniel-burghardt: Is it your preference to have it like so? [82a2c97](https://github.com/stellar/wallet-backend/pull/6/commits/82a2c974ae9b0d356c2be3182304a3b739b21bb5)

leighmcculloch: I think 82a2c97 will be easier to understand which model a dependency uses, but I don't feel strongly and defer to you both.

internal/serve/httphandler/payments_handler.go
```diff
@@ -0,0 +1,34 @@
...
+
+	err = h.Models.Payments.SubscribeAddress(ctx, reqBody.Address)
+	if err != nil {
+		httperror.InternalServerError.Render(w)
+		return
+	}
+}
```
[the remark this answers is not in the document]
daniel-burghardt: As explained on call, it is the intention to let the default response (200 with no body) to be returned for now, but we could in the future agree on a response format to use across all requests (potentially including a `message` field).

internal/data/models.go
```diff
@@ -6,12 +6,15 @@ import (
 )
 
 type Models struct {
+	Payments *PaymentModel
 }
 
-func NewModels(db db.DBConnectionPool) (*Models, error) {
+func NewModels(db db.ConnectionPool) (*Models, error) {
 	if db == nil {
 		return nil, errors.New("DBConnectionPool must be initialized")
```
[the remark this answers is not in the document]
daniel-burghardt: [70dbfea](https://github.com/stellar/wallet-backend/pull/6/commits/70dbfeaebc0dafb7a1215cb709312bb5a57aa618)

internal/serve/httphandler/payments_handler.go
```diff
@@ -0,0 +1,34 @@
...
+
+	var reqBody PaymentsSubscribeRequest
+	err := httpdecode.DecodeJSON(r, &reqBody)
+	if err != nil {
+		httperror.BadRequest("Invalid request body").Render(w)
+		return
+	}
+
+	err = h.Models.Payments.SubscribeAddress(ctx, reqBody.Address)
+	if err != nil {
```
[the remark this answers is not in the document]
daniel-burghardt: I think that's a good idea! 
I was considering how we should implement Sentry though. I think it might be worth implementing an error tracking interface so that contributors could easily implement other providers.
I've created a [task](https://app.asana.com/0/1203779254475146/1207245825676510/f) for it and left a TODO in the code: [1c32a92](https://github.com/stellar/wallet-backend/pull/6/commits/1c32a922ff4286bf8126a351a7ac43e0d57b856e)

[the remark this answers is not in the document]
daniel-burghardt: Btw, I see that we are currently sharing Asana links over PR's (it's even in the template). Should we stop doing so and remove it from the template?
cc: @howardtw @mtwtfss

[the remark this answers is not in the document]
daniel-burghardt: Or move the all wallet-backend tasks to Github issues?

PR #7, opened 2024-05-06
payments: unsubscribe

daniel-burghardt:
### What

Adds a new `POST /payments/unsubscribe` endpoint.

### Why

So that a Proprietary wallet backend can call to unregister its user accounts for payments tracking.

### Known limitations

N/A

[[OSB Ingestion] Implement user management](https://app.asana.com/0/1202672669313222/1207149132580452/f)

#### PR Structure

#### Thoroughness

#### Release


internal/serve/httphandler/payments_handler.go
```diff
@@ -32,3 +32,20 @@ func (h PaymentsHandler) SubscribeAddress(w http.ResponseWriter, r *http.Request
...
+
+	var reqBody PaymentsSubscribeRequest
+	err := httpdecode.DecodeJSON(r, &reqBody)
+	if err != nil {
+		httperror.BadRequest("Invalid request body").Render(w)
+		return
+	}
+
+	err = h.Models.Payments.UnsubscribeAddress(ctx, reqBody.Address)
+	if err != nil {
```
CaioTeixeira95, line 47: Can we log and send a report to the sentry when a 5xx happens?

internal/serve/httphandler/payments_handler_test.go
```diff
@@ -42,9 +42,9 @@ func TestSubscribeAddress(t *testing.T) {
 		require.NoError(t, err)
 	}
 
-	t.Run("success", func(t *testing.T) {
+	t.Run("success_happy_path", func(t *testing.T) {
```
CaioTeixeira95, line 45: Just out of curiosity, why are we naming with camel case the test name?

CaioTeixeira95 (approved):
LGTM

internal/serve/httphandler/payments_handler_test.go
daniel-burghardt: I decided to start naming it with snake case as opposed to with spaces because it makes it easier to search for a failing test based on the `go test` output (which replaces whitespaces with `_`).
e.g.: ` --- PASS: TestUnsubscribeAddress/success_happy_path (0.00s)`

But I'm also open to suggestions here, do you think camelCase would be more appropriate?

internal/serve/httphandler/payments_handler.go
daniel-burghardt: Applied the same resolution as in the `/subscribe` PR: https://github.com/stellar/wallet-backend/pull/6#discussion_r1593921662

internal/serve/httphandler/payments_handler_test.go
CaioTeixeira95: Got it! Thank you, It may help find the test when it fails.

internal/serve/httphandler/payments_handler.go
CaioTeixeira95: Thank you.

PR #8, opened 2024-05-06
git actions: go workflow

daniel-burghardt:
### What

Adds the Github Workflow running three steps: a linter, a builder and the unit tests.

### Why

To ensure code quality.

### Known limitations

N/A

N/A

#### PR Structure

#### Thoroughness

#### Release


.github/workflows/go.yaml
```diff
@@ -0,0 +1,74 @@
+name: Go
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: ["**"]
+
+jobs:
+  check:
```
CaioTeixeira95, line 10: I'd recommend we add `Exhaustive` and `Deadcode` too. `Exhaustive` check if the code verifies every possible way on if and switch statements. `Deadcode` checks for code that is never used.

https://github.com/stellar/stellar-disbursement-platform-backend/blob/develop/.github/workflows/ci.yml#L36-L56

CaioTeixeira95 (approved):
Thank you for this! 🚀

.github/workflows/go.yaml
daniel-burghardt: Good idea! Will do. 👍

PR #12, opened 2024-05-17
all: use fmt for error wrapping

CaioTeixeira95:
### What

Use the `fmt` package for error wrapping.

### Why

The third-party errors package is unnecessary because we can now use the standard library.

### Known limitations

N/A

#### PR Structure

#### Thoroughness

#### Release


daniel-burghardt (approved):
Thank you! Minus one dependency. :rocket:

PR #13, opened 2024-05-21
auth: add signature middleware

CaioTeixeira95:
### What

### Why

### Known limitations

[TODO or N/A]

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/cmd/utils/custom_set_value.go
+++ b/cmd/utils/custom_set_value.go
@@ -1,0 +1,26 @@
+package utils
+
+import (
+	"fmt"
+
+	"github.com/spf13/viper"
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/support/config"
+)
+
+func SetConfigOptionStellarPublicKey(co *config.ConfigOption) error {
+	publicKey := viper.GetString(co.Name)
+
+	kp, err := keypair.ParseAddress(publicKey)
+	if err != nil {
+		return fmt.Errorf("error validating public key in %s: %w", co.Name, err)
+	}
+
+	key, ok := co.ConfigKey.(*string)
+	if !ok {
+		return fmt.Errorf("the expected type for the config key in %s is a string, but a %T was provided instead", co.Name, co.ConfigKey)
+	}
+	*key = kp.Address()
+
+	return nil
+}
--- a/internal/serve/auth/signature_verifier.go
+++ b/internal/serve/auth/signature_verifier.go
@@ -1,0 +1,157 @@
+package auth
+
+import (
+	"context"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"net/url"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/strkey"
+	"github.com/stellar/go/support/log"
+)
+
+type SignatureVerifier interface {
+	VerifySignature(ctx context.Context, signatureHeaderContent string, rawReqBody []byte) error
+}
+
+var (
+	ErrStellarSignatureNotVerified = errors.New("neither Signature nor X-Stellar-Signature header could be verified")
+)
+
+type ErrInvalidTimestampFormat struct {
+	TimestampString string
+}
+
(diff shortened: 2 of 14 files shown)


internal/serve/auth/signature_verifier.go
```diff
@@ -0,0 +1,157 @@
...
+func ExtractTimestampedSignature(signatureHeaderContent string) (t string, s string, err error) {
+	parts := strings.SplitN(signatureHeaderContent, ",", 2)
+	if len(parts) != 2 {
+		return "", "", fmt.Errorf("malformed header format: %s", signatureHeaderContent)
+	}
+
+	tHeaderContent := parts[0]
+	timestampParts := strings.SplitN(tHeaderContent, "=", 2)
+	if len(timestampParts) != 2 || strings.TrimSpace(timestampParts[0]) != "t" {
+		return "", "", fmt.Errorf("malformed timestamp format: %s", timestampParts)
```
daniel-burghardt, line 106: Should we use `ErrInvalidTimestampFormat` here too? It is checked for in `VerifySignature` but it isn't returned in this function currently.

internal/serve/auth/signature_verifier.go
```diff
@@ -0,0 +1,157 @@
...
+		return ErrStellarSignatureNotVerified
+	}
+
+	return nil
+}
+
+func ExtractTimestampedSignature(signatureHeaderContent string) (t string, s string, err error) {
+	parts := strings.SplitN(signatureHeaderContent, ",", 2)
+	if len(parts) != 2 {
+		return "", "", fmt.Errorf("malformed header format: %s", signatureHeaderContent)
```
daniel-burghardt, line 100: Very nit-picky but I think we could lose the "format" at the end, since "malformed" already gives away that the format is wrong.
```suggestion
		return "", "", fmt.Errorf("malformed header: %s", signatureHeaderContent)
```

internal/serve/auth/signature_verifier.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+var _ SignatureVerifier = (*StellarSignatureVerifier)(nil)
+
+// VerifySignature verifies the Signature or X-Stellar-Signature content and checks if the signature is signed for a known caller.
+func (sv *StellarSignatureVerifier) VerifySignature(ctx context.Context, signatureHeaderContent string, rawReqBody []byte) error {
+	t, s, err := ExtractTimestampedSignature(signatureHeaderContent)
+	if err != nil {
+		log.Ctx(ctx).Error(err)
+		var errTimestampFormat *ErrInvalidTimestampFormat
```
daniel-burghardt, line 56: Currently we would only return an specific error if the problem is with the `t` value, but not with `s` or the overall formatting of the header.
Would it make sense to standardize it for all? E.g.: either return `ErrStellarSignatureNotVerified` in all cases or have a custom error for each.

internal/serve/auth/signature_verifier.go
```diff
@@ -0,0 +1,157 @@
...
+		if errors.As(err, &errTimestampFormat) {
+			return err
+		}
+		return ErrStellarSignatureNotVerified
+	}
+
+	// 2 minutes in seconds
+	err = VerifyGracePeriodSeconds(t, 2*time.Second)
```
daniel-burghardt, line 64: The comment says 2 minutes but it being set as 2s. Which of the two do we want to set it to?

cmd/utils/custom_set_value.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/support/config"
+)
+
+func SetConfigOptionStellarPublicKey(co *config.ConfigOption) error {
+	publicKey := viper.GetString(co.Name)
+
+	kp, err := keypair.ParseAddress(publicKey)
+	if err != nil {
+		return fmt.Errorf("error validating public key in %s: %w", co.Name, err)
```
daniel-burghardt, line 16:
```suggestion
		return fmt.Errorf("validating public key in %s: %w", co.Name, err)
```

daniel-burghardt (approved):
Very minor comments, but overall LGTM! 🚀

PR #14, opened 2024-05-22
Ingest: service created

daniel-burghardt:
### What

### Why

### Known limitations

[TODO or N/A]

#### PR Structure

#### Thoroughness

#### Release


.gitignore
```diff
@@ -1,2 +1,3 @@
 # Visual Studio Code
-.vscode
\ No newline at end of file
+.vscode
+captive-core*/
```
CaioTeixeira95, line 3: Is this name correct?

cmd/ingest.go
```diff
@@ -0,0 +1,101 @@
...
+			ConfigKey:   &cfg.NetworkPassphrase,
+			FlagDefault: network.TestNetworkPassphrase,
+			Required:    true,
+		},
+		{
+			Name:        "captive-core-bin-path",
+			Usage:       "Path to Captive Core's binary file.",
+			OptType:     types.String,
+			ConfigKey:   &cfg.CaptiveCoreBinPath,
+			FlagDefault: "/usr/local/bin/stellar-core",
```
CaioTeixeira95, line 40: I wonder if this one couldn't also point to a local directory like the `captive-core-config-dir`.

cmd/ingest.go
```diff
@@ -0,0 +1,101 @@
...
+			OptType:     types.String,
+			ConfigKey:   &cfg.NetworkPassphrase,
+			FlagDefault: network.TestNetworkPassphrase,
+			Required:    true,
+		},
+		{
+			Name:        "captive-core-bin-path",
+			Usage:       "Path to Captive Core's binary file.",
+			OptType:     types.String,
+			ConfigKey:   &cfg.CaptiveCoreBinPath,
```
CaioTeixeira95, line 39: I was thinking about adding a `CustomSetValue` for it since it's a folder, we could add validations to see if the folder exists before the application starts.

cmd/ingest.go
```diff
@@ -0,0 +1,101 @@
...
+			FlagDefault: "/usr/local/bin/stellar-core",
+			Required:    true,
+		},
+		{
+			Name:        "captive-core-config-dir",
+			Usage:       "Path to Captive Core's configuration files directory.",
+			OptType:     types.String,
+			ConfigKey:   &cfg.CaptiveCoreConfigDir,
```
CaioTeixeira95, line 47: Same here (custom set value for folders value).

cmd/ingest.go
```diff
@@ -0,0 +1,101 @@
...
+			FlagDefault: "./internal/ingest/config",
+			Required:    true,
+		},
+		{
+			Name:        "ledger-cursor-name",
+			Usage:       "Name of last synced ledger cursor. Attention: there should never be more than one container running with a same cursor name.",
```
CaioTeixeira95, line 53: Could you elaborate on it? Is this value an enum or should it be a unique value? We should make it more clear and, also remove the `FlagDefault` since it cannot have the same name.

cmd/root.go
```diff
@@ -30,8 +28,9 @@ func Execute() {
 }
 
 func init() {
-	logger := supportlog.New()
-	logger.SetLevel(logrus.TraceLevel)
+	log.DefaultLogger = log.New()
+	log.DefaultLogger.SetLevel(logrus.TraceLevel)
```
CaioTeixeira95, line 32: We would like to have fewer logs (debug) in production, perhaps we can move this to an option when running the service, then we pass the `LogLevel`.

internal/data/payments.go
```diff
@@ -2,15 +2,107 @@ package data
...
+	To              string
+	SrcAssetCode    string
+	SrcAssetIssuer  string
+	SrcAmount       int64
+	DestAssetCode   string
+	DestAssetIssuer string
+	DestAmount      int64
+	CreatedAt       time.Time
+	Memo            string
+}
```
CaioTeixeira95, line 33: We would like to add the flags to the database table columns.

```
type Payment struct {
	OperationID     int64 `db:"operation_id"`
	...
}
```

internal/data/payments.go
```diff
@@ -2,15 +2,107 @@ package data
...
+
+func (m *PaymentModel) GetLatestLedgerSynced(ctx context.Context, cursorName string) (uint32, error) {
+	var lastSyncedLedger uint32
+	err := m.db.QueryRowxContext(ctx, `SELECT value FROM ingest_store WHERE key = $1`, cursorName).Scan(&lastSyncedLedger)
+	// First run, key does not exist yet
+	if err == sql.ErrNoRows {
+		return 0, nil
+	}
+	if err != nil {
+		return 0, err
```
CaioTeixeira95, line 43:
```suggestion
		return 0, fmt.Errorf("getting latest ledger synced for cursor %s: %w", cursorName, err)
```

internal/data/payments.go
```diff
@@ -2,15 +2,107 @@ package data
...
+}
+
+func (m *PaymentModel) BeginTx(ctx context.Context) (*sqlx.Tx, error) {
+	tx, err := m.db.BeginTxx(ctx, nil)
+	if err != nil {
+		return nil, fmt.Errorf("beginning transaction: %w", err)
+	}
+
+	return tx, nil
+}
```
CaioTeixeira95, line 69: Why do we need this function?

internal/data/payments.go
CaioTeixeira95: Is any field nullable?

internal/data/payments.go
```diff
@@ -2,15 +2,107 @@ package data
...
+
+func (m *PaymentModel) AddPayment(ctx context.Context, tx *sqlx.Tx, payment Payment) error {
```
CaioTeixeira95, line 71: Perhaps we want to use the same strategy we used on the SDP when working with database transactions.

- [SQLExecuter](https://github.com/stellar/stellar-disbursement-platform-backend/blob/develop/db/db.go#L160-L168) that is an interface.
- [A function that runs your query inside a database transaction](https://github.com/stellar/stellar-disbursement-platform-backend/blob/develop/db/db.go#L136-L144)
- [Your Model receives the `SQLExecutor` as a parameter](https://github.com/stellar/stellar-disbursement-platform-backend/blob/develop/internal/data/assets.go#L154).

internal/data/payments.go
CaioTeixeira95: Please, read my comment about the database transaction. We don't need to do everything on this PR, but I believe that's a better approach when handling database transactions.

internal/data/payments.go
CaioTeixeira95, line 71: It would be great if this function returns the entire object of the payment inserted in the database or at least the ID generated if it has.

internal/db/migrations/2024-05-22.0-ingest_payments.sql
```diff
@@ -0,0 +1,34 @@
...
+CREATE TABLE ingest_store (
+  key varchar(255) NOT NULL,
+  value varchar(255) NOT NULL,
+  PRIMARY KEY (key)
+);
+
+CREATE TABLE ingest_payments (
+  operation_id bigint NOT NULL,
+  operation_type text NOT NULL,
+  transaction_id bigint NOT NULL,
```
CaioTeixeira95, line 12: I'd name this field `stellar_transaction_id` to make it clear that it is from the network. Only a suggestion.

internal/db/migrations/2024-05-22.0-ingest_payments.sql
```diff
@@ -0,0 +1,34 @@
...
+  transaction_hash text NOT NULL,
```
CaioTeixeira95, line 13: same here

internal/data/payments.go
CaioTeixeira95: The `Memo` field should be a pointer.

internal/ingest/ingest.go
```diff
@@ -0,0 +1,102 @@
...
+
+	return &services.IngestManager{
+		NetworkPassphrase: cfg.NetworkPassphrase,
+		LedgerCursorName:  cfg.LedgerCursorName,
+		LedgerBackend:     ledgerBackend,
+		PaymentModel:      models.Payments,
+	}, nil
+}
+
+func getCaptiveCoreConfig(cfg Configs) (ledgerbackend.CaptiveCoreConfig, error) {
```
CaioTeixeira95, line 69: I'm missing tests for these functions, I think at least we should validate if the values are being passed correctly.

internal/utils/ingestion_utils.go
CaioTeixeira95, line 1: Can we add tests for these functions?

internal/utils/utils.go
```diff
@@ -0,0 +1,14 @@
+package utils
+
+import (
+	"bytes"
+	"strings"
+)
+
+// SanitizeUTF8 sanitizes a string to comply to the UTF-8 character set and Postgres' code zero byte constraint
+func SanitizeUTF8(input string) string {
```
CaioTeixeira95, line 9: Can we add a test to validate this function?

internal/services/ingest.go
```diff
@@ -0,0 +1,231 @@
...
+
+	err = dbTx.Commit()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func fillPayment(payment *data.Payment, operation xdr.OperationBody) {
```
CaioTeixeira95, line 198: can we test these functions to check if all values are being filled with the corrected values?

CaioTeixeira95 (commented):
Thank you for the hard work! I've left some comments.

.gitignore
daniel-burghardt: Yes, whenever you run captive core it creates a folder named like `captive-core-9f8g7a9f` to store the ledger files it downloads in the application's root folder. The "*" is to capture the random name given to the folder.

cmd/ingest.go
daniel-burghardt: Whenever you install captive core locally using [these instructions](https://github.com/stellar/stellar-core/blob/master/INSTALL.md#macos), `/usr/local/bin/stellar-core` will be the default install location. If we wanted it to live inside a project directory we would need to change that install path.
I thought it was easier to use the default location and set it here accordingly. What do you think?

internal/data/payments.go
daniel-burghardt: Would you like to log anything with the resulting object or else why do you think this would be beneficial?

cmd/ingest.go
daniel-burghardt: Good one! [c81a985](https://github.com/stellar/wallet-backend/pull/14/commits/c81a98518778990bedbd71a3f5afa01968b66715)

cmd/ingest.go
daniel-burghardt: Good one! [c81a985](https://github.com/stellar/wallet-backend/pull/14/commits/c81a98518778990bedbd71a3f5afa01968b66715)

cmd/ingest.go
daniel-burghardt: It's not an enum, it's just an unique string value used as the key in the `ingest_store` dictionary-like table.

daniel-burghardt: I've made the description more clear, made it required and removed the default value. Good call, this will make people more aware that this has to be carefully configured when running parallel ingestion.

daniel-burghardt: [72a0dbb](https://github.com/stellar/wallet-backend/pull/14/commits/72a0dbbeeffb8a4c2d127edc74476bd9c69fdf16)

cmd/root.go
daniel-burghardt: [84725af](https://github.com/stellar/wallet-backend/pull/14/commits/84725affc193651ede99e41b49e041b70f9c3e90)

internal/data/payments.go
daniel-burghardt: We used because the service didn't have the DB directly injected into it, but I've now made the `DB` in the model exported. Also, removed this function and implemented the DB transaction wrapper function as you suggested in the other comment.

internal/data/payments.go
daniel-burghardt: [5735693](https://github.com/stellar/wallet-backend/pull/14/commits/573569389d9bccf3c6b45767ea1611288857f555)

internal/ingest/ingest.go
daniel-burghardt: 👍 [65edfa1](https://github.com/stellar/wallet-backend/pull/14/commits/65edfa1990c3966e6271b831f3738c4a2445dadd)

internal/services/ingest.go
daniel-burghardt: Added a test to the `processLedger` function, which calls this method internally: [ac1f46e](https://github.com/stellar/wallet-backend/pull/14/commits/ac1f46ea963cc6657b6329ff0fbc4f4d32ee8d26)

CaioTeixeira95 (approved):
LGTM 🚀

PR #15, opened 2024-06-05
validators: add validators

CaioTeixeira95:
### What

This PR adds the validation layer to the project using the [go-validator](https://github.com/go-playground/validator) library.

### Why

To make the request/object validation easier.

### Known limitations

N/A

N/A

#### PR Structure

#### Thoroughness

#### Release


internal/serve/httperror/errors.go
```diff
@@ -8,61 +8,65 @@ import (
...
-	return errorResponse{
+	return ErrorResponse{
 		Status: http.StatusUnauthorized,
 		Error:  message,
+		Extras: extras,
 	}
 }
 
-func InternalServerError(ctx context.Context, message string, err error) errorResponse {
+func InternalServerError(ctx context.Context, message string, err error, extras map[string]interface{}) *ErrorResponse {
```
marcelosalloum, line 63: Why some responses are a *ErrorResponse pointer and others are a ErrorResponse value? This should be consistent throughout.

internal/serve/httperror/errors.go
```diff
@@ -8,61 +8,65 @@ import (
...
 }
 
 type ErrorHandler struct {
-	Error errorResponse
+	Error ErrorResponse
 }
 
 func (h ErrorHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	h.Error.Render(w)
 }
```
marcelosalloum, line 27: Is the error handler necessary? I don't know the rest of the code, so I'm trying understand if this abstraction is needed.

internal/validators/validate.go
```diff
@@ -0,0 +1,80 @@
...
+	"strings"
+	"unicode"
+
+	"github.com/go-playground/validator/v10"
+	"github.com/stellar/go/strkey"
+)
+
+func NewValidator() *validator.Validate {
+	validate := validator.New()
+	_ = validate.RegisterValidation("public_key", publicKeyValidation)
```
marcelosalloum, line 15: RegisterValidation returns an error, right? You should handle it accordingly here by wrapping it and sending the wrapped version to the caller.

Or at least log the error here.

internal/validators/validate.go
```diff
@@ -0,0 +1,80 @@
...
+	case "oneof":
+		params := strings.Join(strings.Split(fieldError.Param(), " "), ", ")
+		return fmt.Sprintf("Unexpected value %q. Expected one of the following values: %s", fieldError.Value(), params)
+	case "gt":
+		if fieldError.Kind() == reflect.Slice || fieldError.Kind() == reflect.Array {
+			return "Should have at least 1 element"
+		}
+		return fmt.Sprintf("Should be greater than %s", fieldError.Param())
+	case "gte":
+		return fmt.Sprintf("Should be greater than or equal %s", fieldError.Param())
```
marcelosalloum, line 51: I didn't know gt could be used with slices and arrays. In any case, the get and gt messages re not consistent in case of arrays.

marcelosalloum: Also, is `Should have at least 1 element` what you expect to return when using gt with arrays? Isn't `fieldError.Param()` relevant for that case?

internal/validators/validate.go
```diff
@@ -0,0 +1,80 @@
...
+	length := len(namespace)
+	if length == 2 {
+		return lcFirst(namespace[1])
+	}
+
+	if length > 2 {
+		return fmt.Sprintf("%s.%s", lcFirst(namespace[length-2]), lcFirst(namespace[length-1]))
+	}
+
+	return lcFirst(namespace[0])
```
marcelosalloum, line 69: 💡  The logic for len==1 or len==2 is the same, right?

```suggestion
	namespace := strings.Split(fieldError.StructNamespace(), ".")
	length := len(namespace)
	if length <= 2 {
		return lcFirst(namespace[length-1])
	}
	return fmt.Sprintf("%s.%s", lcFirst(namespace[length-2]), lcFirst(namespace[length-1]))
```

internal/validators/validate.go
```diff
@@ -0,0 +1,80 @@
...
+func getFieldName(fieldError validator.FieldError) string {
+	// Ex.: structName.FieldName, structName.nestedStructName.nestedStructFieldName, structName.nestedStructName.nestedStructName....
+	namespace := strings.Split(fieldError.StructNamespace(), ".")
+	length := len(namespace)
+	if length == 2 {
+		return lcFirst(namespace[1])
+	}
+
+	if length > 2 {
+		return fmt.Sprintf("%s.%s", lcFirst(namespace[length-2]), lcFirst(namespace[length-1]))
```
marcelosalloum, line 66: Maybe this format would make things easer to debug, WDYT?

```suggestion
		return fmt.Sprintf("%s...%s.%s", lcFirst(namespace[0]), lcFirst(namespace[length-2]), lcFirst(namespace[length-1]))
```

internal/validators/validate.go
```diff
@@ -0,0 +1,80 @@
...
+	return strkey.IsValidEd25519PublicKey(addr) || strkey.IsValidMuxedAccountEd25519PublicKey(addr)
+}
+
+func ParseValidationError(errors validator.ValidationErrors) map[string]interface{} {
+	fieldErrors := make(map[string]interface{})
+	for _, err := range errors {
+		fieldErrors[getFieldName(err)] = msgForFieldError(err)
+	}
+	return fieldErrors
+}
```
marcelosalloum, line 31: There will be cases for when more than one validation fails, like in `validate:"required,public_key"`.

🤔 Do you think it's worth returning a slice of error messages, for each validation that fails in that field? Of would this library only return one error per field?

internal/serve/httphandler/payments_handler.go
```diff
@@ -22,13 +22,19 @@ func (h PaymentsHandler) SubscribeAddress(w http.ResponseWriter, r *http.Request
...
-		httperror.BadRequest("Invalid request body").Render(w)
+		httperror.BadRequest("Invalid request body", nil).Render(w)
+		return
+	}
+
+	httpErr := ValidateRequestBody(ctx, reqBody)
+	if httpErr != nil {
+		httpErr.Render(w)
 		return
 	}
```
marcelosalloum, line 33: 💡  I wonder if we could add a method that applies this same logic consistently every time. Something like 

```go
func DecodeJSONAndValidate(r *http.Request, ptr interface{}) httperror.Response {
	err := httpdecode.DecodeJSON(r, ptr)  // ptr is the same as &reqBody
	if err != nil {
		return httperror.BadRequest("Invalid request body", nil)
	}

	return ValidateRequestBody(ctx, reqBody)
}
```

marcelosalloum (approved):
LGTM!

I've left a few suggestions, but overall I'm impressed with the utility of this lib, great finding!

internal/serve/httperror/errors.go
CaioTeixeira95: Thank you for the question! It's used to override the default messages for some HTTP statuses. Please refer to this:

https://github.com/stellar/wallet-backend/blob/main/internal/serve/serve.go#L79-L80

internal/validators/validate.go
CaioTeixeira95: It only returns an error if the `key` is empty or the `function` is `nil`. That's why I'm not handling the error here, I don't think we should worry about it, wdyt?

internal/validators/validate.go
CaioTeixeira95: Senior style 😎

internal/validators/validate.go
CaioTeixeira95: It only returns one error per field, the order of the flags decides the validation order.

internal/validators/validate.go
CaioTeixeira95: The goal of this function is to get the field name that raised the validation error. If we return the `namespace[0]` we'll leak the structure name that we are using.

Here's an usage example:

```go
type User struct {
        Name string `json:"name" validate:"required"`
}
```

It will return something like:

```
{
     ...
     "extras": {
          "name": "This field is required"
     }
}
```

For nested fields:

```go
type Address struct {
      StreetName string `json:"streetName" validate:"required"`
}

type User struct {
        Name string `json:"name" validate:"required"`
        Address Address `json:"address"`
}
```

It will return something like:

```
{
     ...
     "extras": {
          "name": "This field is required",
          "address.streetName": "This field is required"
     }
}
```

internal/serve/httphandler/payments_handler.go
CaioTeixeira95: Awesome!

internal/validators/validate.go
CaioTeixeira95: Thank you for this! I'll change it so both will have the same behavior for slices. Also, I'm going to use the param here.

internal/validators/validate.go
marwen-abid: It's fine to do incremental error handling in an API

internal/validators/validate_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+func TestParseValidationError(t *testing.T) {
+	type testStructNested struct {
+		NestedRequiredField string `validate:"required"`
+		NestedEnumField     string `validate:"oneof=foo bar"`
+	}
+
+	type testStruct struct {
+		RequiredField      string             `validate:"required"`
+		RequiredArrayField []string           `validate:"required,gt=0,dive,not_empty"`
```
daniel-burghardt, line 20: How does the required behave in a non-nullable array field like this one? I see that when it's empty it returns the `gt` message.

internal/validators/validate.go
```diff
@@ -0,0 +1,90 @@
...
+	case "gte":
+		if fieldError.Kind() == reflect.Slice || fieldError.Kind() == reflect.Array {
+			v, err := strconv.Atoi(fieldError.Param())
+			if err != nil {
+				log.Errorf(`Error parsing "gte" param %q to integer: %s`, fieldError.Param(), err.Error())
+				return "Should have at least 1 element" // Fallback to this error message
+			}
+			return fmt.Sprintf("Should have at least %d element", v)
+		}
+		return fmt.Sprintf("Should be greater than or equal %s", fieldError.Param())
```
daniel-burghardt, line 66: This might read a little better
```suggestion
		return fmt.Sprintf("Should be greater than or equal to %s", fieldError.Param())
```

internal/validators/validate_test.go
```diff
@@ -0,0 +1,162 @@
...
+import (
+	"testing"
+
+	"github.com/go-playground/validator/v10"
+	"github.com/stellar/go/keypair"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseValidationError(t *testing.T) {
```
daniel-burghardt, line 12: I think it's amazing and super important to have these more complex test cases with dives and etc., but in addition to these, wdyt about adding separate sub tests for each validation error?
I think that way it might be easier to remember to add tests for new validations and we can better test each one individually. For instance, the `Should be greater than or equal` case of `gt` is not tested in this one.

daniel-burghardt (approved):
Amazing work! This will be super helpful! 🔥 
Left some minor suggestions and questions.

internal/serve/httperror/errors.go
```diff
@@ -8,61 +8,65 @@ import (
...
 
-var MethodNotAllowed = errorResponse{
+var MethodNotAllowed = ErrorResponse{
 	Status: http.StatusMethodNotAllowed,
 	Error:  "The method is not allowed for resource at the url requested.",
 }
 
-func BadRequest(message string) errorResponse {
+func BadRequest(message string, extras map[string]interface{}) *ErrorResponse {
```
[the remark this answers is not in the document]
CaioTeixeira95: That's because we are able to do things like:

```go
httpErr := DecodeJSONAndValidate(req, &reqBody)
if httpErr != nil {
        httpErr.Render(rw)
        return
}
```

Additionally, it's not possible to do something like this in go:

```go
return &httperror.BadRequest(...)
```

internal/validators/validate_test.go
```diff
@@ -0,0 +1,162 @@
...
+	val := NewValidator()
+	err := val.Struct(stc)
+	require.Error(t, err)
+
+	vErrs, ok := err.(validator.ValidationErrors)
+	require.True(t, ok)
+	require.Len(t, vErrs, 2)
+
+	assert.Equal(t, "publicKey", getFieldName(vErrs[0]))
+	assert.Equal(t, "children[0].name", getFieldName(vErrs[1]))
```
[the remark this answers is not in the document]
marcelosalloum: +1

internal/validators/validate_test.go
marcelosalloum: +1

internal/validators/validate_test.go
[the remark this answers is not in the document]
CaioTeixeira95: There is no particular reason, I can make it to return the entire path except for the Root struct (otherwise we'll leak an internal structure). I thought that only showing the last two would make it more clear.

internal/validators/validate_test.go
daniel-burghardt: Resolved async

PR #16, opened 2024-06-05
signing: add signature client

CaioTeixeira95:
### What

This PR adds the `SignatureClient` interface and the `EnvSignatureClient` which implements the `SignatureClient` interface. It receives a Stellar Private Key from the environment.

### Why

This layer handles the Stellar transaction signature.

### Known limitations

N/A

#### PR Structure

#### Thoroughness

#### Release


internal/signing/signature_client.go
```diff
@@ -0,0 +1,63 @@
...
+type SignatureClient interface {
+	NetworkPassphrase() string
+	GetDistributionAccountPublicKey() string
+	SignStellarTransaction(ctx context.Context, tx *txnbuild.Transaction) (*txnbuild.Transaction, error)
+}
+
+type envSignatureClient struct {
+	networkPassphrase       string
+	distributionAccountFull *keypair.Full
+}
```
marcelosalloum, line 24: I recommend you to override the String() method, in order to prevent distributionAccountFull from ever being printed.

It should be overridden in the value and not the pointer, like in:

https://github.com/stellar/stellar-disbursement-platform-backend/blob/d3912e908b6c1cea56e7aad4b79b18b7e2c583e4/internal/transactionsubmission/store/db_vault.go#L22-L24



Here are the tests that prove it should be implemented in the value. They'd not pass if the implementation was in the pointer:
https://github.com/stellar/stellar-disbursement-platform-backend/blob/d3912e908b6c1cea56e7aad4b79b18b7e2c583e4/internal/transactionsubmission/store/db_vault_test.go#L32-L64

internal/signing/signature_client.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/stellar/go/network"
+	"github.com/stellar/go/txnbuild"
+)
+
+var ErrInvalidTransaction = errors.New("invalid transaction provided")
+
+type SignatureClient interface {
+	NetworkPassphrase() string
+	GetDistributionAccountPublicKey() string
+	SignStellarTransaction(ctx context.Context, tx *txnbuild.Transaction) (*txnbuild.Transaction, error)
```
marcelosalloum, line 18: I'm assuming you'll eventually add SignFeeBumpStellarTransaction when needed.

internal/signing/signature_client_test.go
```diff
@@ -0,0 +1,78 @@
...
+	ctx := context.Background()
+
+	sc.On("NetworkPassphrase").Return(network.TestNetworkPassphrase)
+	assert.Equal(t, network.TestNetworkPassphrase, sc.NetworkPassphrase())
+	sc.On("GetDistributionAccountPublicKey").Return("pubkey")
+	assert.Equal(t, "pubkey", sc.GetDistributionAccountPublicKey())
+	sc.On("SignStellarTransaction", ctx, &txnbuild.Transaction{}).Return(nil, nil)
+	tx, err := sc.SignStellarTransaction(ctx, &txnbuild.Transaction{})
+	assert.Nil(t, tx)
+	assert.Nil(t, err)
```
marcelosalloum, line 77: This command gives better feedback when it fails:
```suggestion
	assert.NoError(t, err)
```

go.sum
marcelosalloum, line 1: I'm confused in why go.mod was not updated and go.sum was. Did you execute `go mod tidy` here? Is this something that should have been done in a previous PR?

marcelosalloum (approved):
LGTM! Plz, feel free to merge it after addressing the comments below.

go.sum
CaioTeixeira95: I'm confused too. These PRs were split into 3, but I don't know why it appeared on this PR. May I fix it in another PR that only cleans up the dependencies? I think it would be easier.

marcelosalloum: Since you're merging this into the `validators` branch (and not develop or main), this should be fine as long as @daniel-burghardt agrees with it.

daniel-burghardt: Since this is now pointing to `main` shall we just run a `go mod tidy` before merging?

daniel-burghardt (approved):
LGTM! 🔥

PR #18, opened 2024-06-05
cmd: migrate command

daniel-burghardt:
### What

Adds the `migrate up/down` command.

### Why

For applying the database migration. This is a required component for the CI/CD setup.

### Known limitations

N/A

[[Wallet Backend] Implement migrate command](https://app.asana.com/0/1202672669313222/1207481390776851/f)

#### PR Structure

#### Thoroughness

#### Release


cmd/migrate.go
```diff
@@ -0,0 +1,107 @@
...
+
+	migrateCmd := &cobra.Command{
+		Use:   "migrate",
+		Short: "Schema migration helpers",
+		PersistentPreRun: func(_ *cobra.Command, _ []string) {
+			cfgOpts.Require()
+			if err := cfgOpts.SetValues(); err != nil {
+				log.Fatalf("Error setting values of config options: %s", err.Error())
+			}
+		},
```
marcelosalloum, line 39: Before you write too many CLI commands, there's a good practice you should adopt here.

These commands have counterparts that return errors:
- `PersistentPreRun` -> `PersistentPreRunE`
- `Run` -> `RunE`
- ...

Instead of logging with Fatal, it's better to return an error and log + Panic in the root caller. It's easier to debug and to write tests for it.

`log.Fatalf` calls os.Exit(0), which terminates the program without very good feedback.

daniel-burghardt: Thanks for raising this @marcelosalloum! Great suggestion. 
I'll add to this one and refactor the other two commands we already have in place. 👍

internal/db/migrate.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	migrate "github.com/rubenv/sql-migrate"
+	"github.com/stellar/wallet-backend/internal/db/migrations"
+)
+
+func Migrate(databaseURL string, direction migrate.MigrationDirection, count int) (int, error) {
```
CaioTeixeira95, line 12: Can we add the context here instead of creating a new one?

cmd/migrate.go
```diff
@@ -0,0 +1,107 @@
...
+			}
+
+			if err := executeMigrations(cmd.Context(), databaseURL, migrate.Down, count); err != nil {
+				log.Fatalf("Error executing migrate down: %v", err)
+			}
+		},
+	}
+
+	migrateCmd.AddCommand(&migrateUpCmd)
+	migrateCmd.AddCommand(migrateDownCmd)
```
CaioTeixeira95, line 79: Is there a reason to pass one as a reference and the other as value?

cmd/migrate.go
```diff
@@ -0,0 +1,107 @@
...
+
+	if err := cfgOpts.Init(migrateCmd); err != nil {
+		log.Fatalf("Error initializing a config option: %s", err.Error())
+	}
+
+	return migrateCmd
+}
+
+func executeMigrations(ctx context.Context, databaseURL string, direction migrate.MigrationDirection, count int) error {
+	numMigrationsRun, err := db.Migrate(databaseURL, direction, count)
```
CaioTeixeira95, line 89: We can pass the context here to the `Migrate` function.

internal/db/dbtest/dbtest.go
```diff
@@ -24,3 +24,8 @@ func Open(t *testing.T) *dbtest.DB {
 
 	return db
 }
+
+func OpenWithoutMigrations(t *testing.T) *dbtest.DB {
```
CaioTeixeira95, line 28: 🚀

cmd/migrate.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/spf13/cobra"
+	"github.com/stellar/go/support/config"
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/wallet-backend/internal/db"
+)
+
+type migrateCmd struct{}
+
+func (c *migrateCmd) Command() *cobra.Command {
+	var databaseURL string
```
CaioTeixeira95, line 19: I'm seeing this config option repeating multiple times in the code perhaps we can make only this one a shared config.

CaioTeixeira95 (approved):
Thank you so much for this command! 🚀

cmd/migrate.go
daniel-burghardt: [2e5e3d0](https://github.com/stellar/wallet-backend/pull/18/commits/2e5e3d01baf15920f91f889987766224a92b723e)[2e5e3d0](https://github.com/stellar/wallet-backend/pull/18/commits/2e5e3d01baf15920f91f889987766224a92b723e)

internal/db/migrate.go
daniel-burghardt: Good call, added here 👍 : [2e5e3d0](https://github.com/stellar/wallet-backend/pull/18/commits/2e5e3d01baf15920f91f889987766224a92b723e)

cmd/migrate.go
daniel-burghardt: No reason, it was a typo, I've made them equal now ✅

cmd/migrate.go
daniel-burghardt: Agreed, done! ✅  [e6c6df9](https://github.com/stellar/wallet-backend/pull/18/commits/e6c6df99aefd4c5439ecaef1d3321c030b6a64ea)

PR #24, opened 2024-06-13
all: channel accounts

CaioTeixeira95:
### What

This PR implements channel accounts.

### Why

To have a high rate of transactions.

### Known limitations

Add tests and more channel account commands.

#### PR Structure

#### Thoroughness

#### Release


internal/services/account_sponsorship_service.go
```diff
@@ -33,8 +33,8 @@ func (e ErrOperationNotAllowed) Error() string {
 
 const (
 	// Sufficient to cover three average ledger close time.
-	CreateAccountTxnTimeBounds             = 18
-	CreateAccountTxnTimeBoundsSafetyMargin = 12
+	CreateAccountTxnTimeBounds             = 60
+	CreateAccountTxnTimeBoundsSafetyMargin = 60
```
marcelosalloum, line 37: 👋  This change is a bad idea, because the network usually times out in less than 30s. If you use 60 seconds here, you'll add complexity top how you will handle some indeterminate responses.

internal/services/account_sponsorship_service.go
```diff
@@ -113,15 +119,19 @@ func (s *accountSponsorshipService) SponsorAccountCreationTransaction(ctx contex
 		},
 	)
 
-	// TODO: use Channel Accounts instead of using the Distribution Account.
-	distributionAccount, err := s.HorizonClient.AccountDetail(horizonclient.AccountRequest{AccountID: s.SignatureClient.GetDistributionAccountPublicKey()})
+	channelAccountPublicKey, err := s.ChannelAccountSignatureClient.GetAccountPublicKey(ctx)
+	if err != nil {
+		return "", "", fmt.Errorf("getting channel account public key: %w", err)
+	}
```
marcelosalloum, line 125: 💡 It'd be a good idea to add a custom error for when there is no account available, or no account at all.

internal/data/accounts.go
```diff
@@ -31,12 +31,19 @@ func (m *AccountModel) Delete(ctx context.Context, address string) error {
...
-	const query = `SELECT EXISTS(SELECT stellar_address FROM accounts WHERE stellar_address = $1)`
+func (m *AccountModel) IsAccountFeeBumpEligible(ctx context.Context, address string) (bool, error) {
+	const query = `
+		SELECT 
+			EXISTS(
+				SELECT stellar_address FROM accounts WHERE stellar_address = $1
+				UNION
+				SELECT public_key FROM channel_accounts WHERE public_key = $1
+			)
+	`
```
marcelosalloum, line 42: 👋  I'm confused here... is this method saying that channel accounts can sign a FeeBumpTransaction? Fee bump transactions are used when you have a distribution account paying for the gas fees, so I don't think a Channel Account should be eligible to sign fee bump transactions.

internal/services/account_sponsorship_service.go
```diff
@@ -201,12 +221,16 @@ func (s *accountSponsorshipService) WrapTransaction(ctx context.Context, tx *txn
...
 	}
 
-	return feeBumpTxe, s.SignatureClient.NetworkPassphrase(), nil
+	return feeBumpTxe, s.DistributionAccountSignatureClient.NetworkPassphrase(), nil
 }
 
-func NewAccountSponsorshipService(signatureClient signing.SignatureClient, horizonClient horizonclient.ClientInterface, maxSponsoredBaseReserves int, baseFee int64, models *data.Models) (*accountSponsorshipService, error) {
-	if signatureClient == nil {
-		return nil, fmt.Errorf("signature client cannot be nil")
+func NewAccountSponsorshipService(distributionAccountSignatureClient, channelAccountSignatureClient signing.SignatureClient, horizonClient horizonclient.ClientInterface, maxSponsoredBaseReserves int, baseFee int64, models *data.Models) (*accountSponsorshipService, error) {
```
marcelosalloum, line 227: polishing: when we start adding lots of arguments to a method, it starts becoming hard to figure out what they're about. From a caller's perspective, it's hard to grasp what an int vs int64 means unless you take a deeper look into the docs.

I suggest using a struct to pass on these options.

cmd/channel_account.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (c *channelAccountCmd) Command() *cobra.Command {
+	cfg := channelAccountCmdConfigOptions{}
+	cfgOpts := config.ConfigOptions{
+		utils.DatabaseURLOption(&cfg.DatabaseURL),
+		utils.NetworkPassphraseOption(&cfg.NetworkPassphrase),
+		utils.BaseFeeOption(&cfg.BaseFee),
+		utils.HorizonClientURLOption(&cfg.HorizonClientURL),
+		utils.DistributionAccountPrivateKeyOption(&cfg.DistributionAccountPrivateKey),
```
marcelosalloum, line 39: 👋 you should have a separate encryption hey just for encrypting channel accounts. You don't want the host to have to handle the distribution secret when they need to rotate the channel account encryption key.

internal/services/channel_account_service.go
```diff
@@ -0,0 +1,150 @@
...
+	return &channelAccountService{
+		DB:                                 opts.DB,
+		HorizonClient:                      opts.HorizonClient,
+		BaseFee:                            opts.BaseFee,
+		DistributionAccountSignatureClient: opts.DistributionAccountSignatureClient,
+		ChannelAccountStore:                opts.ChannelAccountStore,
+		PrivateKeyEncrypter:                opts.PrivateKeyEncrypter,
+		EncryptionPassphrase:               opts.EncryptionPassphrase,
+	}
+}
```
marcelosalloum, line 150: polishing: what's the advantage of such constructor if all you're doing is assign the variables one by one?

internal/signing/channelaccounts/channel_accounts_model.go
```diff
@@ -0,0 +1,150 @@
...
+			FOR UPDATE SKIP LOCKED
+		)
+		RETURNING *
+	`, int64(lockedUntil.Seconds()))
+
+	var channelAccount ChannelAccount
+	for range ChannelAccountWaitTime {
+		err := ca.DB.GetContext(ctx, &channelAccount, query)
+		if errors.Is(err, sql.ErrNoRows) {
+			log.Ctx(ctx).Info("All channel accounts are in use. Retry in 1 second.")
```
marcelosalloum, line 68: 👋 This log should be a warning.

internal/signing/channelaccounts/channel_accounts_model.go
```diff
@@ -0,0 +1,150 @@
...
+				OR locked_until < NOW()
+			ORDER BY random()
+			LIMIT 1
+			FOR UPDATE SKIP LOCKED
+		)
+		RETURNING *
+	`, int64(lockedUntil.Seconds()))
+
+	var channelAccount ChannelAccount
+	for range ChannelAccountWaitTime {
```
marcelosalloum, line 65: 👋 this retry logic should 100% not be in the model. You should remove it and address it in another PR if needed.

internal/signing/env_signature_client.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/network"
+	"github.com/stellar/go/txnbuild"
+)
+
+type envSignatureClient struct {
+	networkPassphrase       string
+	distributionAccountFull *keypair.Full
+}
```
marcelosalloum, line 15: Plz, make sure to override the String() method to prevent the secret key from ever being printed. You can do that in another PR.

marcelosalloum (commented):
Thanks a lot for putting all of this together.

I've left a few comments below, the most important ones are marked with the emoji 👋

internal/services/account_sponsorship_service.go
CaioTeixeira95: I'm sorry, this was only for testing. the correct one is 30 seconds.

internal/serve/middleware/middleware.go
```diff
@@ -47,3 +48,30 @@ func SignatureMiddleware(signatureVerifier auth.SignatureVerifier) func(next htt
 		})
 	}
 }
+
+// RecoverHandler is a middleware that recovers from panics and logs the error.
+func RecoverHandler(next http.Handler) http.Handler {
```
daniel-burghardt, line 53: Nice! 🚀 
Shall we also add a TODO comment to add sentry logging here?

internal/serve/serve.go
```diff
@@ -140,6 +151,7 @@ func handler(deps handlerDeps) http.Handler {
 			handler := &httphandler.AccountHandler{
 				AccountSponsorshipService: deps.AccountSponsorshipService,
 				SupportedAssets:           deps.SupportedAssets,
+				BlockedOperationsTypes:    blockedOperationTypes,
```
daniel-burghardt, line 154: This is unused after the refactor moving it to the service, right?
```suggestion
```

internal/services/account_sponsorship_service.go
daniel-burghardt: I think that would be quite useful indeed.

internal/services/channel_account_service.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	hash, err := signedTx.Hash(s.DistributionAccountSignatureClient.NetworkPassphrase())
+	if err != nil {
+		return fmt.Errorf("getting transaction hash: %w", err)
+	}
+
+	_, err = s.HorizonClient.SubmitTransaction(signedTx)
+	if err != nil {
+		if hError := horizonclient.GetError(err); hError != nil {
```
daniel-burghardt, line 114: Maybe we could start thinking of creating a submission service for HTTP requests with retry logic in it (not in this PR for sure). Wdyt of that idea?

internal/signing/channel_account_db_signature_client.go
```diff
@@ -0,0 +1,108 @@
...
+		if _, ok := accountsAlreadyAccountedFor[account]; ok {
+			continue
+		}
+		accountsAlreadyAccountedFor[account] = struct{}{}
+
+		if account == "" {
+			return nil, fmt.Errorf("account %d is empty", i)
+		}
+
+		chAcc, err := sc.channelAccountStore.Get(ctx, sc.dbConnectionPool, account)
```
daniel-burghardt, line 68: 💡 if we want to make it more performant, perhaps we could get all accounts from the database at once, before the loop

internal/signing/channelaccounts/channel_accounts_model.go
```diff
@@ -0,0 +1,150 @@
...
+
+		publicKeys[i] = ca.PublicKey
+		encryptedPrivateKeys[i] = ca.EncryptedPrivateKey
+	}
+
+	const q = `
+		INSERT INTO 
+			channel_accounts (public_key, encrypted_private_key)
+		SELECT * 
+			FROM UNNEST($1::text[], $2::text[])
```
daniel-burghardt, line 120: One nice alternative for Postgres is this: https://pkg.go.dev/github.com/lib/pq#hdr-Bulk_imports
But might not be such a good idea because it make the implementation more coupled to one database system.

daniel-burghardt (commented):
Great work! 💪 Can't believe you pulled this off in a day 😦.
I'm happy to discuss the comments on a call if you'd like, might be more efficient.

internal/signing/channelaccounts/channel_accounts_model.go
CaioTeixeira95: I prefer keeping it the way it is now.

internal/serve/serve.go
```diff
@@ -126,6 +144,17 @@ func getHandlerDeps(cfg Configs) (handlerDeps, error) {
 	}, nil
 }
 
+func ensureChannelAccounts(channelAccountService services.ChannelAccountService, numberOfChannelAccounts int64) {
+	ctx := context.Background()
+	log.Ctx(ctx).Info("Ensuring the number of channel accounts in the database...")
+	err := channelAccountService.EnsureChannelAccounts(ctx, numberOfChannelAccounts)
+	if err != nil {
+		log.Ctx(ctx).Errorf("error ensuring the number of channel accounts: %s", err.Error())
+		return
```
daniel-burghardt, line 153: Whenever we implement Sentry, I think this might be worthy of a Sentry exception as well, wdyt?

internal/signing/channel_account_db_signature_client.go
```diff
@@ -41,18 +43,47 @@ func (sc *channelAccountDBSignatureClient) NetworkPassphrase() string {
...
+	}
+
+	numOfChannelAccounts, err := sc.channelAccountStore.Count(ctx)
 	if err != nil {
-		return "", fmt.Errorf("getting idle channel account: %w", err)
+		return "", fmt.Errorf("getting the number of channel accounts: %w", err)
+	}
+
+	if numOfChannelAccounts > 0 {
+		return "", channelaccounts.ErrNoIdleChannelAccountAvailable
```
daniel-burghardt, line 66: 🚀

internal/signing/channelaccounts/channel_accounts_model.go
```diff
@@ -31,6 +31,7 @@ type ChannelAccount struct {
 type ChannelAccountStore interface {
 	GetIdleChannelAccount(ctx context.Context, lockedUntil time.Duration) (*ChannelAccount, error)
 	Get(ctx context.Context, sqlExec db.SQLExecuter, publicKey string) (*ChannelAccount, error)
+	GetAllByPublicKey(ctx context.Context, sqlExec db.SQLExecuter, publicKeys ...string) ([]*ChannelAccount, error)
```
daniel-burghardt, line 34: Do we still need the `Get` function in this case?

internal/data/accounts.go
daniel-burghardt: Addressed here: https://github.com/stellar/wallet-backend/blob/049a643b4996aa355e7b6365907d1000fa3f51d4/internal/data/accounts.go#L34-L35

daniel-burghardt (approved):
LGTM! Awesome! 🚀

internal/signing/channelaccounts/channel_accounts_model.go
CaioTeixeira95: Yes, we do want to preserve compatibility with the TSS, which uses this method.

PR #27, opened 2024-06-19
all: channel accounts tests

CaioTeixeira95:
### What

This PR adds tests for the channel accounts feature

### Why

Tech debit.

### Known limitations

N/A

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/serve/middleware/middleware_test.go
+++ b/internal/serve/middleware/middleware_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	"github.com/go-chi/chi"
+	"github.com/sirupsen/logrus"
 	"github.com/stellar/go/support/log"
 	"github.com/stellar/wallet-backend/internal/serve/auth"
 	"github.com/stretchr/testify/assert"
@@ -135,4 +136,34 @@ func TestSignatureMiddleware(t *testing.T) {
 		assert.Equal(t, http.StatusOK, resp.StatusCode)
 		assert.JSONEq(t, `{"status":"ok"}`, string(respBody))
 	})
+}
+
+func Test_RecoverHandler(t *testing.T) {
+	// setup logger to assert the logged texts later
+	buf := new(strings.Builder)
+	log.DefaultLogger.SetOutput(buf)
+	log.DefaultLogger.SetLevel(logrus.TraceLevel)
+
+	// setup
+	r := chi.NewRouter()
+	r.Use(RecoverHandler)
+	r.Get("/", func(w http.ResponseWriter, r *http.Request) {
+		panic("test panic")
+	})
+
+	// test
+	req, err := http.NewRequest("GET", "/", nil)
+	require.NoError(t, err)
+	rr := httptest.NewRecorder()
+	r.ServeHTTP(rr, req)
+
+	// assert response
+	assert.Equal(t, http.StatusInternalServerError, rr.Code)
+	wantJson := `{
+		"error": "An error occurred while processing this request."
+	}`
+	assert.JSONEq(t, wantJson, rr.Body.String())
+
+	// assert logged text
+	assert.Contains(t, buf.String(), "panic: test panic", "should log the panic message")
 }
--- a/internal/serve/middleware/middleware_test.go
+++ b/internal/serve/middleware/middleware_test.go
@@ -10,7 +10,6 @@ import (
 	"testing"
 
 	"github.com/go-chi/chi"
-	"github.com/sirupsen/logrus"
 	"github.com/stellar/go/support/log"
 	"github.com/stellar/wallet-backend/internal/serve/auth"
 	"github.com/stretchr/testify/assert"
@@ -138,11 +137,8 @@ func TestSignatureMiddleware(t *testing.T) {
 	})
 }
 
-func Test_RecoverHandler(t *testing.T) {
-	// setup logger to assert the logged texts later
-	buf := new(strings.Builder)
-	log.DefaultLogger.SetOutput(buf)
-	log.DefaultLogger.SetLevel(logrus.TraceLevel)
+func TestRecoverHandler(t *testing.T) {
+	getEntries := log.DefaultLogger.StartTest(log.ErrorLevel)
 
 	// setup
 	r := chi.NewRouter()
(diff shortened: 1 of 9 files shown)


internal/serve/middleware/middleware_test.go
```diff
@@ -136,3 +137,33 @@ func TestSignatureMiddleware(t *testing.T) {
 		assert.JSONEq(t, `{"status":"ok"}`, string(respBody))
 	})
 }
+
+func Test_RecoverHandler(t *testing.T) {
+	// setup logger to assert the logged texts later
+	buf := new(strings.Builder)
+	log.DefaultLogger.SetOutput(buf)
+	log.DefaultLogger.SetLevel(logrus.TraceLevel)
```
daniel-burghardt, line 145: Shall we use the same as we do in https://github.com/stellar/wallet-backend/blob/a652c1e07264e4503698326bb33038a03474b34d/internal/signing/channel_account_db_signature_client_test.go#L64?

internal/signing/channel_account_db_signature_client_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/stellar/go/network"
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/go/txnbuild"
+	"github.com/stellar/wallet-backend/internal/signing/channelaccounts"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestChannelAccountDBSignatureClientGetAccountPublicKey(t *testing.T) {
+	t.Parallel()
```
daniel-burghardt, line 18: Why are using `t.Parallel()` here?

internal/signing/channel_account_db_signature_client_test.go
```diff
@@ -0,0 +1,154 @@
...
+		channelAccountStore.
+			On("GetIdleChannelAccount", ctx, time.Minute).
+			Return(nil, channelaccounts.ErrNoIdleChannelAccountAvailable).
+			Times(6).
+			On("Count", ctx).
+			Return(0, nil).
+			Once()
+		defer channelAccountStore.AssertExpectations(t)
+
+		getEntries := log.DefaultLogger.StartTest(log.WarnLevel)
```
daniel-burghardt, line 64: 🔥 this is awesome

internal/signing/channelaccounts/channel_accounts_model_test.go
```diff
@@ -0,0 +1,211 @@
...
+	"time"
+
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/wallet-backend/internal/db"
+	"github.com/stellar/wallet-backend/internal/db/dbtest"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func createChannelAccountFixture(t *testing.T, ctx context.Context, dbConnectionPool db.ConnectionPool, publicKey, encryptedPrivateKey string) {
```
daniel-burghardt, line 15: Maybe we can make this take an array of keypairs as param to avoid multiple rounds to the DB.

internal/signing/channelaccounts/channel_accounts_model_test.go
```diff
@@ -0,0 +1,211 @@
...
+			SET
+				locked_at = NOW(),
+				locked_until = NOW() + '5 minutes'::INTERVAL
+			WHERE
+				public_key = $1
+		`
+		_, err := dbConnectionPool.ExecContext(ctx, lockChannelAccountQuery, channelAccount1.Address())
+		require.NoError(t, err)
+		_, err = dbConnectionPool.ExecContext(ctx, lockChannelAccountQuery, channelAccount2.Address())
+		require.NoError(t, err)
```
daniel-burghardt, line 56: Same here, perhaps we could make this into a single query? Optimizing the tests to avoid the huge test runs later on 😅

internal/signing/channelaccounts/channel_accounts_model_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+		ca, err := m.GetIdleChannelAccount(ctx, time.Minute)
+		assert.ErrorIs(t, err, ErrNoIdleChannelAccountAvailable)
+		assert.Nil(t, ca)
+	})
+
+	t.Run("returns_error_when_there's_no_channel_account_available", func(t *testing.T) {
```
daniel-burghardt, line 63:
```suggestion
	t.Run("returns_channel_account_available", func(t *testing.T) {
```
Interesting that Go does not complain about subtests with duplicate names.

internal/signing/channelaccounts/channel_accounts_model_test.go
```diff
@@ -0,0 +1,211 @@
...
+		assert.EqualError(t, err, "private key cannot be empty")
+	})
+
+	t.Run("inserts_channel_accounts_successfully", func(t *testing.T) {
+		channelAccount1 := keypair.MustRandom()
+		channelAccount2 := keypair.MustRandom()
+		channelAccounts := []*ChannelAccount{
+			{
+				PublicKey:           channelAccount1.Address(),
+				EncryptedPrivateKey: channelAccount1.Seed(),
```
daniel-burghardt, line 175: Perhaps would be interesting to check on `BatchInsert` that the private key is not in its raw format but actually encrypted, wdyt?

daniel-burghardt (approved):
Awesome! The interfaces really help testing these more complex services 🚀 
Left some small comments.

internal/signing/channel_account_db_signature_client_test.go
CaioTeixeira95: Because the function tested has a sleep statement, we will block the other tests for at least 6 seconds if we don't run it in parallel.

internal/signing/channelaccounts/channel_accounts_model_test.go
CaioTeixeira95: Here I've to disagree, it will not make tests slower. But, I'll refactor it to run in one query.

internal/signing/channelaccounts/channel_accounts_model_test.go
CaioTeixeira95: This logic shouldn't be in the model layer.

PR #28, opened 2024-06-24
payments: get endpoint

daniel-burghardt:
### What

Adds a new `GET /payments` endpoint for fetching the ingested payments.

### Why

New endpoint.

### Known limitations

N/A

[[Wallet Backend] Implement /payments endpoint](https://app.asana.com/0/1202672669313222/1207481390776866/f)

#### PR Structure

#### Thoroughness

#### Release


internal/data/payments_test.go
```diff
@@ -145,3 +145,67 @@ func TestPaymentModelUpdateLatestLedgerSynced(t *testing.T) {
...
+		{OperationID: 5, OperationType: "OperationTypePayment", TransactionID: 55, TransactionHash: "edfab36f9f104c4fb74b549de44cfbcc", FromAddress: "GA4CMYJEC5W5", ToAddress: "GAZ37ZO4TU3H", SrcAssetCode: "USDC", SrcAssetIssuer: "GAHLU7PDIQMZ", SrcAmount: 50, DestAssetCode: "USDC", DestAssetIssuer: "GAHLU7PDIQMZ", DestAmount: 50, CreatedAt: time.Date(2024, 6, 25, 0, 0, 0, 0, time.UTC), Memo: nil},
+	}
+
+	const query = `INSERT INTO ingest_payments (operation_id, operation_type, transaction_id, transaction_hash, from_address, to_address, src_asset_code, src_asset_issuer, src_amount, dest_asset_code, dest_asset_issuer, dest_amount, created_at, memo) VALUES (:operation_id, :operation_type, :transaction_id, :transaction_hash, :from_address, :to_address, :src_asset_code, :src_asset_issuer, :src_amount, :dest_asset_code, :dest_asset_issuer, :dest_amount, :created_at, :memo);`
+	_, err = dbConnectionPool.NamedExecContext(ctx, query, dbPayments)
+	require.NoError(t, err)
```
CaioTeixeira95, line 171: This is great! My only suggestion here is we move it to a fixture, so we can reuse it as much as possible.

CaioTeixeira95: Also TIL

internal/data/payments.go
```diff
@@ -91,3 +91,32 @@ func (m *PaymentModel) AddPayment(ctx context.Context, tx db.Transaction, paymen
...
+	}
+
+	const query = `
+		SELECT * FROM ingest_payments
+		WHERE
+			($1 = '' OR $1 IN (from_address, to_address)) AND
+			($2 = 0 OR CASE
+				WHEN $3 = 'DESC' THEN operation_id < $2
+				ELSE operation_id > $2
+			END)
```
CaioTeixeira95, line 107: This is a good approach.

internal/db/db.go
```diff
@@ -77,6 +77,7 @@ var _ Transaction = (*sqlx.Tx)(nil)
 type SQLExecuter interface {
 	DriverName() string
 	ExecContext(ctx context.Context, query string, args ...interface{}) (sql.Result, error)
+	NamedExecContext(ctx context.Context, query string, arg interface{}) (sql.Result, error)
```
CaioTeixeira95, line 80: Nice! 🚀

internal/serve/httphandler/account_handler.go
```diff
@@ -35,12 +34,7 @@ func (h AccountHandler) SponsorAccountCreation(rw http.ResponseWriter, req *http
 	ctx := req.Context()
 
 	var reqBody SponsorAccountCreationRequest
-	if err := httpdecode.DecodeJSON(req, &reqBody); err != nil {
-		httperror.BadRequest("", nil).Render(rw)
-		return
-	}
-
-	httpErr := ValidateRequestBody(ctx, reqBody)
+	httpErr := DecodeJSONAndValidate(ctx, req, &reqBody)
```
CaioTeixeira95, line 37: Thank you! 🚀

internal/serve/httphandler/payments_handler_test.go
```diff
@@ -201,3 +202,244 @@ func TestUnsubscribeAddress(t *testing.T) {
...
+		{OperationID: 3, OperationType: "OperationTypePayment", TransactionID: 33, TransactionHash: "d9521ed7057d4d1e9b9dd22ab515cbf1", FromAddress: "GCXBGEYNIEIUJ56YX5UVBM27NTKCBMLDD2NEPTTXZGQMBA2EOKG5VA2W", ToAddress: "GAX6VPTVC2YNJM52OYMJAZKTQMSLNQ6NKYYU77KSGRVHINZ2D3EUJWAN", SrcAssetCode: "XLM", SrcAssetIssuer: "", SrcAmount: 30, DestAssetCode: "XLM", DestAssetIssuer: "", DestAmount: 30, CreatedAt: time.Date(2024, 6, 23, 0, 0, 0, 0, time.UTC), Memo: nil},
+	}
+
+	const query = `INSERT INTO ingest_payments (operation_id, operation_type, transaction_id, transaction_hash, from_address, to_address, src_asset_code, src_asset_issuer, src_amount, dest_asset_code, dest_asset_issuer, dest_amount, created_at, memo) VALUES (:operation_id, :operation_type, :transaction_id, :transaction_hash, :from_address, :to_address, :src_asset_code, :src_asset_issuer, :src_amount, :dest_asset_code, :dest_asset_issuer, :dest_amount, :created_at, :memo);`
```
CaioTeixeira95, line 233: Just like my previous suggestions, I'd recommend we create a fixture in the data package to insert payments.

internal/serve/httphandler/request_body_validator.go
CaioTeixeira95, line 1: I think we can rename this file to `request_params_validator.go`

internal/serve/httphandler/payments_handler_test.go
```diff
@@ -201,3 +202,244 @@ func TestUnsubscribeAddress(t *testing.T) {
...
+		r.ServeHTTP(rr, req)
+
+		// Assert 200 response
+		assert.Equal(t, http.StatusOK, rr.Code)
+
+		resp := rr.Result()
+		respBody, err := io.ReadAll(resp.Body)
+		require.NoError(t, err)
+		expectedRespBody := `{
+			"payments": [
```
CaioTeixeira95, line 253: I liked this response, but I think this is not a paginated query. We need to add the cursor here so the caller won't need to get always the last value to calculate the next page.

CaioTeixeira95 (commented):
Thank you for working on this! Clear and easy to follow! 🚀

internal/serve/httphandler/payments_handler_test.go
daniel-burghardt: As agreed on call, I've added a `_links` object containing the URLs to the previous and next pages

internal/data/fixtures.go
```diff
@@ -0,0 +1,3 @@
+package data
+
+const InsertPaymentsQuery = `INSERT INTO ingest_payments (operation_id, operation_type, transaction_id, transaction_hash, from_address, to_address, src_asset_code, src_asset_issuer, src_amount, dest_asset_code, dest_asset_issuer, dest_amount, created_at, memo) VALUES (:operation_id, :operation_type, :transaction_id, :transaction_hash, :from_address, :to_address, :src_asset_code, :src_asset_issuer, :src_amount, :dest_asset_code, :dest_asset_issuer, :dest_amount, :created_at, :memo);`
```
CaioTeixeira95, line 3: I'd recommend we work with functions instead of providing queries for fixtures because using functions is easier. Additionally, it's easy to maintain. For example, I as a developer would need to look a this query every time to understand what field I need to pass to it. Also, it's not clear that this is used only for testing purposes, someone would misunderstand and use it on the actual code.

This is not a blocker, it's just that we need to define a pattern.

internal/data/payments.go
```diff
@@ -91,3 +92,86 @@ func (m *PaymentModel) AddPayment(ctx context.Context, tx db.Transaction, paymen
...
+
+	const filteredSetCTE = `
+		WITH filtered_set AS (
+			SELECT * FROM ingest_payments WHERE $1 = '' OR $1 IN (from_address, to_address)
+		)
+	`
+
+	var selectQ string
+	if beforeID != 0 && sort == DESC {
+		selectQ = fmt.Sprintf("SELECT * FROM (SELECT * FROM filtered_set WHERE operation_id > %d ORDER BY operation_id ASC LIMIT $2) AS reverse_set ORDER BY operation_id DESC", beforeID)
```
CaioTeixeira95, line 113: This is a security issue, we should never use external values directly in our queries. Here we should use the `?` operator, then we can use the `Rebind` function to translate it to `$n`.

internal/entities/pagination.go
```diff
@@ -0,0 +1,11 @@
+package entities
+
+type PaginationLinks struct {
```
CaioTeixeira95, line 3: 🚀

internal/serve/httphandler/payment_handler.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/stellar/go/support/render/httpjson"
+	"github.com/stellar/wallet-backend/internal/data"
+	"github.com/stellar/wallet-backend/internal/entities"
+	"github.com/stellar/wallet-backend/internal/serve/httperror"
+	"github.com/stellar/wallet-backend/internal/services"
+)
+
+type PaymentsHandler struct {
+	Models  *data.Models
+	Service *services.PaymentService
```
CaioTeixeira95, line 15: A pattern that we are for handlers is that we always use an interface to the services it uses, so it's easy to mock and test here.

internal/services/payment_service.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/stellar/wallet-backend/internal/data"
+	"github.com/stellar/wallet-backend/internal/entities"
+)
+
+type PaymentService struct {
+	Models        *data.Models
+	ServerBaseURL string
+}
+
+func (s *PaymentService) GetPaymentsPaginated(ctx context.Context, address string, beforeID, afterID int64, sort data.SortOrder, limit int) ([]data.Payment, entities.Pagination, error) {
```
CaioTeixeira95, line 18: I loved this! It will make this logic reusable and easy to test! Thank you.

CaioTeixeira95 (commented):
Thank you for the pagination implementation! This is looking great.

internal/serve/httphandler/payment_handler.go
CaioTeixeira95: nit: can we rename the `PaymentsHandler` to `PaymentHandler`?

CaioTeixeira95 (approved):

internal/data/fixtures.go
daniel-burghardt: [fixture reorganized](https://github.com/stellar/wallet-backend/pull/28/commits/77e0be4d68fccbfa99174ab67fd51955b08524fb)

internal/data/payments.go
daniel-burghardt: See [b3aac4f](https://github.com/stellar/wallet-backend/pull/28/commits/b3aac4f2478549997b8c8e0eed4c5584623cea0f)

PR #29, opened 2024-06-27
ingest: asset_types and memo_type columns added

daniel-burghardt:
### What

Adds three columns to the `ingest_payments` table, to be persisted when running ingestion.
Columns: `src_asset_type`, `dest_asset_type` and `memo_type`.

### Why

These are relevant columns to the Vibrant wallet and potentially to other wallets as well.

### Known limitations

If ingestion had been running on any instance prior to this merge and there are any records in the `ingest_payments` table, those must be wiped before the migration can run.
Since the Wallet Backend is in development phase and there are no instances running yet, this shouldn't be a problem.

[[Wallet Backend] Add new columns to payments ingestion](https://app.asana.com/0/1202672669313222/1207550304546018/f)

#### PR Structure

#### Thoroughness

#### Release


internal/db/migrations/2024-05-22.0-ingest_payments.sql
```diff
@@ -20,7 +20,7 @@ CREATE TABLE ingest_payments (
   dest_asset_issuer text NOT NULL,
   dest_amount bigint NOT NULL,
   created_at timestamp with time zone NOT NULL,
-  memo text,
+  memo text NULL,
```
CaioTeixeira95, line 23: Now that we've got a deployed version, we should be careful about modifying the migrations.

internal/db/migrations/2024-06-27.0-ingest_payments_2.sql
```diff
@@ -0,0 +1,13 @@
+-- +migrate Up
+
+ALTER TABLE ingest_payments
+  ADD COLUMN src_asset_type text NOT NULL,
+  ADD COLUMN dest_asset_type text NOT NULL,
```
CaioTeixeira95, line 5: We are safe, but if we have entries in this table this migration will fail.

internal/services/ingest_test.go
```diff
@@ -120,25 +121,27 @@ func TestProcessLedger(t *testing.T) {
...
 		SrcAssetIssuer:  "GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5",
+		SrcAssetType:    xdr.AssetTypeAssetTypeCreditAlphanum4.String(),
 		SrcAmount:       50,
 		DestAssetCode:   "USDC",
 		DestAssetIssuer: "GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5",
+		DestAssetType:   xdr.AssetTypeAssetTypeCreditAlphanum4.String(),
 		DestAmount:      50,
 		CreatedAt:       time.Date(2024, 5, 28, 11, 0, 0, 0, time.UTC),
-		Memo:            &expectedMemo,
+		Memo:            utils.PointOf("memo_test"),
```
CaioTeixeira95, line 144: 🚀

CaioTeixeira95 (approved):
LGTM 🚀

internal/db/migrations/2024-05-22.0-ingest_payments.sql
daniel-burghardt: For sure! But in this case it should be safe, since this is just making the default value explicit.

internal/db/migrations/2024-06-27.0-ingest_payments_2.sql
daniel-burghardt: Indeed! I also tried explaining that in the **Known limitations** section.

internal/db/migrations/2024-05-22.0-ingest_payments.sql
CaioTeixeira95: That's great! 🚀

PR #30, opened 2024-07-17
main: stellar core added to docker image

daniel-burghardt:
### What

Adds Stellar Core to the Docker image plus makes some minor improvements to the ingest command to make it easier to run the service without any manual configuration needed.

### Why

This is needed to deploy the ingest service to the cloud infra.

### Known limitations

N/A

[[Wallet Backend] Write Ingestion Kubernetes manifest](https://app.asana.com/0/1202672669313222/1207481390776874/f)

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,8 @@ RUN go build -o /bin/wallet-backend -ldflags "-X main.GitCommit=$GIT_COMMIT" .
 # Step 2: Install Stellar Core and copy over app binary
 FROM ubuntu:jammy AS core-build
 
+ARG STELLAR_CORE_VERSION
+
 RUN apt-get update && \
     apt-get install -y --no-install-recommends ca-certificates curl wget gnupg apt-utils gpg && \
     curl -sSL https://apt.stellar.org/SDF.asc | gpg --dearmor >/etc/apt/trusted.gpg.d/SDF.gpg && \
@@ -19,7 +21,7 @@ RUN apt-get update && \
     echo "deb https://apt.stellar.org jammy testing" >/etc/apt/sources.list.d/SDF-testing.list && \
     echo "deb https://apt.stellar.org jammy unstable" >/etc/apt/sources.list.d/SDF-unstable.list && \
     apt-get update && \
-    apt-get install -y stellar-core=21.0.0-1872.c6f474133.jammy && \
+    apt-get install -y stellar-core=${STELLAR_CORE_VERSION} && \
     apt-get clean
 
 COPY --from=api-build /bin/wallet-backend /app/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,34 @@
-# To build:
-#    make docker-build
-# To push:
-#    make docker-push
-
-FROM golang:1.22.4-bullseye as build
+# Step 1: Build Go API
+FROM golang:1.22.4-bullseye as api-build
 ARG GIT_COMMIT
 
 WORKDIR /src/wallet-backend
-ADD go.mod go.sum ./
+COPY go.mod go.sum ./
 RUN go mod download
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -8,7 +8,14 @@ TAG ?= stellar/wallet-backend:$(LABEL)
...
 
 docker-build:
-	$(SUDO) docker build -f Dockerfile --pull --label org.opencontainers.image.created="$(BUILD_DATE)" -t $(TAG) --build-arg GIT_COMMIT=$(LABEL) .
+	$(SUDO) docker build \
+		--file Dockerfile \
+		--pull \
+		--label org.opencontainers.image.created="$(BUILD_DATE)" \
+		--tag $(TAG) \
+		--build-arg GIT_COMMIT=$(LABEL) \
+		--platform linux/amd64 \
```
mwtzzz, line 17: what's the reason for specifying platform amd64 rather than leaving it platform-agnostic?

mwtzzz: first time deploy. needs it.

mwtzzz (approved):
LGTM

gsoarescosta (approved):
LGTM

CaioTeixeira95 (approved):

Dockerfile
```diff
@@ -1,22 +1,34 @@
...
+FROM ubuntu:jammy AS core-build
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends ca-certificates curl wget gnupg apt-utils gpg && \
+    curl -sSL https://apt.stellar.org/SDF.asc | gpg --dearmor >/etc/apt/trusted.gpg.d/SDF.gpg && \
+    echo "deb https://apt.stellar.org jammy stable" >/etc/apt/sources.list.d/SDF.list && \
+    echo "deb https://apt.stellar.org jammy testing" >/etc/apt/sources.list.d/SDF-testing.list && \
+    echo "deb https://apt.stellar.org jammy unstable" >/etc/apt/sources.list.d/SDF-unstable.list && \
+    apt-get update && \
+    apt-get install -y stellar-core=21.0.0-1872.c6f474133.jammy && \
```
jacekn, line 22: It would be worth parameterising the version so that you don't have to edit the Dockerfile each time new core is released.

daniel-burghardt: Good point @jacekn! To be honest I had it parameterized before, but what happened is that that way I had to update two different places when upgrading the version: the Makefile's default value and the docker-compose.yaml so that they're matching. So I ended up changing it to be fixed in a central place (the Dockerfile).

In Vibrant at least, every time we wanted to upgrade the version I made a deploy changing the value in the Makefile. I guess we could also update the Jenkins pipeline. But I'm wondering how this is usually being done in Horizon / Soroban RPC @jacekn?

jacekn: In case of horizon we choose core version from a dropdown in Jenkins and pass as parameter:
https://github.com/stellar/go/blob/master/services/horizon/docker/Makefile

But if you find it easier to maintain it here that's OK. Another option could be to make it a parameter but have default version set in the Makefile. That way anybody can build with any version and you'd open PRs against the Makefile to update the default version.

daniel-burghardt: I've changed it back here: [45cb905](https://github.com/stellar/wallet-backend/pull/30/commits/45cb905de44d5e47121842c851cae0780878619e)
I think it will be easier for other dev teams to configure it all in one place (the Makefile args).

daniel-burghardt: Good to know about how Horizon is doing it. I think in our case for now we don't need it to be configurable to that level (with the Jenkins dropdown) but will be easy to add it if necessary.

jacekn (approved):

PR #32, opened 2024-07-25
payment: tx and operation id types made string

daniel-burghardt:
### What

Refactors the `OperationID` and `TransactionID` properties of `Payment` to be of type `string` instead of `int64`.

### Why

The number type was causing overflow problems in clients consuming the endpoint for production environments where the ID's are too large.

### Known limitations

The database columns types are still `bigint`, but the mapping doesn't seem to be a problem.

[Change operationId prop type to string on /payments endpoint](https://app.asana.com/0/1202672669313222/1207811711125272/f)

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/data/payments.go
+++ b/internal/data/payments.go
@@ -15,9 +15,9 @@ type PaymentModel struct {
 }
 
 type Payment struct {
-	OperationID     int64     `db:"operation_id" json:"operationId"`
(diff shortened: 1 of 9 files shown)


gsoarescosta (approved):
LGTM!

PR #33, opened 2024-08-07
signing: KMS signature client

CaioTeixeira95:
### What

This PR adds the AWS KMS support for managing the Distribution Account Private Key. We encrypt the Private Key using KMS and store it in the database. Then, when a signing is needed, we decrypt using KMS.

### Why

Security request/improvement.

### Known limitations

N/A

#### PR Structure

#### Thoroughness

#### Release


cmd/utils/global_options.go
```diff
@@ -66,23 +67,76 @@ func HorizonClientURLOption(configKey *string) *config.ConfigOption {
...
 		OptType:        types.String,
 		CustomSetValue: SetConfigOptionStellarPrivateKey,
 		ConfigKey:      configKey,
-		Required:       true,
+		Required:       false,
 	}
 }
 
-func ChannelAccountEncryptionPassphraseOption(configKey *string) *config.ConfigOption {
+func DistributionAccountSignatureClientProviderOption(configKey *signing.SignatureClientType) *config.ConfigOption {
```
daniel-burghardt, line 102: Shall we standardize the config and function names?
```suggestion
func DistributionAccountSignatureProviderOption(configKey *signing.SignatureClientType) *config.ConfigOption {
```

cmd/utils/utils.go
```diff
@@ -18,3 +19,11 @@ func DefaultPersistentPreRunE(cfgOpts config.ConfigOptions) func(_ *cobra.Comman
 		return nil
 	}
 }
+
+func SignatureClientResolver(signatureClientOpts *signing.SignatureClientOptions) (signing.SignatureClient, error) {
+	signatureClient, err := signing.NewSignatureClient(signatureClientOpts)
```
daniel-burghardt, line 24: I think it might make sense to just move the logic from `signing.NewSignatureClient` to this function, as that seems to be actually resolving the `singing.SignatureClient`. Wdyt?

internal/signing/kms_signature_client.go
```diff
@@ -0,0 +1,138 @@
+package signing
```
daniel-burghardt, line 1: Wdyt about moving this to the `awskms` package so that we have things very well separated? Do you think that'd cause some import cycles?
Maybe we could have three subpackages in `signing`: `kms`, `env`, `channelaccount`

internal/signing/kms_signature_client.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	if len(stellarAccounts) == 0 {
+		return nil, fmt.Errorf("stellar accounts cannot be empty in %T", sc)
+	}
+
+	// Ensure that the distribution account is the only account signing the transaction
+	for _, stellarAccount := range stellarAccounts {
+		if stellarAccount != sc.distributionAccountPublicKey {
+			return nil, fmt.Errorf("stellar account %s is not allowed to sign in %T", stellarAccount, sc)
```
daniel-burghardt, line 82:
```suggestion
			return nil, fmt.Errorf("stellar account %s is not allowed to sign %T", stellarAccount, sc)
```

internal/signing/kms_signature_client.go
```diff
@@ -0,0 +1,138 @@
...
+		EncryptionContext: awskms.GetPrivateKeyEncryptionContext(kp.PublicKey),
+		KeyId:             &sc.kmsKeyARN,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("decrypting distribution account private key in %T: %w", sc, err)
+	}
+
+	kpFull, err := keypair.ParseFull(string(output.Plaintext))
+	if err != nil {
+		return nil, fmt.Errorf("parsing distribution account private key in %T: %w", sc, err)
```
daniel-burghardt, line 134: The `err` here would not leak the seed, right?

internal/signing/store/channel_accounts_model.go
```diff
@@ -19,23 +19,6 @@ var (
 	ErrChannelAccountNotFound        = errors.New("channel account not found")
 )
```
daniel-burghardt, line 21: How about we just move the model to the `data` package alongside the other models?

daniel-burghardt (approved):
This is great! Thanks @CaioTeixeira95! 🚀 
No blockers but raising some packages restructuring ideas for us to discuss.

internal/signing/kms_signature_client.go
CaioTeixeira95: Let's keep it simple. Having it separate in multiple packages would make it harder to find the implementations, having everything on the same package makes it easier to find and debug whenever necessary.

internal/signing/kms_signature_client.go
CaioTeixeira95: This error comes from the Stellar Go Library and looks like it won't return the key on the error.

internal/signing/store/channel_accounts_model.go
CaioTeixeira95: Here it's a different concept from the regular models that we've on the `data` package. This is the default implementation of the `Store` interface. This coincidentally uses the database, but that's not a rule, this could be stored anywhere, and that's why we have this model separated from the others.

daniel-burghardt: Makes sense 👍

daniel-burghardt (approved):
Nice! 🚀

PR #34, opened 2024-08-11
apptracker: Error tracking interface and it's Sentry implementation

gouthamp-stellar:
### What

Error tracking interface and it's Sentry implementation

Task: https://app.asana.com/0/1207947010297920/1207245825676510

### Why

Adding sentry tracking to the Wallet Backend 

### Known limitations

N/A

https://app.asana.com/0/1207947010297920/1207245825676510

#### PR Structure

#### Thoroughness

#### Release


cmd/ingest.go
```diff
@@ -62,12 +65,18 @@ func (c *ingestCmd) Command() *cobra.Command {
 			FlagDefault: 0,
 			Required:    false,
 		},
+		//SentryOps()
```
gsoarescosta, line 68: perhaps this was forgotten? 😅 
```suggestion
```

gsoarescosta (commented):
I see a few errors in our CI, maybe you can run a `go mod tidy` locally to check it;

also I see that we are using [golangci-lint](https://github.com/golangci/golangci-lint) in our CI too, maybe you can install it and run `golangci-lint run -v` to see if there are any lint errors locally 😄

internal/apptracker/sentry/sentry_tracker.go
```diff
@@ -0,0 +1,31 @@
...
+	if err := sentry.Init(sentry.ClientOptions{
+		Dsn:         dsn,
+		Environment: stellarEnv,
+	}); err != nil {
+		return err
+	}
+	defer sentry.Flush(time.Second * time.Duration(s.FlushFreq))
+	defer sentry.Recover()
+	return nil
+}
```
CaioTeixeira95, line 31: Why do we need this method? Can't we create a `NewSentryAppTracker` constructor that receives the DNS, environment, and Flush Frequency?

internal/apptracker/sentry/sentry_tracker.go
```diff
@@ -0,0 +1,31 @@
...
+
+func (s *SentryTracker) InitTracker(dsn string, stellarEnv string) error {
+	if err := sentry.Init(sentry.ClientOptions{
+		Dsn:         dsn,
+		Environment: stellarEnv,
+	}); err != nil {
+		return err
+	}
+	defer sentry.Flush(time.Second * time.Duration(s.FlushFreq))
+	defer sentry.Recover()
```
CaioTeixeira95, line 29: Shouldn't this be run every `CaptureMessage` or `CaptureException` call?

internal/ingest/ingest.go
```diff
@@ -25,6 +26,9 @@ type Configs struct {
 	StartLedger          int
 	EndLedger            int
 	LogLevel             logrus.Level
+	AppTracker           sentry.SentryTracker
+	SentryDsn            string
+	StellarEnvironment   string
```
CaioTeixeira95, line 31: I'd suggest we only have the app tracker here. Then on the config we only use a local variable:

```go
var sentryDSN string

utils.SentryOptions(&sentryDSN)
```

internal/serve/httperror/errors.go
```diff
@@ -60,9 +61,9 @@ func Unauthorized(message string, extras map[string]interface{}) *ErrorResponse
 	}
 }
 
-func InternalServerError(ctx context.Context, message string, err error, extras map[string]interface{}) *ErrorResponse {
-	// TODO: track error in Sentry
+func InternalServerError(ctx context.Context, message string, err error, extras map[string]interface{}, appTracker apptracker.AppTracker) *ErrorResponse {
 	log.Ctx(ctx).Error(err)
+	appTracker.CaptureException(err)
```
CaioTeixeira95, line 66: Here we should check if the AppTracker is nil.

internal/serve/httperror/errors_test.go
```diff
@@ -8,18 +8,22 @@ import (
...
 
+	"github.com/stellar/wallet-backend/internal/apptracker"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 )
 
 func TestErrorResponseRender(t *testing.T) {
+	appTrackerMock := apptracker.MockAppTracker{}
+	appTrackerMock.On("CaptureException", mock.Anything).Return(nil)
```
CaioTeixeira95, line 19: You are not testing the code here. It would be best if you did something like:

```go
appTrackerMock.On("CaptureException", errors.New("Some error"))
defer appTrackerMock.AssertExpectations(t)
```

I'm pretty sure that we can omit the `Return` since the capture exception doesn't return errors. Additionally, we should call the `AssertExpectations` method.

Then we should change the test a little:

```go
in: *InternalServerError(context.Background(), "", errors.New("Some error"), nil, &appTrackerMock),
```

internal/serve/httperror/errors_test.go
```diff
@@ -54,12 +58,14 @@ func TestErrorResponseRender(t *testing.T) {
 }
 
 func TestErrorHandler(t *testing.T) {
+	appTrackerMock := apptracker.MockAppTracker{}
+	appTrackerMock.On("CaptureException", mock.Anything).Return(nil)
```
CaioTeixeira95, line 62: Same here.

internal/serve/middleware/middleware.go
```diff
@@ -50,7 +51,7 @@ func SignatureMiddleware(signatureVerifier auth.SignatureVerifier) func(next htt
 }
 
 // RecoverHandler is a middleware that recovers from panics and logs the error.
-func RecoverHandler(next http.Handler) http.Handler {
+func RecoverHandler(next http.Handler, appTracker apptracker.AppTracker) http.Handler {
```
CaioTeixeira95, line 54: Here instead of creating a Wrapper, we can add one more layer to the function.

```suggestion
func RecoverHandler(appTracker apptracker.AppTracker) func(http.Handler) http.Handler {
        return func(next http.Handler) http.Handler {
                return http.HandlerFunc(.....
        }
```

internal/serve/serve.go
```diff
@@ -56,6 +58,11 @@ type Configs struct {
 	HorizonClientURL                   string
 	DistributionAccountSignatureClient signing.SignatureClient
 	ChannelAccountSignatureClient      signing.SignatureClient
+
+	// Error Tracker
+	AppTracker         sentry.SentryTracker
+	SentryDsn          string
+	StellarEnvironment string
```
CaioTeixeira95, line 65: Here is the same comment I've made for the Ingestion configs, we can use local variables for the sentry DSN.

internal/apptracker/app_tracker.go
```diff
@@ -0,0 +1,6 @@
+package apptracker
+
+type AppTracker interface {
+	CaptureMessage(message string)
+	CaptureException(exception error)
```
CaioTeixeira95, line 5: Can we also add the context here?

```suggestion
	CaptureMessage(ctx context.Context, message string)
	CaptureException(ctx context.Context, exception error)
```

internal/serve/middleware/middleware_test.go
```diff
@@ -19,11 +20,12 @@ import (
 
 func TestSignatureMiddleware(t *testing.T) {
 	signatureVerifierMock := auth.MockSignatureVerifier{}
+	appTrackerMock := apptracker.MockAppTracker{}
```
CaioTeixeira95, line 23: Please can you move this line below the `defer signatureVerifierMock.AssertExpectations(t)`.

```suggestion
	appTrackerMock := apptracker.MockAppTracker{}
	defer appTrackerMock.AssertExpectations(t)
```

internal/serve/middleware/middleware_test.go
```diff
@@ -151,6 +154,7 @@ func TestRecoverHandler(t *testing.T) {
 	req, err := http.NewRequest("GET", "/", nil)
 	require.NoError(t, err)
 	rr := httptest.NewRecorder()
+	appTrackerMock.On("CaptureException", mock.Anything).Return(nil)
```
CaioTeixeira95, line 157: The same suggestion that I've made on the `InternalServerError`, can we test the error value here?

internal/serve/serve.go
```diff
@@ -141,6 +150,7 @@ func initHandlerDeps(cfg Configs) (handlerDeps, error) {
 		SignatureVerifier:         signatureVerifier,
 		SupportedAssets:           cfg.SupportedAssets,
 		AccountSponsorshipService: accountSponsorshipService,
+		AppTracker:                &cfg.AppTracker,
```
CaioTeixeira95, line 153: Here you don't need to pass the reference, because the interface is already a reference.

```suggestion
		AppTracker:                cfg.AppTracker,
```

internal/apptracker/sentry/sentry_tracker.go
```diff
@@ -0,0 +1,31 @@
...
+	FlushFreq int64
+}
+
+func (s *SentryTracker) CaptureMessage(message string) {
+	sentry.CaptureMessage(message)
+}
+
+func (s *SentryTracker) CaptureException(exception error) {
+	sentry.CaptureException(exception)
+}
```
CaioTeixeira95, line 19: Can we add tests for these methods?

cmd/ingest.go
```diff
@@ -65,9 +68,17 @@ func (c *ingestCmd) Command() *cobra.Command {
 	}
 
 	cmd := &cobra.Command{
-		Use:               "ingest",
-		Short:             "Run Ingestion service",
-		PersistentPreRunE: utils.DefaultPersistentPreRunE(cfgOpts),
+		Use:   "ingest",
+		Short: "Run Ingestion service",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			utils.DefaultPersistentPreRunE(cfgOpts)
```
CaioTeixeira95, line 74: This won't work, the `utils.DefaultPersistentPreRunE` will only return a function, and not call it. I'd suggest you look at this [example](https://github.com/stellar/wallet-backend/pull/33/files#diff-87123721308af5d4a9f1dac03f5438ba1b3b5adf3e4bbf12666587666953a5d6R53-R78) to see how we implement the `PersistentPreRunE`.

CaioTeixeira95 (commented):
Thank you for working on this. I've left a few comments.

cmd/utils/global_options.go
```diff
@@ -86,3 +86,23 @@ func ChannelAccountEncryptionPassphraseOption(configKey *string) *config.ConfigO
...
+	return &config.ConfigOption{
+		Name:      "tracker-dsn",
+		Usage:     "The Sentry DSN",
+		OptType:   types.String,
+		ConfigKey: configKey,
+		Required:  true,
+	}
+}
+
+func StellarEnvironmentOption(configKey *string) *config.ConfigOption {
```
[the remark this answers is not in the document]
gouthamp-stellar: Sentry needs a variable called Environment when you call sentry.Init(...) and this basically sets the value for that

internal/apptracker/app_tracker.go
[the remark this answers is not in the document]
gouthamp-stellar: makes sense. done

cmd/utils/global_options.go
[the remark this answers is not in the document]
gouthamp-stellar: It is used to set up the sentry environment which is a variable we need to initialize the sentry tracker

internal/apptracker/app_tracker.go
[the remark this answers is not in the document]
gouthamp-stellar: Done

internal/apptracker/sentry/sentry_tracker.go
gouthamp-stellar: Done

internal/apptracker/sentry/sentry_tracker.go
gouthamp-stellar: As we discussed, this should not be called on every CaptureMessage

internal/ingest/ingest.go
gouthamp-stellar: Done

internal/serve/httperror/errors.go
gouthamp-stellar: Done

internal/serve/httperror/errors_test.go
gouthamp-stellar: Done

internal/serve/httperror/errors_test.go
gouthamp-stellar: Done

internal/serve/middleware/middleware.go
gouthamp-stellar: Done

internal/serve/serve.go
gouthamp-stellar: Done

internal/apptracker/app_tracker.go
gouthamp-stellar: Done

internal/serve/middleware/middleware_test.go
gouthamp-stellar: Done

internal/serve/middleware/middleware_test.go
gouthamp-stellar: Done

internal/serve/serve.go
gouthamp-stellar: I need it because the receiver is a pointer (take a look at sentry_tracker.go)

internal/apptracker/sentry/sentry_tracker.go
gouthamp-stellar: Done

cmd/ingest.go
gouthamp-stellar: Done

cmd/utils/global_options.go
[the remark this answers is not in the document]
CaioTeixeira95: I think we can use the `Network Passphrase` for that, what do you think?

cmd/ingest.go
gouthamp-stellar: Done

cmd/utils/global_options.go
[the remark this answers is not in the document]
gouthamp-stellar: Seems like we use the sentry environment variable to initialize the tracker in the Vibrant backend so I did the same here

cmd/ingest.go
```diff
@@ -65,9 +70,22 @@ func (c *ingestCmd) Command() *cobra.Command {
...
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if err := cfgOpts.RequireE(); err != nil {
+				return fmt.Errorf("requiring values of config options: %w", err)
+			}
+			if err := cfgOpts.SetValues(); err != nil {
+				return fmt.Errorf("setting values of config options: %w", err)
+			}
+			appTracker, err := sentry.NewSentryTracker(sentryDSN, stellarEnvironment, 5)
+			if err != nil {
+				log.Fatalf("Error initializing App Tracker: %s", err.Error())
```
CaioTeixeira95, line 84:
```suggestion
				return fmt.Errorf("initializing app tracker: %w", err)
```

cmd/ingest.go
```diff
@@ -65,9 +70,22 @@ func (c *ingestCmd) Command() *cobra.Command {
...
+			}
+			cfg.AppTracker = *appTracker
```
CaioTeixeira95, line 86: Here you should pass the reference because it is an interface!

cmd/serve.go
```diff
@@ -105,6 +110,11 @@ func (c *serveCmd) Command() *cobra.Command {
 				return fmt.Errorf("resolving distribution account signature client: %w", err)
 			}
 			cfg.DistributionAccountSignatureClient = signatureClient
+			appTracker, err := sentry.NewSentryTracker(sentryDSN, stellarEnvironment, 5)
+			if err != nil {
+				return fmt.Errorf("initializing App Tracker: %w", err)
+			}
+			cfg.AppTracker = *appTracker
```
CaioTeixeira95, line 117: Here you should pass the reference because it's an interface!

internal/apptracker/sentry/sentry_tracker.go
```diff
@@ -0,0 +1,42 @@
...
+var (
+	captureMessageFunc   = sentry.CaptureMessage
+	captureExceptionFunc = sentry.CaptureException
+	InitFunc             = sentry.Init
+	FlushFunc            = sentry.Flush
+	RecoverFunc          = sentry.Recover
+)
+
+type SentryTracker struct {
+	FlushFreq int64
```
CaioTeixeira95, line 20: Why do we need this value, it seems it's not being used. Shall we remove it?

internal/ingest/ingest.go
```diff
@@ -25,6 +26,7 @@ type Configs struct {
 	StartLedger          int
 	EndLedger            int
 	LogLevel             logrus.Level
+	AppTracker           sentry.SentryTracker
```
CaioTeixeira95, line 29: Here you should use the interface!

```suggestion
	AppTracker           apptracker.AppTracker
```

internal/serve/serve.go
```diff
@@ -57,6 +59,9 @@ type Configs struct {
 	HorizonClientURL                   string
 	DistributionAccountSignatureClient signing.SignatureClient
 	ChannelAccountSignatureClient      signing.SignatureClient
+
+	// Error Tracker
+	AppTracker sentry.SentryTracker
```
CaioTeixeira95, line 64: Here you should use the `apptracker.AppTracker`

CaioTeixeira95 (approved):
Overall, it looks good to me. Please solve the remaining commentaries before merging it.

internal/apptracker/sentry/sentry_tracker.go
gsoarescosta: that's a good point, where are we using it?

cmd/ingest.go
gouthamp-stellar: Done

internal/apptracker/sentry/sentry_tracker.go
gouthamp-stellar: good point, removed

cmd/serve.go
gouthamp-stellar: I made it
cfg.AppTracker = appTracker

cmd/ingest.go
gouthamp-stellar: I made it
cfg.AppTracker = appTracker

internal/ingest/ingest.go
gouthamp-stellar: Done

internal/serve/serve.go
gouthamp-stellar: Done

cmd/utils/global_options.go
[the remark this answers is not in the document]
daniel-burghardt: @gouthamp-stellar this is something we're refactoring on the Vibrant backend. Unfortunately we have this duplicate ENV there.
Wdyt about using the Network Passphrase env that we already use for ingestion (like @CaioTeixeira95 suggested) and mapping it to `testnet/development` or `mainnet/production` in the `NewSentryTracker` function.
That way we don't have this duplicate config which at the end of the day means the same thing.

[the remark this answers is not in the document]
daniel-burghardt: If not maybe we could name the env TRACKER_ENVIRONMENT so it's not misused in other places.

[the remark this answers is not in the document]
gouthamp-stellar: Down to do that in a future pr. I already pushed a pr to kube-vibrant with this env variable. Since it's only a couple lines, I can add this change to an upcoming pr

PR #36, opened 2024-08-30
tss: Tss tables and the channel interface

gouthamp-stellar:
### What

The transactions and tries tables for TSS and the interface that the worker pools will implement, as described in this document:
https://docs.google.com/document/d/1xcX86-w8lwT_60flCuYBK9X9co-108-5X6D9e5epw0U/edit?usp=sharing

### Why

Laying the foundation for upcoming TSS work

### Known limitations

N/A

https://app.asana.com/0/1207947010297920/1208138733594296

#### PR Structure

#### Thoroughness

#### Release


internal/tss/channel.go
```diff
@@ -0,0 +1,34 @@
+package tss
+
+type RPCGetIngestTxResponse struct {
+	// a status that indicated whether this transaction failed or successly made it to the ledger
+	Status string
+	// the raw TransactionEnvelope XDR for this transaction
+	EnvelopeXdr string
```
daniel-burghardt, line 7: I think we should try to stick to [Go's styleguide](https://google.github.io/styleguide/go/decisions#initialisms) as much as possible
```suggestion
	EnvelopeXDR string
```

internal/tss/channel.go
```diff
@@ -0,0 +1,34 @@
...
+	// the raw TransactionResult XDR of the envelopeXdr
+	ResultXdr string
```
daniel-burghardt, line 9:
```suggestion
	ResultXDR string
```

internal/tss/channel.go
```diff
@@ -0,0 +1,34 @@
+package tss
+
+type RPCGetIngestTxResponse struct {
+	// a status that indicated whether this transaction failed or successly made it to the ledger
+	Status string
+	// the raw TransactionEnvelope XDR for this transaction
+	EnvelopeXdr string
+	// the raw TransactionResult XDR of the envelopeXdr
```
daniel-burghardt, line 8: Let's try to stay consistent with the formatting across the codebase
```suggestion
	// The raw TransactionResult XDR of the envelopeXdr
```

internal/tss/channel.go
```diff
@@ -0,0 +1,34 @@
...
+	ResultXdr string
+	// The unix timestamp of when the transaction was included in the ledger
+	CreatedAt int64
+}
+
+type RPCSendTxResponse struct {
+	// the status of an RPC sendTransaction call. Can be one of [PENDING, DUPLICATE, TRY_AGAIN_LATER, ERROR]
+	Status string
```
daniel-burghardt, line 16:
```suggestion
	Status RPCTXStatus
```

When working with string enums it's best to define a type for it:
```
type RPCTXStatus string

const (
	PendingStatus       RPCTXStatus = "PENDING"
	DuplicateStatus     RPCTXStatus = "DUPLICATE"
	TryAgainLaterStatus RPCTXStatus = "TRY_AGAIN_LATER"
	ErrorStatus         RPCTXStatus = "ERROR"
)
```
Then you can also define a validation method for its values
```
func (s RPCTXStatus) IsValid() bool {
	switch s {
	case
		PendingStatus,
		DuplicateStatus,
		TryAgainLaterStatus,
		ErrorStatus:
		return true
	default:
		return false
	}
}
```

internal/db/migrations/2024-08-28.0-tss_transactions.sql
```diff
@@ -0,0 +1,37 @@
+-- +migrate Up
+
+CREATE TABLE tss_transactions (
+    transaction_hash VARCHAR(70) PRIMARY KEY,
+    transaction_xdr VARCHAR(400),
+    webhook_url VARCHAR(250),
+    current_status VARCHAR(50),
+    creation_time TIMESTAMPTZ DEFAULT NOW(),
+    last_updated_time TIMESTAMPTZ DEFAULT NOW(),
```
daniel-burghardt, line 9: Let's be consistent in all tables
```suggestion
    created_at TIMESTAMPTZ DEFAULT NOW(),
    updated_at TIMESTAMPTZ DEFAULT NOW(),
```

internal/db/migrations/2024-08-28.0-tss_transactions.sql
```diff
@@ -0,0 +1,37 @@
+-- +migrate Up
+
+CREATE TABLE tss_transactions (
+    transaction_hash VARCHAR(70) PRIMARY KEY,
+    transaction_xdr VARCHAR(400),
```
daniel-burghardt, line 5: What is the 400 based on?
I just tried creating a simple transaction with 2 operations and it resulted in a 512 characters long string.
```
AAAAAgAAAAD54NA+ZF4ieMZuoyyAsDJSH8EEylTn08xEecwNABU97AAAAMgAAAMdFHk3/wAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAxzZGZzZGFmZGFzZmEAAAACAAAAAQAAAAD54NA+ZF4ieMZuoyyAsDJSH8EEylTn08xEecwNABU97AAAAAEAAAAA+eDQPmReInjGbqMsgLAyUh/BBMpU59PMRHnMDQAVPewAAAABQVNEQQAAAAD54NA+ZF4ieMZuoyyAsDJSH8EEylTn08xEecwNABU97AAAAiGT/OIAAAAAAQAAAAD54NA+ZF4ieMZuoyyAsDJSH8EEylTn08xEecwNABU97AAAAA0AAAABU0RTRAAAAAD54NA+ZF4ieMZuoyyAsDJSH8EEylTn08xEecwNABU97AAAAAAUQ/0AAAAAAPng0D5kXiJ4xm6jLICwMlIfwQTKVOfTzER5zA0AFT3sAAAAAAAAAAyy1lsAAAAAAAAAAAAAAAAA
```

I think we should stick to `TEXT` in these VARCHAR columns as there is no real need to limit them imo. See [Postgres Wiki](https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_varchar.28n.29_by_default).

internal/db/migrations/2024-08-28.0-tss_transactions.sql
```diff
@@ -0,0 +1,37 @@
...
+    webhook_url VARCHAR(250),
```
daniel-burghardt, line 6: Weren't we going to store the webhook URL on the application level since we expect it would generally be the same for all transactions?

gouthamp-stellar: True, I know that at first we figured they'd be a 1:1 correspondence between client and TSS, but I thought about this and I do think it is possible that the client may want to have more than one webhook. For example, they may only have one instance of TSS/wallet backend running for multiple of their applications, each one of which will have a different webhook. They may also want to have different webhooks for testing vs production

internal/tss/channel.go
gouthamp-stellar: Done

internal/tss/channel.go
gouthamp-stellar: Sounds good

internal/tss/channel.go
gouthamp-stellar: I like it, added statuses for both sendTransaction and getTransaction RPC calls

internal/db/migrations/2024-08-28.0-tss_transactions.sql
gouthamp-stellar: Yeap, changed all varchar fields to TEXT

internal/db/migrations/2024-08-28.0-tss_transactions.sql
daniel-burghardt: > I do think it is possible that the client may want to have more than one webhook

Yeah I think that is possible indeed. How would the clients specify the webhook URL in this case (that would feed this column in the DB)?

> They may also want to have different webhooks for testing vs production

That I don't think would be possible because the wallet backend is very environment specific (e.g.: your distribution account will only exist in a single network, the RPC will only be connected to a single network, etc.)

gouthamp-stellar: There's 2 ways I can think of in which the clients can specify their webhook url:

1. We have an explicit endpoint that they can call to register their endpoint
2. They send their webhook endpoint along with their transactions

I like 2 better, but open to other ideas

daniel-burghardt: The downsides I see with 2 is that it results in a lot of duplicate data in the `tss_transactions` table and makes the process of submitting transactions a little less simple.
To me this feels like project level type of setting that could be set once and the client wouldn't have to worry about it anymore. Given that I think I'd lean to the 1. side. Wdyt?

gouthamp-stellar: If we go with 1. I picture it looking something like this:

1. we will have a /register-webhook endpoint for the clients to call where they provide a client_id (string) and a webhook url (string)
2. we will have a webhooks table that will store the client id along with the webhook url
3. Now with every transaction submission request, the client will send their client-id instead of the webhook url
4. So we will have to add a client id column to the transactions table so we know which client to send the transaction result to

Alternatively, we could also generate the client id ourselves (could just an auto incrementing id) when the client calls the /register-webhook endpoint and just have the client give us that client id for transaction submission requests

Thing is, we will still have that client id column in the transactions table which will have a lot of duplicate data

Thoughts?

gouthamp-stellar: I like the idea of us generating the client id on the /register-webhook endpoint better

gouthamp-stellar: Our slack discussion summarized below:

Decides to keep the table design as is because it was decided that a client can have more than 1 webhook url. We will have the client pass in the webhook url with every transaction submission request

internal/db/migrations/2024-08-28.0-tss_transactions.sql
```diff
@@ -0,0 +1,35 @@
+-- +migrate Up
+
+CREATE TABLE tss_transactions (
+    transaction_hash TEXT PRIMARY KEY,
+    transaction_xdr TEXT,
+    webhook_url TEXT,
+    current_status TEXT,
+    created_at TIMESTAMPTZ DEFAULT NOW(),
+    updated_at TIMESTAMPTZ DEFAULT NOW(),
+    claimed_until TIMESTAMPTZ
```
daniel-burghardt, line 10: Shall we add NOT NULL constraints on the `transaction_hash` (tho it's PK), `transaction_xdr`, `current_status`, `created_at` and `updated_at` columns?

internal/db/migrations/2024-08-28.0-tss_transactions.sql
```diff
@@ -0,0 +1,35 @@
...
+);
+
+CREATE TABLE tss_transaction_submission_tries (
+    try_transaction_hash TEXT PRIMARY KEY,
+    original_transaction_hash TEXT,
+    try_transaction_xdr TEXT,
+    status INTEGER,
+    updated_at TIMESTAMPTZ DEFAULT NOW()
```
daniel-burghardt, line 18: Here I guess all columns would be required, right?

internal/tss/types.go
```diff
@@ -0,0 +1,76 @@
...
+
+type RPCSendTxResponse struct {
+	// The hash of the transaction submitted to RPC
+	TransactionHash string
+	TransactionXDR  string
+	// The status of an RPC sendTransaction call. Can be one of [PENDING, DUPLICATE, TRY_AGAIN_LATER, ERROR]
+	Status RPCTXStatus
+	// The (optional) error code that is derived by deserialzing the errorResultXdr string in the sendTransaction response
+	// list of possible errror codes: https://developers.stellar.org/docs/data/horizon/api-reference/errors/result-codes/transactions
+	Code RPCTXCode
```
daniel-burghardt, line 57: The code in the RPC response is a struct?

daniel-burghardt (approved):
Let's please just add the NOT NULL constraints on the required columns before merging.

internal/db/migrations/2024-08-28.0-tss_transactions.sql
gouthamp-stellar: PK columns are non null by default but will add NOT NULL to others

internal/tss/types.go
gouthamp-stellar: Yes. Thats because I want to encapsulate both RPC tx codes and add a couple of our own

gouthamp-stellar:
What
The transactions and tries tables for TSS and the interface that the worker pools will implement, as described in this document:
https://docs.google.com/document/d/1xcX86-w8lwT_60flCuYBK9X9co-108-5X6D9e5epw0U/edit?usp=sharing

Why
Laying the foundation for upcoming TSS work

internal/db/migrations/2024-08-28.0-tss_transactions.sql
gouthamp-stellar: That's right

PR #40, opened 2024-09-15
all: TSS RPC Caller Service

gouthamp-stellar:
### What

This is TSS's RPC Caller Service as described in this doc: 
https://docs.google.com/document/d/1xcX86-w8lwT_60flCuYBK9X9co-108-5X6D9e5epw0U/edit#heading=h.76emp0zcbvdy

### Why

This service is responsible for handling incoming transaction submission requests from clients and calling RPC (for the very first time) to submit those transactions to the network 

### Known limitations

N/A

https://app.asana.com/0/1207947010297920/1208138733594316

#### PR Structure

#### Thoroughness

#### Release


internal/tss/channels/rpc_caller_service_channel.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		log.Errorf("RPCCallerService: Unable to upsert transaction into transactions table: %s", err.Error())
+		return
+	}
+	/*
+		The reason we return on each error we encounter is so that the transaction status
+		stays at NEW, so that it does not progress any further and
+		can be picked up for re-processing when this pool is restarted.
+	*/
+	feeBumpTx, err := p.txService.SignAndBuildNewTransaction(payload.TransactionXDR)
```
gouthamp-stellar, line 55: note: make this SignAndBuildNewFeeBumpTx...

internal/tss/channels/utils.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return tss.RPCSendTxResponse{}, fmt.Errorf("%s: Unable to upsert try in tries table: %s", channelName, err.Error())
+	}
+	rpcSendResp, rpcErr := txService.SendTransaction(feeBumpTxXDR)
+
+	err = store.UpsertTry(ctx, payload.TransactionHash, feeBumpTxHash, feeBumpTxXDR, rpcSendResp.Code)
+	if err != nil {
+		return tss.RPCSendTxResponse{}, fmt.Errorf("%s: Unable to upsert try in tries table: %s", channelName, err.Error())
+	}
+	if rpcErr != nil && rpcSendResp.Code.OtherCodes == tss.RPCFailCode || rpcSendResp.Code.OtherCodes == tss.UnMarshalBinaryCode {
```
gouthamp-stellar, line 37: should this be done before the 1st UpsertTry?

internal/tss/services/rpc_caller_service.go
```diff
@@ -0,0 +1,21 @@
+package services
+
+import (
```
gouthamp-stellar, line 3: Maybe make channels and services one struct

internal/tss/channels/utils.go
gouthamp-stellar: We need to update the try with the rpcfailcode or unmarshallerrror code, which is why we call UpsertTry first before checking to see if there was an error

internal/tss/services/rpc_caller_service.go
gouthamp-stellar: Done. Removed services, now only have channels and the router

internal/entities/rpc.go
```diff
@@ -18,12 +18,22 @@ const (
 	SuccessStatus  RPCStatus = "SUCCESS"
 )
 
+type RPCEntry struct {
+	Key                   string `json:"key"`
+	XDR                   string `json:"xdr"`
+	LastModifiedLedgerSeq int64  `json:"lastModifiedLedgerSeq"`
+}
```
daniel-burghardt, line 25: These won't be used anymore, we can clean them up. I had deleted them from the RPC Service PR.

internal/serve/serve.go
```diff
@@ -74,7 +82,10 @@ type handlerDeps struct {
 	AccountService            services.AccountService
 	AccountSponsorshipService services.AccountSponsorshipService
 	PaymentService            services.PaymentService
-	AppTracker                apptracker.AppTracker
+	// TSS
+	RPCCallerServiceChannel tss.Channel
+	TSSRouter               tssrouter.Router
+	AppTracker              apptracker.AppTracker
```
daniel-burghardt, line 88: The AppTracker is not a TSS-specific component, right?

internal/serve/serve.go
```diff
@@ -155,6 +167,47 @@ func initHandlerDeps(cfg Configs) (handlerDeps, error) {
 	}
 	go ensureChannelAccounts(channelAccountService, int64(cfg.NumberOfChannelAccounts))
 
+	// TSS
+	txServiceOpts := tssservices.TransactionServiceOptions{
+		DistributionAccountSignatureClient: cfg.DistributionAccountSignatureClient,
+		ChannelAccountSignatureClient:      cfg.ChannelAccountSignatureClient,
+		HorizonClient:                      &horizonClient,
+		BaseFee:                            int64(cfg.BaseFee), // Reuse horizon base fee for RPC??
```
daniel-burghardt, line 175: I think that's fair 👍 
```suggestion
		BaseFee:                            int64(cfg.BaseFee),
```

internal/serve/serve.go
```diff
@@ -155,6 +167,47 @@ func initHandlerDeps(cfg Configs) (handlerDeps, error) {
...
+	if err != nil {
+		return handlerDeps{}, fmt.Errorf("instantiating tss transaction service: %w", err)
+	}
+	httpClient := http.Client{Timeout: time.Duration(30 * time.Second)}
+	rpcService, err := services.NewRPCService(cfg.RPCURL, &httpClient)
+	if err != nil {
+		return handlerDeps{}, fmt.Errorf("instantiating rpc service: %w", err)
+	}
+
+	// re-use same context as above??
```
daniel-burghardt, line 187: Which context?

internal/serve/serve.go
```diff
@@ -155,6 +167,47 @@ func initHandlerDeps(cfg Configs) (handlerDeps, error) {
...
+		RPCService: rpcService,
+		Store:      store,
+	})
+	tssChannelConfigs := tsschannel.RPCCallerChannelConfigs{
+		TxManager:     txManager,
+		Store:         store,
+		MaxBufferSize: cfg.RPCCallerServiceChannelBufferSize,
+		MaxWorkers:    cfg.RPCCallerServiceChannelMaxWorkers,
+	}
+	rpcCallerServiceChannel := tsschannel.NewRPCCallerChannel(tssChannelConfigs)
```
daniel-burghardt, line 200: Shall we initialize all services the same way for better readability? Same for the `NewTransactionService` above.
```suggestion
	rpcCallerServiceChannel := tsschannel.NewRPCCallerChannel(tsschannel.RPCCallerChannelConfigs{
		TxManager:     txManager,
		Store:         store,
		MaxBufferSize: cfg.RPCCallerServiceChannelBufferSize,
		MaxWorkers:    cfg.RPCCallerServiceChannelMaxWorkers,
	})
```

internal/serve/serve.go
```diff
@@ -155,6 +167,47 @@ func initHandlerDeps(cfg Configs) (handlerDeps, error) {
...
+
+	router := tssrouter.NewRouter(tssrouter.RouterConfigs{
```
daniel-burghardt, line 202: It's a good idea to check in the `NewRouter` and `NewStore` functions that the parameters are passed in correctly and are not `nil`. If a parameter is invalid, they can return an error.

internal/serve/serve.go
```diff
@@ -163,6 +216,9 @@ func initHandlerDeps(cfg Configs) (handlerDeps, error) {
 		AccountSponsorshipService: accountSponsorshipService,
 		PaymentService:            paymentService,
 		AppTracker:                cfg.AppTracker,
+		// TSS
+		RPCCallerServiceChannel: rpcCallerServiceChannel,
+		TSSRouter:               router,
```
daniel-burghardt, line 221: Which of the two will be the entrypoint when handling HTTP requests? Probably that one can be our only dependency to be injected in `handlerDeps`, right?

internal/services/servicesmocks/rpc_service_mocks.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func (r *RPCServiceMock) GetTransaction(transactionHash string) (entities.RPCGetTransactionResult, error) {
+	args := r.Called(transactionHash)
+	return args.Get(0).(entities.RPCGetTransactionResult), args.Error(1)
+}
+
+type TransactionManagerMock struct {
+	mock.Mock
+}
```
daniel-burghardt, line 27: Can this one be cleaned up? I see it's not being used

internal/tss/channels/rpc_caller_channel.go
```diff
@@ -0,0 +1,81 @@
...
+	err := p.Store.UpsertTransaction(ctx, payload.WebhookURL, payload.TransactionHash, payload.TransactionXDR, tss.RPCTXStatus{OtherStatus: tss.NewStatus})
+
+	if err != nil {
+		log.Errorf("%s: Unable to upsert transaction into transactions table: %e", RPCCallerChannelName, err)
+		return
+	}
+	rpcSendResp, err := p.TxManager.BuildAndSubmitTransaction(ctx, RPCCallerChannelName, payload)
+
+	if err != nil {
+		log.Errorf("%s: Unable to sign and submit transaction: %e", RPCCallerChannelName, err)
```
daniel-burghardt, line 63:
```suggestion
		log.Errorf("%s: unable to sign and submit transaction: %w", RPCCallerChannelName, err)
```

internal/tss/channels/rpc_caller_channel.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (p *rpcCallerPool) Receive(payload tss.Payload) {
+
+	ctx := context.Background()
+	// Create a new transaction record in the transactions table.
+	err := p.Store.UpsertTransaction(ctx, payload.WebhookURL, payload.TransactionHash, payload.TransactionXDR, tss.RPCTXStatus{OtherStatus: tss.NewStatus})
+
+	if err != nil {
+		log.Errorf("%s: Unable to upsert transaction into transactions table: %e", RPCCallerChannelName, err)
```
daniel-burghardt, line 57:
```suggestion
		log.Errorf("%s: unable to upsert transaction into transactions table: %w", RPCCallerChannelName, err)
```

internal/tss/channels/rpc_caller_channel.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if err != nil {
+		log.Errorf("%s: Unable to sign and submit transaction: %e", RPCCallerChannelName, err)
+		return
+	}
+	payload.RpcSubmitTxResponse = rpcSendResp
+	if rpcSendResp.Status.RPCStatus == entities.TryAgainLaterStatus || rpcSendResp.Status.RPCStatus == entities.ErrorStatus {
+		err = p.Router.Route(payload)
+		if err != nil {
+			log.Errorf("%s: Unable to route payload: %e", RPCCallerChannelName, err)
```
daniel-burghardt, line 70:
```suggestion
			log.Errorf("%s: unable to route payload: %w", RPCCallerChannelName, err)
```

internal/tss/channels/rpc_caller_channel.go
```diff
@@ -0,0 +1,81 @@
...
+		return
+	}
+	rpcSendResp, err := p.TxManager.BuildAndSubmitTransaction(ctx, RPCCallerChannelName, payload)
+
+	if err != nil {
+		log.Errorf("%s: Unable to sign and submit transaction: %e", RPCCallerChannelName, err)
+		return
+	}
+	payload.RpcSubmitTxResponse = rpcSendResp
+	if rpcSendResp.Status.RPCStatus == entities.TryAgainLaterStatus || rpcSendResp.Status.RPCStatus == entities.ErrorStatus {
```
daniel-burghardt, line 67: Would it make sense to keep any logic related to routing (e.g. `if`'s like this one) to be only inside the router?

internal/tss/router/router.go
```diff
@@ -0,0 +1,89 @@
...
+			// Do nothing for PENDING / DUPLICATE statuses
+			return nil
+		}
+	}
+	if channel == nil {
+		return fmt.Errorf("payload could not be routed - channel is nil")
+	}
+	channel.Send(payload)
+	return nil
+}
```
daniel-burghardt, line 89: 💡 No problem with the logic, just an idea for improved readability:

```suggestion
func (r *router) Route(payload tss.Payload) error {
	switch payload.RpcSubmitTxResponse.Status.Status() {
	case string(tss.NewStatus):
		r.RPCCallerChannel.Send(payload)

	case string(entities.TryAgainLaterStatus):
		r.ErrorJitterChannel.Send(payload)

	case string(entities.ErrorStatus):
		if payload.RpcSubmitTxResponse.Code.OtherCodes == tss.NoCode {
			if slices.Contains(JitterErrorCodes, payload.RpcSubmitTxResponse.Code.TxResultCode) {
				r.ErrorJitterChannel.Send(payload)
			} else if slices.Contains(NonJitterErrorCodes, payload.RpcSubmitTxResponse.Code.TxResultCode) {
				r.ErrorNonJitterChannel.Send(payload)
			} else if slices.Contains(FinalErrorCodes, payload.RpcSubmitTxResponse.Code.TxResultCode) {
				r.WebhookChannel.Send(payload)
			} else {
				return fmt.Errorf("payload could not be routed, status %+v", payload.RpcGetIngestTxResponse.Status)
			}
		}

	// Do nothing for PENDING / DUPLICATE statuses
	case string(entities.PendingStatus), string(entities.DuplicateStatus):
		return nil

	default:
		return fmt.Errorf("unexpected status %+v", payload.RpcGetIngestTxResponse.Status)
	}

	return nil
}
```

internal/tss/services/transaction_manager.go
```diff
@@ -0,0 +1,76 @@
...
+	err = t.Store.UpsertTry(ctx, payload.TransactionHash, feeBumpTxHash, feeBumpTxXDR, rpcSendResp.Code)
+	if err != nil {
+		return tss.RPCSendTxResponse{}, fmt.Errorf("%s: Unable to upsert try in tries table: %s", channelName, err.Error())
+	}
+
+	if parseErr != nil {
+		return rpcSendResp, fmt.Errorf("%s: RPC fail: %w", channelName, parseErr)
+	}
+
+	if rpcErr != nil && rpcSendResp.Code.OtherCodes == tss.RPCFailCode || rpcSendResp.Code.OtherCodes == tss.UnmarshalBinaryCode {
```
daniel-burghardt, line 67: Isn't the idea to use only the `parseErr`?
```suggestion
	if parseErr != nil && rpcSendResp.Code.OtherCodes == tss.RPCFailCode || rpcSendResp.Code.OtherCodes == tss.UnmarshalBinaryCode {
```

internal/tss/services/transaction_manager.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return tss.RPCSendTxResponse{}, fmt.Errorf("%s: Unable to upsert try in tries table: %w", channelName, err)
+	}
+	rpcResp, rpcErr := t.RPCService.SendTransaction(feeBumpTxXDR)
+	rpcSendResp, parseErr := tss.ParseToRPCSendTxResponse(feeBumpTxHash, rpcResp, rpcErr)
+
+	err = t.Store.UpsertTry(ctx, payload.TransactionHash, feeBumpTxHash, feeBumpTxXDR, rpcSendResp.Code)
+	if err != nil {
+		return tss.RPCSendTxResponse{}, fmt.Errorf("%s: Unable to upsert try in tries table: %s", channelName, err.Error())
+	}
```
daniel-burghardt, line 61: The `UpsertTry` should likely be called after checking the previous errors, right?

internal/tss/store/store.go
```diff
@@ -0,0 +1,66 @@
...
+	const q = `
+	INSERT INTO 
+		tss_transactions (transaction_hash, transaction_xdr, webhook_url, current_status)
+	VALUES
+		($1, $2, $3, $4)
+	ON CONFLICT (transaction_hash) 
+	DO UPDATE SET 
+		transaction_xdr = $2,
+		webhook_url = $3,
+    	current_status = $4,
```
daniel-burghardt, line 38: 💡 When handling ON CONFLICT's you can use Postgre's EXCLUDED keywork.
Btw, seems there's something up with the indentation here.
```suggestion
		transaction_xdr = EXCLUDED.transaction_xdr,
		webhook_url = EXCLUDED.webhook_url,
    	        current_status = EXCLUDED.current_status,
```

internal/entities/rpc.go
gouthamp-stellar: Yeap, removed these so you can add them later

internal/serve/serve.go
gouthamp-stellar: Changed this so the comment is more specific

internal/serve/serve.go
gouthamp-stellar: Fair

internal/serve/serve.go
gouthamp-stellar: Nvm, this is obselete

internal/tss/channels/rpc_caller_channel.go
gouthamp-stellar: Good point, I changed this

internal/tss/router/router.go
gouthamp-stellar: That is more readable, but the thing is, I want to be able to pass nil values for some/all channels. Because with the ingestor for example, it will only ever need the webhook channel. And if im passing nil values for these, I think I should explicitly check to see if the channels are nil before sending payloads to them

internal/tss/services/transaction_manager.go
gouthamp-stellar: Good catch!

internal/tss/services/transaction_manager.go
gouthamp-stellar: No, because I want UpsertTry to capture the error code returned by SendTransaction

internal/tss/store/store.go
gouthamp-stellar: Nice! made this change

internal/serve/serve.go
gouthamp-stellar: Yeap, these are all being initialized the same now

internal/serve/serve.go
gouthamp-stellar: I did add a nil check for NewStore, but kept NewRouter the way it is right now, for reasons I explained in another comment, I would like the option of some of these channels being nil

internal/serve/serve.go
gouthamp-stellar: We actually need both, because the channel is being closed (via Stop()) in the Serve function

gouthamp-stellar: But the entrypoint will be the router

internal/services/servicesmocks/rpc_service_mocks.go
gouthamp-stellar: Yeap removed

internal/tss/channels/rpc_caller_channel.go
gouthamp-stellar: Done

internal/tss/channels/rpc_caller_channel.go
gouthamp-stellar: Done

internal/tss/channels/rpc_caller_channel.go
gouthamp-stellar: Done

internal/serve/serve.go
daniel-burghardt: Hmm got it, maybe we can create an `OnStop` function on the router for this?

internal/tss/channels/rpc_caller_channel.go
daniel-burghardt: You forgot to fix the `%e`, this is what it's printing now: https://go.dev/play/p/wp-a11Jbr3C

daniel-burghardt: Same for the other two errors below

internal/tss/services/transaction_manager.go
```diff
@@ -64,7 +64,7 @@ func (t *transactionManager) BuildAndSubmitTransaction(ctx context.Context, chan
 		return rpcSendResp, fmt.Errorf("%s: RPC fail: %w", channelName, parseErr)
 	}
 
-	if rpcErr != nil && rpcSendResp.Code.OtherCodes == tss.RPCFailCode || rpcSendResp.Code.OtherCodes == tss.UnmarshalBinaryCode {
+	if parseErr != nil && rpcSendResp.Code.OtherCodes == tss.RPCFailCode || rpcSendResp.Code.OtherCodes == tss.UnmarshalBinaryCode {
```
daniel-burghardt, line 67: This case is impossible atm, isn't it? If `parseErr != nil` it would fall into the previous check.

daniel-burghardt (approved):
Just two small new comments: about the `%e` and the `parseErr != nil` check

PR #43, opened 2024-09-19
all: TSS Error handler service

gouthamp-stellar:
### What

This is TSS's Error Handler Service as described in this doc:
https://docs.google.com/document/d/1xcX86-w8lwT_60flCuYBK9X9co-108-5X6D9e5epw0U/edit#heading=h.76emp0zcbvdy

### Why

This service is responsible for retrying transactions till they reach a terminal state

### Known limitations

N/A

https://app.asana.com/0/1207947010297920/1208138733594321

#### PR Structure

#### Thoroughness

#### Release


cmd/utils/global_options.go
```diff
@@ -163,6 +131,128 @@ func DistributionAccountSignatureClientProviderOption(configKey *signing.Signatu
...
+		Name:        "rpc-url",
+		Usage:       "The URL of the RPC Server.",
+		OptType:     types.String,
+		ConfigKey:   configKey,
+		FlagDefault: "localhost:8080",
+		Required:    true,
+	}
+}
+
+func RPCCallerServiceChannelBufferSizeOption(configKey *int) *config.ConfigOption {
```
daniel-burghardt, line 145: Would these be considered "global options"? If they're only gonna be used in the `serve` command then I'd suggest we could move them to a separate file, `tss_options.go` perhaps? The `global_options.go` file was meant to hold the options that are shared across different commands.

internal/tss/channels/error_jitter_channel.go
```diff
@@ -0,0 +1,97 @@
...
+		if !slices.Contains(tss.JitterErrorCodes, rpcSendResp.Code.TxResultCode) {
+			err = p.Router.Route(payload)
+			if err != nil {
+				log.Errorf("%s: unable to route payload: %e", ErrorJitterChannelName, err)
+				return
+			}
+			return
+		}
+	}
+	if i == p.MaxRetries {
```
daniel-burghardt, line 81: Is this `if` necessary? Seems that will always be the case since there are no `break`'s in the loop, or am I missing something?

internal/tss/channels/error_jitter_channel.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (p *errorJitterPool) Receive(payload tss.Payload) {
+	ctx := context.Background()
+	var i int
+	for i = 0; i < p.MaxRetries; i++ {
+		currentBackoff := p.MinWaitBtwnRetriesMS * (1 << i)
+		time.Sleep(jitter(time.Duration(currentBackoff)) * time.Microsecond)
+		rpcSendResp, err := p.TxManager.BuildAndSubmitTransaction(ctx, ErrorJitterChannelName, payload)
+		if err != nil {
+			log.Errorf("%s: unable to sign and submit transaction: %e", ErrorJitterChannelName, err)
```
daniel-burghardt, line 68: Please watch out for the correct formatter in all errors, as mentioned in the previous PR.
```suggestion
			log.Errorf("%s: unable to sign and submit transaction: %w", ErrorJitterChannelName, err)
```

internal/tss/channels/error_jitter_channel.go
```diff
@@ -0,0 +1,97 @@
...
+		p.Receive(payload)
+	})
+}
+
+func (p *errorJitterPool) Receive(payload tss.Payload) {
+	ctx := context.Background()
+	var i int
+	for i = 0; i < p.MaxRetries; i++ {
+		currentBackoff := p.MinWaitBtwnRetriesMS * (1 << i)
+		time.Sleep(jitter(time.Duration(currentBackoff)) * time.Microsecond)
```
daniel-burghardt, line 65: Should this be Millisecond?
```suggestion
		time.Sleep(jitter(time.Duration(currentBackoff)) * time.Millisecond)
```

internal/tss/channels/error_jitter_channel.go
```diff
@@ -0,0 +1,97 @@
...
+			if err != nil {
+				log.Errorf("%s: unable to route payload: %e", ErrorJitterChannelName, err)
+				return
+			}
+			return
+		}
+	}
+	if i == p.MaxRetries {
+		// Retry limit reached, route the payload to the router so it can re-route it to this pool and keep re-trying
+		// NOTE: Is this a good idea? Infinite tries per transaction ?
```
daniel-burghardt, line 83: Wdyt about sending an error webhook request in this case instead of the infinite loop?
Given that there is the `max retries` env var as a client I wouldn't expect this to be retried indefinitely.

internal/tss/channels/error_non_jitter_channel.go
```diff
@@ -0,0 +1,92 @@
...
+	p.Pool.Submit(func() {
+		p.Receive(payload)
+	})
+}
+
+func (p *errorNonJitterPool) Receive(payload tss.Payload) {
+	ctx := context.Background()
+	var i int
+	for i = 0; i < p.MaxRetries; i++ {
+		time.Sleep(time.Duration(p.WaitBtwnRetriesMS) * time.Microsecond)
```
daniel-burghardt, line 59: Please see comments from the jitter channel, most would apply here too

internal/tss/channels/error_jitter_channel.go
```diff
@@ -0,0 +1,97 @@
...
+		err := p.Router.Route(payload)
```
daniel-burghardt, line 84: 💡  Might be good to include a log here too, informing that transaction X errored after max tries.

cmd/utils/global_options.go
gouthamp-stellar: That makes sense. I moved these to tss_options.go

internal/tss/channels/error_jitter_channel.go
gouthamp-stellar: Lol good catch! I removed the if statement

internal/tss/channels/error_jitter_channel.go
gouthamp-stellar: I actually tried to change this to a %w, but apparently it won't let me. I get the following error:
github.com/stellar/go/support/log.Errorf does not support error-wrapping directive %w

internal/tss/channels/error_jitter_channel.go
gouthamp-stellar: Changed this to millisecond

internal/tss/channels/error_jitter_channel.go
daniel-burghardt: Ah that's right, when working with log you can use `%v`.

internal/tss/channels/error_jitter_channel.go
gouthamp-stellar: So max_retries is really a variable for how many retries we want in a single task - since we do not want each task submitted to the go routine pool to run infinitely. But in theory, each transaction can be retried an infinite number of times. In this case, like the comment points out, the payload will get routed to the same go routine pool, but it will just be a brand new task. 
I'm allowing for infinite retries on transactions because that's the way SDP does it. If we find that this is becoming an issue in testing, we can cap it

internal/tss/channels/error_non_jitter_channel.go
gouthamp-stellar: Yeap made changes in both

internal/tss/channels/error_jitter_channel.go
gouthamp-stellar: added a log.info statement for when this limit is reached

internal/tss/channels/error_jitter_channel.go
daniel-burghardt: Got it. Does SDP also have this concept of "max retries"?

gouthamp-stellar: No, that is specific to this implementation. SDP retries transactions infinitely. They essentially have an infinite loop which picks up transactions that need to be processed from a table

internal/tss/channels/error_jitter_channel.go
gouthamp-stellar: Changed this to %v

internal/tss/channels/error_jitter_channel.go
daniel-burghardt: Do you think it would make sense for the error jitter and non jitter channels to simply sleep for the required time (based on the records of # of tries in the DB) and route the payload to the rpc caller service?
That way we could achieve the same, but would have the build and submission of TX's in a central place/pool.
And then the sleep time is always exponential (it never resets when it reached `max tries`.

daniel-burghardt (approved):

internal/tss/channels/error_jitter_channel.go
gouthamp-stellar: But the jitter and non jitter channel dont wait the same way. There is no/constant wait between retries in the non jitter channel and an exponential backoff in the jitter channel. And its not based on the # or records in the tries table - although that could be done that way, but i feel like that would be a little less clean

PR #45, opened 2024-09-25
services: rpc service

daniel-burghardt:
### What

Creates a new RPC service, extracting the logic from the TSS package into a more general one to be available and used across the application.

### Why

The `Ingest Service` for the new Balances Ingestion feature will also require RPC communication, therefore the RPC functions are being extracted into its own service. 

### Known limitations

N/A

[[Wallet Backend] Refactor RPC code into a separate service](https://app.asana.com/0/1202672669313222/1208409334251161/f)

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/services/ingest.go
+++ b/internal/services/ingest.go
@@ -29,8 +29,7 @@ type ingestService struct {
 	networkPassphrase string
 	ledgerCursorName  string
 	appTracker        apptracker.AppTracker
-	httpClient        utils.HTTPClient
-	rpcURL            string
+	rpcService        RPCService
 }
 
 func NewIngestService(
@@ -39,8 +38,7 @@ func NewIngestService(
 	networkPassphrase string,
 	ledgerCursorName string,
 	appTracker apptracker.AppTracker,
-	httpClient utils.HTTPClient,
-	rpcURL string,
+	rpcService RPCService,
 ) (*ingestService, error) {
 	if models == nil {
 		return nil, errors.New("models cannot be nil")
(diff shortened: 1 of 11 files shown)


internal/services/ingest.go
```diff
@@ -16,25 +17,75 @@ import (
...
+		return nil, errors.New("appTracker cannot be nil")
+	}
+	if httpClient == nil {
+		return nil, errors.New("httpClient cannot be nil")
+	}
+	if rpcURL == "" {
+		return nil, errors.New("rpcURL cannot be nil")
+	}
+
+	return &ingestService{
```
gouthamp-stellar, line 67: how are you planning on making rpc calls inside the ingestor? Won't you need the transaction service for it?

gouthamp-stellar: So then will the implementation of getLedgerEntries come in a following pr?

internal/tss/types.go
```diff
@@ -13,50 +51,57 @@ const (
...
 	TransactionXDR  string
 	// The status of an RPC sendTransaction call. Can be one of [PENDING, DUPLICATE, TRY_AGAIN_LATER, ERROR]
-	Status RPCTXStatus
+	Status entities.RPCStatus
 	// The (optional) error code that is derived by deserialzing the errorResultXdr string in the sendTransaction response
 	// list of possible errror codes: https://developers.stellar.org/docs/data/horizon/api-reference/errors/result-codes/transactions
 	Code RPCTXCode
 }
 
+func parseToRPCSendTxResponse(transactionXDR string, result entities.RPCSendTransactionResult, err error) (RPCSendTxResponse, error) {
```
gouthamp-stellar, line 73: Upper case

gouthamp-stellar:
This mostly looks good to me aside from the tiny comments, let me know when you have the tests running and I'll take another look

gouthamp-stellar (approved):

PR #47, opened 2024-09-30
services: rpc payment ingestor

gouthamp-stellar:
### What

This pr replaces the ingestion of payments with a getTransactions RPC call

### Why

We are removing all dependency on core in the wallet bakend, replacing it with rpc instead

### Known limitations

N/A

#### PR Structure

#### Thoroughness

#### Release


internal/tss/types.go
```diff
@@ -38,33 +41,70 @@ func ParseToRPCGetIngestTxResponse(result entities.RPCGetTransactionResult, err
...
+	EmptyCode           OtherCodes = 103
 )
 
 type RPCTXCode struct {
 	TxResultCode xdr.TransactionResultCode
 	OtherCodes   OtherCodes
 }
 
+var FinalErrorCodes = []xdr.TransactionResultCode{
+	xdr.TransactionResultCodeTxSuccess,
```
CaioTeixeira95, line 83: Why is this value on `FinalErrorCodes` slice?

CaioTeixeira95: Perhaps we can rename it to `FinalCodes` only or `FinalStateCodes`.

internal/utils/ingestion_utils.go
```diff
@@ -62,13 +57,13 @@ func Memo(memo xdr.Memo, txHash string) (*string, string) {
...
 
-func SourceAccount(op xdr.Operation, tx ingest.LedgerTransaction) string {
+func SourceAccount(op xdr.Operation, txEnvelope xdr.TransactionEnvelope) string {
 	account := op.SourceAccount
 	if account != nil {
 		return account.ToAccountId().Address()
 	}
-
-	return tx.Envelope.SourceAccount().ToAccountId().Address()
+	txEnvelope.SourceAccount()
```
CaioTeixeira95, line 65: Perhaps we can remove this line.

PR #77, opened 2024-11-15
Call RPC getHealth during ingestion

### LINKED ISSUE #62: Have the wallet backend's ingestor / api services call RPC's getHealth endpoint before calling RPC methods
Wallet-backend's ingest service depends on RPC to get the ledgers and transactions data. On first startup, we want to start the ingest service only when RPC becomes healthy.

A bonus functionality would be to do periodic health checks for RPC service within the ingest flow.

aditya1702:
Closes #62 

This PR introduces the following:
- Spin up a goroutine that keeps checking RPC health every 5 seconds.
- Use this new health check service within the wallet-backend ingestion process to make sure RPC is always healthy. The ingestion will wait if RPC becomes unhealthy at any point and will resume from the last ingested ledger.
- Add a way for ingestion to be in sync with RPC's ledger range:
     - If the wallet-backend is behind RPC's oldest ledger, then it will start ingestion from that ledger sequence.
     - If RPC is behind, then ingestion will wait for it to catch up with wallet-backend.
- Add unittests for the new features.


internal/entities/rpc.go
```diff
@@ -49,7 +56,7 @@ type Transaction struct {
 	ResultXDR           string    `json:"resultXdr"`
 	ResultMetaXDR       string    `json:"resultMetaXdr"`
 	Ledger              int64     `json:"ledger"`
-	DiagnosticEventsXDR string    `json:"diagnosticEventsXdr"`
+	DiagnosticEventsXDR []string  `json:"diagnosticEventsXdr"`
```
gouthamp-stellar, line 59: why is this an array now?

internal/services/ingest.go
```diff
@@ -42,6 +49,7 @@ func NewIngestService(
 	tssRouter tssrouter.Router,
 	tssStore tssstore.Store,
 ) (*ingestService, error) {
+	log.SetLevel(log.InfoLevel)
```
gouthamp-stellar, line 52: Was this line mean to be here?

internal/services/ingest.go
```diff
@@ -72,37 +80,64 @@ func NewIngestService(
...
+			// Case-1: wallet-backend is running behind rpc's oldest ledger. In this case, we start
+			// ingestion from rpc's oldest ledger.
+			case ingestLedger < resp.OldestLedger:
+				ingestLedger = resp.OldestLedger
+			// Case-2: rpc is running behind wallet-backend's latest synced ledger. We wait for rpc to
+			// catch back up to wallet-backend.
+			case ingestLedger > resp.LatestLedger:
+				log.Debugf("waiting for RPC to catchup to ledger %d (latest: %d)",
+					ingestLedger, resp.LatestLedger)
+				time.Sleep(5 * time.Second)
```
gouthamp-stellar, line 114: Is this sleep of 5 seconds necessary? What happens if you remove it?

internal/services/ingest.go
```diff
@@ -72,37 +80,64 @@ func NewIngestService(
...
+	for endLedger == 0 || ingestLedger <= endLedger {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("context cancelled: %w", ctx.Err())
+		case resp := <-rpcHeartbeat:
+			switch {
+			// Case-1: wallet-backend is running behind rpc's oldest ledger. In this case, we start
+			// ingestion from rpc's oldest ledger.
+			case ingestLedger < resp.OldestLedger:
+				ingestLedger = resp.OldestLedger
```
JakeUrban, line 108: Is this the desired functionality? Wouldn't a gap in the wallet backend's history be created if it was behind RPC's oldest ledger?

internal/services/ingest.go
```diff
@@ -72,37 +80,64 @@ func NewIngestService(
...
+			ledgerTransactions, err := m.GetLedgerTransactions(int64(ingestLedger))
+			if err != nil {
+				log.Error("getTransactions: %w", err)
+				continue
+			}
+			ingestHeartbeat <- true
+			err = m.ingestPayments(ctx, ledgerTransactions)
+			if err != nil {
+				return fmt.Errorf("error ingesting payments: %w", err)
+			}
```
JakeUrban, line 128: I'm assuming `ingestPayments` is a function that ingests all _inbound_ payments, since we have another function `processTSSTransactions` that I assume ingests all _outbound_ payments? Maybe we should update the function names to make that more clear?

Actually, I'm assuming both functions commit changes to the DB, which means that its possible to process _inbound_ payments for a ledger and fail to process the _outbound_ payments. If thats true, wouldn't this result in a bad state where we wouldn't be able to reingest the ledger, since retries would fail at the _inbound_ processing step?

internal/entities/rpc.go
aditya1702: Ummm, this has always been an array since the time I added the `GetTransactions` endpoint - https://github.com/stellar/stellar-rpc/pull/174/files#diff-5276db8001730f32b267cab637b316958b69aef6f8184343338855fdc118b700R66

I was actually confused when I got a parsing error in wallet-backend when it was just a `string`. And also the fact that it has been working fine before too.

internal/services/ingest.go
aditya1702: Actually I added this during one of the earlier iterations of this code. But its no longer needed. Good catch. Will remove it

internal/entities/rpc.go
gouthamp-stellar: Makes sense

internal/services/ingest.go
aditya1702: Resolving this as we discussed in 1-1

internal/services/ingest.go
aditya1702: @gouthamp-stellar I added the sleep here, so we give a little bit more time to rpc to catch up, instead of pinging it. Although, we can remove it since the goroutine invokes getHealth pings every 5s. What do you think?

aditya1702: I dont see a negative of giving it extra time to catch up though

internal/services/ingest.go
aditya1702: I would let @gouthamp-stellar answer here but yes you are correct that both these functions involve changes to db. So in order to avoid any of the issues you mentioned, we should use a single db transaction for both the functions.

@gouthamp-stellar any thoughts?

aditya1702:
> This mostly looks good, just one observation - once getHealth returns an actual healthy response, the expectation is that the ingestor _should_ never get a "ledger out of range" type error when calling getTransactions. We just need to be sure that this implementation supports that

@gouthamp-stellar Yes based on this implementation, whenever rpc and backend go out of sync, the health check will ensure that the ingest service only calls the db when both are back up to sync

gouthamp-stellar (approved):

internal/services/ingest.go
gouthamp-stellar: ingestPayments ingests inbound and outbound payments to (and from) a stellar account. I don't think we should update the names of these functions because they do exactly what the name suggests. processTSSTransactions for example looks for transactions that have been submitted by TSS and only updates those particular transactions in our tables. 

And yes, they both do commit to the DB, but to a completely different set of tables - so I don't think they should all be bundled up into one atomic transaction. The way that the ingestor works is that if even one of these functions fails, the ingestor ingests the same ledger again, which is fine, because both these functions are idempotent

JakeUrban: So if I submit a payment transaction to the TSS for one of the accounts registered with the wallet backend, there will be two records for the same payment right? One provided by the TSS' transaction history, and one provided by the `/payments` endpoint.

PR #83, opened 2024-11-26
Generic TSS

gouthamp-stellar:
### What

Making TSS generic - work for all transactions. Implementation of the doc: https://docs.google.com/document/d/1MJ3iyiTbASJp8hq1cTJO_aHDtP3Ko7kpMnKqzjYhITE/edit?tab=t.0

### Why

We need to make TSS work for all types of transactions and not just the disbursement use case

### Known limitations

N/A

https://github.com/orgs/stellar/projects/58/views/2?filterQuery=&sliceBy%5BcolumnId%5D=Status&sliceBy%5Bvalue%5D=In+progress&pane=issue&itemId=87419862&issue=stellar%7Cwallet-backend%7C79

#### PR Structure

#### Thoroughness

#### Release


aristidesstaffieri:
Should this work for transactions that need a specific signer(s)? 
If so maybe we could use an option to submit an already signed tx and only build it instead of build + sign it?

internal/tss/utils/operation_builder_test.go
```diff
@@ -0,0 +1,382 @@
...
+	"testing"
+
+	xdr3 "github.com/stellar/go-xdr/xdr3"
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/txnbuild"
+	"github.com/stellar/go/xdr"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBuildOperations(t *testing.T) {
```
aristidesstaffieri, line 15: should there be a case for `t.Run("op_invoke_host_function"...`?

aristidesstaffieri (approved):

internal/tss/services/mocks.go
```diff
@@ -20,6 +20,23 @@ func (t *TransactionServiceMock) NetworkPassphrase() string {
...
+
+func (t *TransactionServiceMock) BuildFeeBumpTransaction(ctx context.Context, tx *txnbuild.Transaction) (*txnbuild.FeeBumpTransaction, error) {
+	args := t.Called(ctx, tx)
+	if result := args.Get(0); result != nil {
+		return result.(*txnbuild.FeeBumpTransaction), args.Error(1)
+	}
+	return nil, args.Error(1)
+}
+
+/*
```
aditya1702, line 39: Remove commented functions

internal/tss/services/pool_populator.go
```diff
@@ -85,44 +83,8 @@ func (p *poolPopulator) routeNewTransactions(ctx context.Context) error {
 		if err != nil {
 			return fmt.Errorf("getting latest try for transaction: %w", err)
 		}
-		if try == (store.Try{}) {
-			// there is no try for this transactionm - route to RPC caller channel
+		if try == (store.Try{}) || try.Code == int32(tss.RPCFailCode) || try.Code == int32(tss.NewCode) {
```
aditya1702, line 86: why are we assigning a tss.NewStatus when try.code is RPCFailCode? In the else-if block below you are sending it to error-jitter channel for the same try.code status

internal/serve/httphandler/tss_handler.go
```diff
@@ -5,67 +5,110 @@ import (
...
+	Router             router.Router
+	Store              store.Store
+	AppTracker         apptracker.AppTracker
+	NetworkPassphrase  string
+	TransactionService tssservices.TransactionService
 }
 
 type Transaction struct {
 	Operations []string `json:"operations" validate:"required"`
+	TimeOut    int64    `json:"timeout" validate:"required"`
```
aditya1702, line 28: nit: can we rename this to Timebounds since that is what we use in our docs and other places

internal/tss/services/transaction_service_test.go
```diff
@@ -79,7 +79,204 @@ func TestBuildPayments(t *testing.T) {
...
+		}
+		tx, err := txService.BuildAndSignTransactionWithChannelAccount(context.Background(), []txnbuild.Operation{&payment}, 30)
+		assert.Equal(t, signedTx, tx)
+		assert.NoError(t, err)
+	})
+}
+
+func TestBuildFeeBumpTransaction(t *testing.T) {
+	distributionAccountSignatureClient := signing.SignatureClientMock{}
+	defer distributionAccountSignatureClient.AssertExpectations(t)
```
aditya1702, line 208: Hmm, shouldnt this be in each separate test case instead of a defer statement? Since we are setting different expectations in each case for this mock client, we would need to assert this in each test run

internal/tss/utils/operation_builder_test.go
aditya1702: +1

internal/serve/httphandler/tss_handler.go
```diff
@@ -5,67 +5,110 @@ import (
...
+	for _, txXDR := range reqParams.Transactions {
+		genericTx, err := txnbuild.TransactionFromXDR(txXDR)
+		if err != nil {
+			httperror.BadRequest("bad transaction xdr", nil).Render(w)
+			return
+		}
+		tx, txEmpty := genericTx.Transaction()
+		if !txEmpty {
+			httperror.BadRequest("bad transaction xdr", nil).Render(w)
+			return
```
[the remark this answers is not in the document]
gouthamp-stellar: I actually thought about this, but I think, at least for now, I'd rather take the all or nothing approach - ie: either all transactions are submitted to the network via TSS or none. If we do it the other way where we send the client back ones that are submitted and ones that are not, I think it just makes the API interface a little less clean. I think it's highly unlikely that this error will be triggered anyway.

internal/serve/httphandler/tss_handler.go
```diff
@@ -5,67 +5,110 @@ import (
...
+		}
+		txHash, err := tx.HashHex(t.NetworkPassphrase)
+		if err != nil {
+			httperror.InternalServerError(ctx, "unable to hashhex transaction", err, nil, t.AppTracker).Render(w)
+			return
```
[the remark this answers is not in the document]
gouthamp-stellar: See my comment above

internal/tss/services/mocks.go
gouthamp-stellar: done

internal/tss/services/pool_populator.go
gouthamp-stellar: Because these transactions have a status of NEW. Sure, I could route them to one of the other channels, but since they are currently in the NEW status, I want to build a payload such that the retries go to the rpc caller channel

internal/serve/httphandler/tss_handler.go
gouthamp-stellar: good point! made this change

internal/tss/services/transaction_service_test.go
gouthamp-stellar: I thought that if I called distributionAccountSignatureClient.AssertExpectations(t) with a defer, it would effectively call this function for ALL the individual test cases at the end of this test case. But I changed it so that I'm calling it in each individual function

internal/serve/httphandler/tss_handler.go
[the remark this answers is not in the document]
aditya1702: Actually yeah this step is building the transactionHashes so it makes sense that it would all fail together.

aditya1702 (approved):
LGTM

PR #85, opened 2024-12-05
openapi spec for the wallet backend

gouthamp-stellar:
### What

The openapi spec for the wallet backend

### Why

For api documentation

### Known limitations

N/A

https://github.com/orgs/stellar/projects/58/views/2?pane=issue&itemId=85462088&issue=stellar%7Cwallet-backend%7C69

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/openapi/main.yaml
+++ b/openapi/main.yaml
@@ -15,7 +15,7 @@ servers:
 tags:
   - name: Account Registration
     x-displayName: Account Registration
-    description: Clients of the Wallet Backed API can register and de-register stellar accounts whose payments they want to track
+    description: Clients of the Wallet Backed API can register and de-register stellar accounts whose payments they want to track, and whose transactions they want wrapped in fee bump transactions.
   - name: Payments
     description: Endpoint to retrieve paginated incoming and outgoing payments of a stellar account.
   - name: Transactions
@@ -28,7 +28,7 @@ paths:
       tags:
         - Account Registration
       summary: Register Stellar Account
-      description: 'Allows clients to register stellar accounts so that the wallet backend can index incoming and outgoing payments to and from these accounts.'
+      description: 'Allows clients to register stellar accounts so that the wallet backend can index incoming and outgoing payments to and from these accounts, and wrap transactions with these stellar accounts as the source in fee bump transactions'
       operationId: AccountRegistration
       parameters:
         - name: address
@@ -73,7 +73,7 @@ paths:
       tags:
         - Account Registration
       summary: De-register Stellar Account
-      description: 'Allows clients to de-register previously registered stellar accounts so that the wallet backend can stop indexing incoming and outgoing payments for the stellar account.'
+      description: 'Allows clients to de-register previously registered stellar accounts.'
       operationId: AccountDeRegistration
       parameters:
         - name: address
@@ -384,10 +384,10 @@ paths:
         - Transactions
       summary: Create a fee bump transaction where the distribution account pays for the fee.
       description:
-        Create a fee bump transaction where the distribution account pays for the transaction fee.
+        Create a fee bump transaction where the distribution account pays for the transaction fee. Note that the source account of the inner transaction should be registered with the wallet backend via the register API endpoint before calling this API endpoint
       operationId: createFeeBumpAccount
       requestBody:
-        description: "Request body for configuring address signers with weights and types."
+        description: "Request body containing the base64 transaction xdr string to be wrapped in a fee bump transaction."
         required: true
         content:
           application/json:
@@ -650,6 +650,15 @@ paths:
                   status:
                     type: string
                     description: "The current status of the transaction"
+                    enum:
+                      - NEW (not processed yet)
+                      - SENT (final tx result sent to client via webhook)
+                      - NOT_SENT (failed to send final tx to client via webhook)
+                      - FAILED (tx failed)
+                      - SUCCESS (tx success)
+                      - PENDING (tx pending)
+                      - ERROR (tx errored out when submitting, retries still in progress unless error is terminal)
+                      - TRY_AGAIN_LATER (an rpc response indicating a gongested network, retries in progress)
               example:
                 transactionhash: "Y6MF7SMT2a2d6pt3i37Xx9"
                 transactionxdr: "AAAAAgAAAADnY6MF7SMT2a2d6pt3i37Xx9IhaHqJ2lCcibNOISzhOwABX5AAFs2YAAAADAAAAAEAAAAAAAAAAAAAAABmYGswAAAAAAAAAAkAAAABAAAAAOdjowXtIxPZrZ3qm3eLftfH0iFoeonaUJyJs04hLOE7AAAAEAAAAACOC8v8STBDIULGM3FlZ6O7N3vHpNns7bcwRDFlIxTMiwAAAAEAAAAA52OjBe0jE9mtneqbd4t+18fSIWh6idpQnImzTiEs4TsAAAAAAAAAAI4L"
--- a/openapi/main.yaml
+++ b/openapi/main.yaml
@@ -477,7 +477,7 @@ paths:
                           description: base64 operation xdr string
                       timebounds:
                         type: integer
-                        description: time in seconds after which this transaction will not be accepted by the network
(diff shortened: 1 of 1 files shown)


openapi/main.yaml
```diff
@@ -0,0 +1,704 @@
...
+                          type: string
+                        destAmount:
+                          type: integer
+                        destAssetCode:
+                          type: string
+                        destAssetIssuer:
+                          type: string
+                        fromAddress:
+                          type: string
+                        memo:
```
CassioMG, line 196: should we add an optional `memoType` in addition to the `memo` value?

openapi/main.yaml
```diff
@@ -0,0 +1,704 @@
...
+                      next:
+                        type: string
+                      prev:
+                        type: string
+                  payments:
+                    type: array
+                    description: list of payments 
+                    items:
+                      type: object
+                      properties:
```
CassioMG, line 185: should it cover some [path payment properties](https://developers.stellar.org/docs/learn/fundamentals/transactions/list-of-operations#payment) like the below?
- `path` (e.g. ["XLM", "BTC", "USDC"]
- `destinationMin`
- `sendMax
- and maybe a `kind` prop (e.g. PAYMENT | PATH_PAYMENT_STRICT_SEND | PATH_PAYMENT_STRICT_RECEIVE)

Also, I am assuming we are not supporting Anchor Transaction properties on those payment objects. Is that right?
E.g.: `anchor transaction id`,  `kind`, `status`, `more_info_url`, etc...

JakeUrban:
@CassioMG lets hold off on discussions about changes we want to make to the API in this PR and focus on the spec itself for now. I do want to review the API's design holistically but lets do that separately.

openapi/main.yaml
gouthamp-stellar: All good points @CassioMG. And there's certainly other things that I think need to change in this API as well. I'm going to start a doc with all these recommendations

CassioMG:
> @CassioMG lets hold off on discussions about changes we want to make to the API in this PR and focus on the spec itself for now. I do want to review the API's design holistically as a team but lets keep that discussion separate.

Sounds good to me

openapi/main.yaml
```diff
@@ -0,0 +1,704 @@
...
+  termsOfService: https://stellar.org/terms-of-service
+  license:
+    name: Apache 2.0
+    url: http://www.apache.org/licenses/LICENSE-2.0.html
+servers:
+  - url: ''
+tags:
+  - name: Account Registration
+    x-displayName: Account Registration
+    description: Clients of the Wallet Backed API can register and de-register stellar accounts whose payments they want to track
```
JakeUrban, line 18: So to confirm, registration is only relevant to the data returned from `/payments`, right? These accounts are irrelevant to the `tx` and `tss` endpoints?

openapi/main.yaml
```diff
@@ -0,0 +1,704 @@
...
+                    properties:
+                      operations:
+                        type: array
+                        description: list of operations that make up the transaction
+                        items:
+                          type: string
+                          description: base64 operation xdr string
+                      timebounds:
+                        type: integer
+                        description: time in seconds after which this transaction will not be accepted by the network
```
JakeUrban, line 480: Is this the number of seconds _after the current time_ that the transaction expires, or the unix timestamp for when the transaction expires?

openapi/main.yaml
```diff
@@ -0,0 +1,704 @@
...
+              required:
+                - transactions
+            example:
+              transactions:
+                - opearations:
+                  - "Y6MF7SMT2a2d6pt3i37Xx9"
+                  - "M8GA7SMT2a2d6pt3i37Xx8"
+                  timebounds: 300
+      responses:
+        '200':
```
JakeUrban, line 493: So 200 is returned if all transactions provided in the request were built successfully, and 400 is returned if at least 1 transaction couldn't be built?

openapi/main.yaml
```diff
@@ -0,0 +1,704 @@
...
+                  type: string
+                  description: The url to which the final transaction results are to be sent
+                transactions:
+                  type: array
+                  items:
+                    type: string
+                    description: base64 encoded transaction xdr to submit to the stellar network
+                feebump:
+                  type: boolean
+                  description: boolean indicating if this transaction should be wrapped in a fee bump transaction
```
JakeUrban, line 565: Would it be equivalent for the client to submit transactions individually to `/tx/create-fee-bump`, and then submit all of them to this endpoint with the `feebump` parameter set to `false`?

openapi/main.yaml
```diff
@@ -0,0 +1,704 @@
...
+                - transactions
+            example:
+              webhookurl: "www.sdp.stellar.com/webhook"
+              transactions:
+                - "Y6MF7SMT2a2d6pt3i37Xx9"
+                - "M8GA7SMT2a2d6pt3i37Xx8"
+              feebump: true
+      responses:
+        '200':
+          description: "Successful response with the transaction hashes of the transactions to be submitted to the network. Clients can use these hashes to query TSS for the status of their transaction"
```
JakeUrban, line 577: Does this endpoint submit all transactions provided in the request to RPC before returning? Or are they just queued for submission?

openapi/main.yaml
```diff
@@ -0,0 +1,704 @@
...
+                properties:
+                  transactionhash:
+                    type: string
+                    description: "The hash of the transaction."
+                  transactionxdr:
+                    type: string
+                    description: "The base64-encoded xdr string of the transaction"
+                  status:
+                    type: string
+                    description: "The current status of the transaction"
```
JakeUrban, line 652: What are the values that can be returned here and their definitions?

openapi/main.yaml
```diff
@@ -0,0 +1,704 @@
...
+      parameters:
+        - name: address
+          in: query
+          description: The stellar address to register with the wallet backend
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: OK
```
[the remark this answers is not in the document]
gouthamp-stellar: The API does not have a response body for the registration endpoints as of now

openapi/main.yaml
```diff
@@ -0,0 +1,704 @@
...
+  /tx/create-fee-bump:
+    post:
+      tags:
+        - Transactions
+      summary: Create a fee bump transaction where the distribution account pays for the fee.
+      description:
+        Create a fee bump transaction where the distribution account pays for the transaction fee.
+      operationId: createFeeBumpAccount
+      requestBody:
+        description: "Request body for configuring address signers with weights and types."
```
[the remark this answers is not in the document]
gouthamp-stellar: Ha, yeap. Changed this to reflect the endpoint accurately

openapi/main.yaml
gouthamp-stellar: It actually also applies to the create-fee-bump-tx endpoint - ie: only transactions with stellar accounts that are registered with the wallet backend as the source can call the create-fee-bump-tx endpoint. We may need to revisit that logic though

openapi/main.yaml
gouthamp-stellar: Yeap, it is the number of seconds after the current time that the transaction expires. Changed this description to exactly that

openapi/main.yaml
gouthamp-stellar: That's right. It's all or nothing. Either all transactions are built successfully or none of them are, from the client's perspective

openapi/main.yaml
gouthamp-stellar: Yes thats right. Of course, that would add an additional endpoint for clients of TSS to call though before submitting them. I have to add though that in the /tx/create-fee-bump has some additional logic where it blocks certain operation types (which can be configured). Open to changing the api design around these endpoints though. I'm open to removing the feebump boolean for example

openapi/main.yaml
gouthamp-stellar: Nope. Transactions are queued for submission. These transaction hashes are just receipts that the client can use to call the GET /tss/transactions/{txhash} endpoint

JakeUrban (approved):
LGTM. I'm sure we'll refer back to this PR and the comments made when we're reviewing the API design itself.

PR #93, opened 2024-12-19
Replace Horizon calls with RPC

### LINKED ISSUE #75: [SPIKE] evaluate impact of removing all calls to Horizon within the wallet backend

aditya1702:
### What
- Replace horizon calls for fetching account details with RPC.
- Use TSS to submit the channel account creation operations instead of Horizon.
- Remove all horizon related flags, config variables and code pieces.
- Update tests to account for the above new changes.

### Why
Horizon is huge monolith that is very hard to maintain and this PR is part of org wide effort to move away from Horizon and use RPC instead which is very lightweight and easier to run than Horizon.

### Known limitations

N/A

Closes #75 

#### PR Structure

#### Thoroughness

#### Release


internal/services/channel_account_service.go
```diff
@@ -110,28 +123,55 @@ func (s *channelAccountService) submitCreateChannelAccountsOnChainTransaction(ct
...
+
+func (s *channelAccountService) submitToTSS(_ context.Context, payload tss.Payload) error {
+	err := s.Router.Route(payload)
+	if err != nil {
+		return fmt.Errorf("routing payload: %w", err)
+	}
+	time.Sleep(sleepDurationForChannelAccountCreation)
+
+	for range maxRetriesForChannelAccountCreation {
+		txResult, err := s.RPCService.GetTransaction(payload.TransactionHash)
```
JakeUrban, line 153: Is this approach used elsewhere in the codebase?

It seems odd to me that we are submitting a request to the TSS but getting the result from RPC. I would have figured that the TSS would provide a way to fetch transaction results, or even sent the result back to client who requested it be submitted.

Golang's channels seem like the right choice for something like this.

- client asks for a channel from the tss
- tss provides a golang channel
- client provides the transaction to the tss via the channel
- tss sends the transaction results back over the same channel

This would allow us to avoid polling RPC directly and instead just wait for a response on the channel provided to use from the TSS service.

internal/utils/utils.go
```diff
@@ -32,3 +37,34 @@ func UnwrapInterfaceToPointer[T any](i interface{}) *T {
...
 	return &value
 }
+
+func GetAccountLedgerKey(address string) (string, error) {
+	decoded, err := strkey.Decode(strkey.VersionByteAccountID, address)
+	if err != nil {
+		return "", err
+	}
+	var key xdr.Uint256
+	copy(key[:], decoded)
```
JakeUrban, line 47: Why are we copying `decoded` into `key`?

JakeUrban (commented):
This looks mostly good! Just a couple minor comments and one more meaningful question.

internal/services/account_sponsorship_service.go
```diff
@@ -56,11 +57,11 @@ var _ AccountSponsorshipService = (*accountSponsorshipService)(nil)
 
 func (s *accountSponsorshipService) SponsorAccountCreationTransaction(ctx context.Context, accountToSponsor string, signers []entities.Signer, supportedAssets []entities.Asset) (string, string, error) {
 	// Check the accountToSponsor does not exist on Stellar
-	_, err := s.HorizonClient.AccountDetail(horizonclient.AccountRequest{AccountID: accountToSponsor})
+	_, err := s.RPCService.GetAccountLedgerSequence(accountToSponsor)
 	if err == nil {
 		return "", "", ErrAccountAlreadyExists
 	}
-	if !horizonclient.IsNotFoundError(err) {
+	if err.Error() != accountNotFoundError {
```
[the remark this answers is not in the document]
aditya1702: Ah good catch. Yeah the idiomatic way is to use `errors.Is` instead. I used this way to make sure the code was working but did not realise the negatives of using it. Thanks for spotting

internal/utils/utils.go
aditya1702: Because `decoded` is a byte array and `key` is a fixed size [32]byte array. In order to convert slice to fixed size, the recommended way would be to use `copy()`: https://stackoverflow.com/questions/30285680/how-to-convert-slice-to-fixed-size-array.

We can also use something like this - `key = *(*[32]byte)(decoded)` but `copy()` would be cleaner

internal/serve/serve.go
```diff
@@ -150,9 +149,10 @@ func initHandlerDeps(cfg Configs) (handlerDeps, error) {
 		return handlerDeps{}, fmt.Errorf("instantiating stellar signature verifier: %w", err)
 	}
 
-	horizonClient := horizonclient.Client{
-		HorizonURL: cfg.HorizonClientURL,
-		HTTP:       &http.Client{Timeout: 40 * time.Second},
+	httpClient := http.Client{Timeout: time.Duration(30 * time.Second)}
```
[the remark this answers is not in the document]
aditya1702: Hmm, this line was not added by me. I just rearranged the code lines in this diff. Maybe @gouthamp-stellar could answer this? From the code, it looks like we define a client for new services like the [rpcService](https://github.com/stellar/wallet-backend/blob/main/internal/serve/serve.go#L206) and [webHook channel](https://github.com/stellar/wallet-backend/blob/main/internal/serve/serve.go#L245)

internal/services/channel_account_service.go
aditya1702: @JakeUrban So this approach is actually what I started with and the design also makes sense. The TSS has been configured in a way where we provide a webhook URL to the router in the payload [here](https://github.com/stellar/wallet-backend/blob/main/internal/tss/types.go#L195). This fits into the design of TSS where clients would then listen to this webhook to get updates on the txns they have submitted . The solution would be to write a webhook handler and use a go channel to keep checking for the transaction status update on the URL. (correct me if I am wrong here @gouthamp-stellar )

In this case, however, the channel account txn is only submitted once during the initial starting of wallet-backend server. So implementing a new webhook handler and channels to just poll for this one txn seemed too much and I instead just used the `GetTransaction` to look for the txn.

But we can definitely discuss and change it accordingly

internal/services/rpc_service.go
```diff
@@ -116,6 +134,25 @@ func (r *rpcService) SendTransaction(transactionXDR string) (entities.RPCSendTra
 	return result, nil
 }
 
+func (r *rpcService) GetAccountLedgerSequence(address string) (int64, error) {
```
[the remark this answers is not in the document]
aditya1702: Because currently the only use case for getting an account is to get its latest sequence number. I decided to just add function for that and we can always change the definition in the future to return account object with the other details. What do you think?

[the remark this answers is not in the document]
JakeUrban: Fine by me

internal/services/channel_account_service.go
JakeUrban: Maybe we should just submit to RPC rather than use half of the TSS solution (submission but not results)

aditya1702: @JakeUrban That is not as straightforward as it sounds because in that case, we have to handle all the retries and resubmissions based on the status that RPC sends. The TSS does this automatically through its channels implementation. That is why using TSS for submission is the cleaner approach here.

@gouthamp-stellar do you have any thoughts on this issue?

internal/entities/rpc.go
```diff
@@ -77,6 +77,19 @@ type RPCSendTransactionResult struct {
 	ErrorResultXDR        string    `json:"errorResultXdr"`
 }
 
+type LedgerEntryResult struct {
+	KeyXDR             string `json:"key,omitempty"`
+	DataXDR            string `json:"xdr,omitempty"`
```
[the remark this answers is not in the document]
aditya1702: Yes that is true but this is defined directly in the RPC codebase [here](https://github.com/stellar/stellar-rpc/blob/main/cmd/stellar-rpc/internal/methods/get_ledger_entries.go#L27) so I kept it as is

internal/services/rpc_service.go
```diff
@@ -116,6 +134,25 @@ func (r *rpcService) SendTransaction(transactionXDR string) (entities.RPCSendTra
...
+	if len(result.Entries) == 0 {
+		return 0, fmt.Errorf("entry not found for account public key")
+	}
+	accountEntry, err := utils.GetAccountFromLedgerEntry(result.Entries[0])
+	if err != nil {
+		return 0, fmt.Errorf("decoding account entry for account public key: %w", err)
+	}
+	return int64(accountEntry.SeqNum), nil
+}
+
```
gouthamp-stellar, line 155: nit: I do like the idea of this function, but I wonder if it should also be in utils.go, because I feel like it is essentially just parsing an rpc response

internal/services/channel_account_service.go
gouthamp-stellar: I think we should either use TSS or RPC for this, and not both. 

If we went the TSS route, instead of polling RPC's GetTransaction, we could poll the tss/transactions/{transactionhash} endpoint till we get a 'SUCCESS' response. The tss/transactions/{transactionhash} endpoint looks up the transaction in the tss transactions table, which is updated by TSS/the ingestor. Or you could repeatedly poll the tss transactions table yourself 

If we go with using RPC directly, we may be able to get away with something simple here like using rpc_service.go's SendTransaction and call GetTransaction in a loop. Basically, instead of submitting the transaction to the router like you currently are, submit it to RPC directly

I think the RPC approach is cleaner. I get that it's possible that we'll have cases where we'll have to retry and resubmit, but it's just a single transaction that we send upon startup. I think using TSS, which is really meant for high volume transaction submission, is a little overkill

gouthamp-stellar:
One thing to note here is that if we are using RPC, we need to be sure that it is ready (ie gethealth returns a positive result), which can take upto a few minutes if I'm not mistaken. So you may not be able to create all those channel accounts as soon as wallet backend starts up - you'd have to wait till RPC is ready to create those channel accounts right?

gouthamp-stellar (changes requested):
take a look at my comments

internal/services/channel_account_service.go
```diff
@@ -82,14 +89,22 @@ func (s *channelAccountService) EnsureChannelAccounts(ctx context.Context, numbe
 }
 
 func (s *channelAccountService) submitCreateChannelAccountsOnChainTransaction(ctx context.Context, distributionAccountPublicKey string, ops []txnbuild.Operation) error {
-	distributionAccount, err := s.HorizonClient.AccountDetail(horizonclient.AccountRequest{AccountID: distributionAccountPublicKey})
+	err := waitForRPCServiceHealth(ctx, s.RPCService)
```
gouthamp-stellar, line 92: will this wait indefinitely till RPC is healthy? If so, is it possible that it will wait for a long time since RPC can take a while to get healthy?

internal/services/channel_account_service.go
```diff
@@ -110,27 +125,89 @@ func (s *channelAccountService) submitCreateChannelAccountsOnChainTransaction(ct
...
+			time.Sleep(sleepDelayForChannelAccountCreation)
+			continue
+		}
+
+	}
+
+	return fmt.Errorf("transaction did not complete after %d attempts", maxRetriesForChannelAccountCreation)
+}
+
+func (s *channelAccountService) getTransactionStatus(_ context.Context, hash string) error {
```
gouthamp-stellar, line 186: maybe this should be named something like ensureTransactionSuccess, since it doesn't actually get a transaction's status but rather repeatedly pings RPC's GetTransaction till it returns a success. Wdty?

aditya1702: Hmm, yeah thats a good suggestion.

internal/services/channel_account_service.go
aditya1702: @gouthamp-stellar It would wait till the ctx is cancelled [here](https://github.com/stellar/wallet-backend/pull/93/files#diff-ee84d6b11d3ccdf80838be43c58eedc317f6c196f2ce5ebd627e06ec42c00a97R157). However, I am realising now that `ctx` is a `context.Background()` which does not have a default timeout. So we will have to add our own timeout here to make sure it exits

JakeUrban: To clarify, if we reach the timeout, we would exit the process and log some error about the RPC being unavailable?

JakeUrban: Nvm, the code speaks for itself

JakeUrban (approved):

internal/services/channel_account_service.go
gouthamp-stellar: If rpc does not get healthy within these 5 minutes, then it seems like the channel accounts won't get created, right? If that's the case, it may be worth it to have ensureChannelAccounts (https://github.com/stellar/wallet-backend/blob/main/internal/serve/serve.go#L302) be called on a ticker or something, throughout the lifespan of the wallet backend. Unless we can be sure that RPC will get healthy within the 5 minutes.

gouthamp-stellar (approved):

PR #104, opened 2025-01-10
Centralize RPC health monitoring

### LINKED ISSUE #105: Centralise RPC health monitoring
The RPC health monitoring is used by the ingest and api servers. While it works as intended, its design is not that great as it has some code duplication. We want to centralise the code to rpc service so its cleaner and easier to understand

aditya1702:
### What

- Currently, the code for tracking RPC health is duplicated across both the ingest server and the api server (when creating the channel accounts). This PR organises the RPC heartbeat tracking in a central location and simplifies its design.
- Updated tests accordingly.

### Why

The current solution for tracking rpc health was not a simple one and had some code duplication which are resolved in this PR.

### Known limitations

NA

Closes #105 

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/services/channel_account_service_test.go
+++ b/internal/services/channel_account_service_test.go
@@ -3,6 +3,7 @@ package services
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/stellar/go/keypair"
 	"github.com/stellar/go/network"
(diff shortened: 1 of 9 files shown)


internal/services/channel_account_service_test.go
```diff
@@ -253,36 +267,6 @@ func TestChannelAccountServiceEnsureChannelAccounts(t *testing.T) {
 	})
 }
 
-func TestWaitForRPCServiceHealth(t *testing.T) {
```
gouthamp-stellar, line 256: shouldn't this test be replaced by another one where we are making sure that the creation of channel accounts only happens when rpc is healthy?

aditya1702: @gouthamp-stellar I actually moved these tests to rpc_service_test.go

aditya1702: @gouthamp-stellar Yeah, I dont know why this test got removed. I will add it back again

gouthamp-stellar (approved):

PR #116, opened 2025-01-26
Locking channel accounts to transactions

gouthamp-stellar:
Instead of locking channel accounts for an arbitrary amount of time like we are currently doing, lock them for the same amount of time as the transaction time bounds and also explicitly lock them to transactions, unlocking them in the webhook channel

### What

Instead of locking channel accounts for an arbitrary amount of time like we are currently doing, lock them for the same amount of time as the transaction time bounds and also explicitly lock them to transactions, unlocking them in the webhook channel

### Why

If the channel account is specified as the source of a transaction that hasn't been accepted by the network yet, and the same channel account is "unlocked" and then used as the source for a different transaction that gets accepted by the network prior to the original, the original transaction is invalidated. We want to avoid this behavior

### Known limitations

N/A

https://github.com/orgs/stellar/projects/58/views/2?pane=issue&itemId=92871835&issue=stellar%7Cwallet-backend%7C102
#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/signing/channel_account_db_signature_client.go
+++ b/internal/signing/channel_account_db_signature_client.go
@@ -43,9 +43,16 @@ func (sc *channelAccountDBSignatureClient) NetworkPassphrase() string {
 	return sc.networkPassphrase
 }
 
-func (sc *channelAccountDBSignatureClient) GetAccountPublicKey(ctx context.Context) (string, error) {
+func (sc *channelAccountDBSignatureClient) GetAccountPublicKey(ctx context.Context, opts ...int) (string, error) {
+	var lockedUntil time.Duration
+	if len(opts) > 0 {
+		lockedUntil = time.Duration(opts[0]) * time.Second
+	} else {
+		lockedUntil = time.Minute
+	}
 	for range store.ChannelAccountWaitTime {
-		channelAccount, err := sc.channelAccountStore.GetIdleChannelAccount(ctx, time.Minute)
+		// check to see if the variadic parameter for time exists and if so, use it here
+		channelAccount, err := sc.channelAccountStore.GetIdleChannelAccount(ctx, lockedUntil)
 		if err != nil {
 			if errors.Is(err, store.ErrNoIdleChannelAccountAvailable) {
 				log.Ctx(ctx).Warn("All channel accounts are in use. Retry in 1 second.")
--- a/internal/signing/env_signature_client.go
+++ b/internal/signing/env_signature_client.go
@@ -36,7 +36,7 @@ func (sc *envSignatureClient) NetworkPassphrase() string {
 	return sc.networkPassphrase
 }
 
-func (sc *envSignatureClient) GetAccountPublicKey(ctx context.Context) (string, error) {
+func (sc *envSignatureClient) GetAccountPublicKey(ctx context.Context, opts ...int) (string, error) {
 	return sc.distributionAccountFull.Address(), nil
 }
 
--- a/internal/signing/kms_signature_client.go
+++ b/internal/signing/kms_signature_client.go
@@ -59,7 +59,7 @@ func NewKMSSignatureClient(publicKey string, networkPassphrase string, keypairSt
 	}, nil
 }
 
-func (sc *kmsSignatureClient) GetAccountPublicKey(ctx context.Context) (string, error) {
+func (sc *kmsSignatureClient) GetAccountPublicKey(ctx context.Context, opts ...int) (string, error) {
 	return sc.distributionAccountPublicKey, nil
 }
 
--- a/internal/signing/mocks.go
+++ b/internal/signing/mocks.go
@@ -18,7 +18,7 @@ func (s *SignatureClientMock) NetworkPassphrase() string {
 	return args.String(0)
 }
 
-func (s *SignatureClientMock) GetAccountPublicKey(ctx context.Context) (string, error) {
+func (s *SignatureClientMock) GetAccountPublicKey(ctx context.Context, opts ...int) (string, error) {
 	args := s.Called(ctx)
 	return args.String(0), args.Error(1)
 }
--- a/internal/signing/store/channel_accounts_model.go
+++ b/internal/signing/store/channel_accounts_model.go
@@ -29,6 +29,7 @@ func (ca *ChannelAccountModel) GetIdleChannelAccount(ctx context.Context, locked
 	query := fmt.Sprintf(`
 		UPDATE channel_accounts
 		SET 
+			locked_tx_hash = NULL,
 			locked_at = NOW(),
 			locked_until = NOW() + INTERVAL '%d seconds'
 		WHERE public_key = (
@@ -36,13 +37,12 @@ func (ca *ChannelAccountModel) GetIdleChannelAccount(ctx context.Context, locked
 				public_key
 			FROM channel_accounts
 			WHERE 
-				locked_until IS NULL
-				OR locked_until < NOW()
+				(locked_tx_hash IS NULL AND (locked_until IS NULL OR locked_until < NOW()))
 			ORDER BY random()
 			LIMIT 1
 			FOR UPDATE SKIP LOCKED
 		)
-		RETURNING *
+		RETURNING *;
 	`, int64(lockedUntil.Seconds()))
 
 	var channelAccount ChannelAccount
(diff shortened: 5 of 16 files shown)


internal/signing/store/channel_accounts_model.go
```diff
@@ -29,20 +29,20 @@ func (ca *ChannelAccountModel) GetIdleChannelAccount(ctx context.Context, locked
 	query := fmt.Sprintf(`
 		UPDATE channel_accounts
 		SET 
+			locked_tx_hash = NULL,
```
JakeUrban, line 32: I think this is unnecessary since `UnlockChannelAccountFromTx()` sets this to null

internal/signing/store/channel_accounts_model.go
```diff
@@ -84,6 +84,24 @@ func (ca *ChannelAccountModel) GetAllByPublicKey(ctx context.Context, sqlExec db
...
+
+func (ca *ChannelAccountModel) UnlockChannelAccountFromTx(ctx context.Context, txHash string) error {
+	const query = `UPDATE channel_accounts SET locked_tx_hash = NULL, locked_at = NULL, locked_until = NULL WHERE locked_tx_hash = $1`
+	_, err := ca.DB.ExecContext(ctx, query, txHash)
+	if err != nil {
+		return fmt.Errorf("unlocking channel account: %w", err)
+	}
+	return nil
+}
+
```
JakeUrban, line 104: These aren't reciprocal functions, but their names imply they are. `LockChannelAccountToTx()` is adding the tx hash of the transaction to the already-locked account, whereas `UnlockChannelAccountFromTx()` is unlocking the account. Maybe we can adjust the names to reflect the differences? 

I was also surprised to see `GetIdleChannelAccount()` lock the account since it's name doesn't imply its updating the account's state.

internal/signing/channel_account_db_signature_client.go
```diff
@@ -43,9 +43,16 @@ func (sc *channelAccountDBSignatureClient) NetworkPassphrase() string {
...
+	var lockedUntil time.Duration
+	if len(opts) > 0 {
+		lockedUntil = time.Duration(opts[0]) * time.Second
+	} else {
+		lockedUntil = time.Minute
+	}
 	for range store.ChannelAccountWaitTime {
-		channelAccount, err := sc.channelAccountStore.GetIdleChannelAccount(ctx, time.Minute)
+		// check to see if the variadic parameter for time exists and if so, use it here
+		channelAccount, err := sc.channelAccountStore.GetIdleChannelAccount(ctx, lockedUntil)
```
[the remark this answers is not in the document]
gouthamp-stellar: I wanted to pass in the time for which to lock a channel account as an optional argument, so that not every signature client has to pass it in, which is why I made it a variadic function

internal/signing/store/channel_accounts_model.go
gouthamp-stellar: That is true in theory, but in case the channel accounts do not get unlocked in the webhook channel for some reason, this will act as a fail safe, which is why I added it in there

internal/signing/store/channel_accounts_model.go
gouthamp-stellar: That's a good point. I renamed these functions like so to more accurately depict what they do:

GetIdleChannelAccount -> GetAndLockIdleChannelAccount
LockChannelAccountToTx -> AssignTxToChannelAccount
UnlockChannelAccountFromTx -> UnassignTxAndUnlockChannelAccount

JakeUrban:
resolves #102

internal/signing/env_signature_client.go
```diff
@@ -36,7 +36,7 @@ func (sc *envSignatureClient) NetworkPassphrase() string {
 	return sc.networkPassphrase
 }
 
-func (sc *envSignatureClient) GetAccountPublicKey(ctx context.Context) (string, error) {
+func (sc *envSignatureClient) GetAccountPublicKey(ctx context.Context, opts ...int) (string, error) {
```
aditya1702, line 39:
```suggestion
func (sc *envSignatureClient) GetAccountPublicKey(ctx context.Context, _ ...int) (string, error) {
```

internal/signing/kms_signature_client.go
```diff
@@ -59,7 +59,7 @@ func NewKMSSignatureClient(publicKey string, networkPassphrase string, keypairSt
 	}, nil
 }
 
-func (sc *kmsSignatureClient) GetAccountPublicKey(ctx context.Context) (string, error) {
+func (sc *kmsSignatureClient) GetAccountPublicKey(ctx context.Context, opts ...int) (string, error) {
```
aditya1702, line 62:
```suggestion
func (sc *kmsSignatureClient) GetAccountPublicKey(ctx context.Context, _ ...int) (string, error) {
```

internal/signing/mocks.go
```diff
@@ -18,7 +18,7 @@ func (s *SignatureClientMock) NetworkPassphrase() string {
 	return args.String(0)
 }
 
-func (s *SignatureClientMock) GetAccountPublicKey(ctx context.Context) (string, error) {
+func (s *SignatureClientMock) GetAccountPublicKey(ctx context.Context, opts ...int) (string, error) {
```
aditya1702, line 21:
```suggestion
func (s *SignatureClientMock) GetAccountPublicKey(ctx context.Context, _ ...int) (string, error) {
```

internal/tss/services/transaction_service.go
```diff
@@ -69,7 +77,7 @@ func (t *transactionService) NetworkPassphrase() string {
 }
 
 func (t *transactionService) BuildAndSignTransactionWithChannelAccount(ctx context.Context, operations []txnbuild.Operation, timeoutInSecs int64) (*txnbuild.Transaction, error) {
-	channelAccountPublicKey, err := t.ChannelAccountSignatureClient.GetAccountPublicKey(ctx)
+	channelAccountPublicKey, err := t.ChannelAccountSignatureClient.GetAccountPublicKey(ctx, int(timeoutInSecs))
```
aditya1702, line 80: Correct me if I am wrong but I think there is a bug here: what happens if we get an error in the code until we assign the locked channel account to the txn? In this case the function exits and the db change to lock channel account is not rolled back, since`ExecContext` executes a single SQL statement - any error will leave the database in its current state. The channel account will be unavailable until the timeout expires. 

A better design would be to start a db transaction and do calls to `GetAndLockIdleChannelAccount ` and `AssignTxToChannelAccount ` in the same txn and rollback the txn if the function exits due to an error.

internal/signing/mocks.go
gouthamp-stellar: Not sure why we should be replacing opts with _ here. This is a mock for the signature client, it is mocking the interface

internal/tss/services/transaction_service.go
gouthamp-stellar: Good point! Made the change to have this in an atomic transaction

internal/signing/mocks.go
aditya1702: @gouthamp-stellar This is only for linting purposes. Since the opts variable is not being used in the function, it would throw a lint error.

internal/signing/channel_account_db_signature_client.go
```diff
@@ -43,9 +43,16 @@ func (sc *channelAccountDBSignatureClient) NetworkPassphrase() string {
...
+	var lockedUntil time.Duration
+	if len(opts) > 0 {
+		lockedUntil = time.Duration(opts[0]) * time.Second
+	} else {
+		lockedUntil = time.Minute
+	}
 	for range store.ChannelAccountWaitTime {
-		channelAccount, err := sc.channelAccountStore.GetIdleChannelAccount(ctx, time.Minute)
+		// check to see if the variadic parameter for time exists and if so, use it here
+		channelAccount, err := sc.channelAccountStore.GetAndLockIdleChannelAccount(ctx, lockedUntil)
```
aditya1702, line 55: @gouthamp-stellar Shouldnt we also run this in a txn? So it can be rolled back?

gouthamp-stellar: This is essentially already being wrapped in a transaction in transaction_service.go so doing it again here would make it a nested transaction, which I don't think is even possible in postgres

internal/tss/channels/webhook_channel.go
```diff
@@ -64,6 +73,10 @@ func (p *webhookPool) Receive(payload tss.Payload) {
 	var i int
 	sent := false
 	ctx := context.Background()
+	err = p.UnlockChannelAccount(ctx, payload.TransactionXDR)
```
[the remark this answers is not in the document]
gouthamp-stellar: This is just a single sql statement so I dont think it needs to be. Also, channel accounts are assigned to one and only one transaction at a time, so we wont have 2 transactions trying to modify this same row in the table

aditya1702 (approved):

internal/signing/mocks.go
gouthamp-stellar: got it

PR #142, opened 2025-04-16
all: Adopt recommended linters in CI and address the problems pointed by them

marcelosalloum:
### What

Adopt recommended linters in CI and address the problems pointed by them

Fixes include:
- `GET /tss/transactions/{hash}` would return an invalid JSON when the resource was not found
- Many errors where not being wrapped with `fmt.Errorf("{message}: %w")`, which made issues hard to debug.
- Some occurrences of the wrong error being returned in a funcrtion/method
- Dozens of occurrences of shadowing variables, some of them in the actual code but most of them in the tests.
- Many tests were asserting the wrong object for an error.
- Sentry tests were failing but ut wasn't being captured since the errors were being ignored.
- Error assertions were fixed to use `errors.As` and `errors.Is`, so wrapped errors can also be captured.
- Update naming from some error types (should end with `...Error`) and error variables (should start with `Err...`) to be go idiomatic.

Additionally:
- Moved all mocks to a file in their package called `mocks.go`, for consistency.
- Reorganize imports using `goimports`

### Why

It's good practice to use linters, specially the ones that capture errors and help making the code more debuggable and idiomatic.

#### PR Structure

#### Thoroughness

#### Release


internal/tss/services/transaction_service.go
```diff
@@ -84,48 +84,41 @@ func (t *transactionService) NetworkPassphrase() string {
 }
 
 func (t *transactionService) BuildAndSignTransactionWithChannelAccount(ctx context.Context, operations []txnbuild.Operation, timeoutInSecs int64) (*txnbuild.Transaction, error) {
-	var tx *txnbuild.Transaction
-	var channelAccountPublicKey string
-	err := db.RunInTransaction(ctx, t.DB, nil, func(dbTx db.Transaction) error {
```
marcelosalloum, line 89: I removed the call to `RunInTransaction` here because it was not being used. The dbTx was not being applied anywhere and considering there's only one database query, at the very last step here, then the database transaction serves no purpose here.

internal/services/kms_import_key_service.go
```diff
@@ -59,9 +60,6 @@ func (s *kmsImportService) ImportDistributionAccountKey(ctx context.Context, dis
 
 	err = s.keypairStore.Insert(ctx, kp.Address(), output.CiphertextBlob)
 	if err != nil {
-		if errors.Is(err, store.ErrPublicKeyAlreadyExists) {
-			return err
-		}
 		return fmt.Errorf("storing distribution account encrypted private key: %w", err)
```
marcelosalloum, line 63: I removed this assertion because the function returns the same error anyway, and it can be parsed downstream using `errors.Is`

aristidesstaffieri (approved):

Issue #139, opened 2025-04-03
Bugfix: make the API server executable by fixing the dependency injection issues


marcelosalloum:
### What

Make the API server executable by fixing the dependency injection issues.

### Why

Due to missing dependencies, the API server crashes when we try to run it. Also, some API endpoints may not crash the application at launch but they will trigger 500s due to missing dependency.

PR #154, opened 2025-04-18
{cmd, auth, utils}: Refactor & modularize parts of the code in preparation for the e2e tests

marcelosalloum:
### What

Refactor and modularize export parts of the code in preparation for the e2e tests.

What was exported into reusable functions:
- http signature creator
- conversion helpers to make the following path easier: `txnbuild.Operation <> XDR <> base64 <> XDR <> txnbuild.Operation`

Also:
- Fixed the 500 crash that would prevent the feeBumpTx endpoint from working due to a missing dependency.
- Renamed the variable `wallet-signing-key` to `client-auth-public-keys`, which now is more descriptive and allows multiple signers to be set, making key rotation easier.

### Why

While writing the e2e tests, stumbled upon parts of the code that had only one way methods available. For instance, the code knows how to check a signature but doesn't know how to produce it. That led me to create such methods and export them in the `pkg` folder, so they can be used externally.

### Breaking Change

Renamed `WALLET_SIGNING_KEY` to `CLIENT_AUTH_PUBLIC_KEYS` and allow it to take multiple signers.

Address #139.

#### PR Structure

#### Thoroughness

#### Release


internal/serve/auth/signature_verifier.go
```diff
@@ -54,38 +49,30 @@ var _ SignatureVerifier = (*StellarSignatureVerifier)(nil)
...
-		return ErrStellarSignatureNotVerified
+	for _, kp := range sv.ClientAuthPublicKeys {
+		err = kp.Verify([]byte(payload), signatureBytes)
+		if err == nil {
+			return nil
+		}
 	}
 
-	return nil
+	return fmt.Errorf("unable to verify the signature for the given payload: %w", err)
```
aditya1702, line 75: I actually agree with Copilot's review here - should we move the error return statement inside the FOR loop and we can also log the public key which generates the error

internal/serve/auth/signature_verifier.go
```diff
@@ -40,12 +35,12 @@ type ExpiredSignatureTimestampError struct {
 }
 
 func (e ExpiredSignatureTimestampError) Error() string {
-	return fmt.Sprintf("signature timestamp has expired. sig timestamp: %s, check time %s", e.ExpiredSignatureTimestamp.Format(time.RFC3339), e.CheckTime.Format(time.RFC3339))
+	return fmt.Sprintf("timestamp expired by %s", e.CheckTime.Sub(e.ExpiredSignatureTimestamp).String())
```
aditya1702, line 38: it would be helpful to add the metric after the expiry number. e.g. "timestamp expired by 5s"

internal/serve/auth/signature_verifier.go
marcelosalloum: The only possible error there is `ErrInvalidSignature`:
https://github.com/stellar/go/blob/3ad642733b72f7f7df8262ebc9aa768bf93b7767/keypair/from_address.go#L56-L64

> should we move the error return statement inside the FOR loop and we can also log the public key which generates the error

We can't return from inside the loop because the for loop is testing if _any_ of the available public keys has signed that payload. It means that if we have multiple public keys allowed to sign, only one will succeed and the rest will return an error. Meaning it is only an error if all the loops fail.

I'll aggregate the errors and return them in the end for reference.

marcelosalloum: Addressed in https://github.com/stellar/wallet-backend/pull/154/commits/15b0e7dc3dd5da5a77983b430ab66dfceaa806ee.

internal/serve/auth/signature_verifier.go
marcelosalloum: Good idea, added in https://github.com/stellar/wallet-backend/pull/154/commits/2c5a0554dbb024c4ac5e98ce1b4a7382a3971308.

internal/serve/auth/signature_verifier.go
aditya1702: Ah gotcha makes sense.

aditya1702 (approved):
LGTM

Issue #149, opened 2025-04-16
Feature Request: implement e2e integration tests


marcelosalloum:
### What problem does your feature solve?

Ensure that the happy path is working e2e for different operation types.

### What would you like to see?

A go command that runs e2e integration tests and ensures the happy path works for at least 2 types of operations:
- payments
- contract invocation

Ideally, al types of operations should be covered, but these two are the minimal flows we should ensure.

⚠  The go command NEEDS TO ultimately be added to an e2e GH workflow, to ensure that no PR is breaking the happy path.

### What alternatives are there?

We could implement it without using Go, but including additional languages would only complicate things.

PR #160, opened 2025-04-23
integrationtests: submit tx to RPC and poll from RPC

marcelosalloum:
### What

Update integration tests to submit the transaction to the RPC and poll the RPC waiting for a successful response.

### Why

As part of #149

### Known limitations

The steps to be added are:
- [x] buildTx (`POST /tss/transactions/build`)
- [x] feeBumpTx (`POST /tx/create-fee-bump`)
- [x] submitTx
- [x] Poll the tx status from the Stellar network
- [ ] Poll the tx status from the wallet-backend (`GET /transactions/{transactionhash}`)
- [ ] Expand the number of operations being tested
- [ ] (future) webhooks

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/integrationtests/integrationtests.go
+++ b/internal/integrationtests/integrationtests.go
@@ -117,6 +117,11 @@ func (it *IntegrationTests) Run(ctx context.Context) error {
 		if innerErr != nil {
 			return fmt.Errorf("parsing transaction from XDR: %w", innerErr)
 		}
+		innerTxHash, innerErr := tx.HashHex(it.NetworkPassphrase)
+		if innerErr != nil {
+			return fmt.Errorf("hashing transaction: %w", innerErr)
+		}
+		log.Ctx(ctx).Infof("=====> innerHash: %s", innerTxHash)
 		signedTx, innerErr := tx.Sign(it.NetworkPassphrase, it.SourceAccountKP)
 		if innerErr != nil {
 			return fmt.Errorf("signing transaction: %w", innerErr)
--- a/internal/integrationtests/integrationtests.go
+++ b/internal/integrationtests/integrationtests.go
@@ -2,6 +2,7 @@ package integrationtests
 
 import (
 	"context"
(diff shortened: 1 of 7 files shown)


internal/integrationtests/integrationtests.go
```diff
@@ -74,74 +80,157 @@ type IntegrationTests struct {
...
+	}
+
 	// TODO: verifyTxResult in wallet-backend")
 
 	return nil
 }
 
+func (it *IntegrationTests) waitForTransactionConfirmation(ctx context.Context, hash string, retryOptions ...retry.Option) error {
+	attemptsCount := 0
+	outerErr := retry.Do(
```
aditya1702, line 200: Ohh wow, I did not know about this retry-go library. This is pretty neat.

internal/integrationtests/integrationtests.go
```diff
@@ -74,74 +80,157 @@ type IntegrationTests struct {
...
+		},
+		append(
+			retryOptions,
+			retry.Context(ctx),
+			retry.LastErrorOnly(true),
+		)...,
+	)
+
+	if outerErr != nil {
+		return fmt.Errorf("failed to get transaction status after %d attempts: %w", attemptsCount, outerErr)
```
aditya1702, line 229: I see we have hardcoded the attempts here - are we sure to get the transaction confirmation in those attempts? What if it takes 6 attempts instead? Should we keep this value higher to ensure we can safely get the confirmation from RPC

marcelosalloum: Good point! I updated the code to calculate the `retryCount` and `retryAfter` based on the transaction timebounds: https://github.com/stellar/wallet-backend/pull/160/commits/dcc6e2bce15257bd6f3469699552d049c9d82d13

aditya1702: Ohh nice idea. Looks good to me

Issue #144, opened 2025-04-16
Bugfix: keep blocked operations in sync between the `/tss/transactions/build` and `/tx/create-fee-bump` handlers and make it configurable


marcelosalloum:
### What version are you using?

https://github.com/stellar/wallet-backend/commit/d208e9d6657555b5b207f410e088b573fc13dd03

### What did you do?

I tried to build and then fee bump a transaction with certain operation types.

### What happened?

The txBuild part succeeded but the flow was rejected in the feeBumpTx part. I can see the feeBumpTx obeys the following list of blocked operations:
https://github.com/stellar/wallet-backend/blob/f0b66b9e2d2b65be05253dead12be11588633fdf/internal/serve/serve.go#L37-L49

This doesn't make sense for a few reasons: 
1. The feeBump can be more permissive and the txBuild part is the one that needs to be more strict.
2. If anything, we want them to behave consistently, so we should apply the same list to all
3. This list should actually be configurable, so each instance can decide which types they support.

### What did you expect to see instead?

- Consistency between the methods.
- Nice-to-have: configurable types.

marcelosalloum:
Closed by #167

PR #167, opened 2025-05-01
integrationtests: complete the integration tests happy path

### LINKED ISSUE #149: Feature Request: implement e2e integration tests

marcelosalloum:
### What

Complete the integration tests happy path by increasing the number of operations covered by the integration tests and remove the blocked list of operations.

### Why

Closes #149, #144

#### PR Structure

#### Thoroughness

#### Release


internal/integrationtests/helpers.go
```diff
@@ -93,3 +91,29 @@ func SCAccountID(address string) (xdr.ScAddress, error) {
...
+func NewSet[T comparable](values ...T) *Set[T] {
+	set := &Set[T]{set: make(map[T]struct{})}
+	for _, value := range values {
+		set.Add(value)
+	}
+	return set
+}
+
+func (s *Set[T]) Add(value T) {
+	if !reflect.ValueOf(&value).Elem().IsZero() {
```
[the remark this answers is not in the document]
marcelosalloum: Since we're narrowing it to `comparable` types, your suggestion makes sense.

Addressed in https://github.com/stellar/wallet-backend/pull/167/commits/b74290660259e6e41971753a6b2b738e84e00d64

internal/integrationtests/fixtures.go
```diff
@@ -7,73 +7,366 @@ import (
...
+			Name:          "foo",
+			Value:         []byte("bar"),
+			SourceAccount: f.PrimaryAccountKP.Address(),
+		},
+		&txnbuild.EndSponsoringFutureReserves{
+			SourceAccount: newAccountKP.Address(),
+		},
+	}
+
+	b64OpsXDRs := make([]string, len(operations))
```
aditya1702, line 112: Code quality nit: The code in lines 112-124 are repeated across the functions. We can abstract them into a separate function instead

internal/integrationtests/integrationtests.go
```diff
@@ -75,161 +79,185 @@ func (it *IntegrationTests) Run(ctx context.Context) error {
...
 
-func (it *IntegrationTests) signTransactions(ctx context.Context, builtTxResponse *types.BuildTransactionsResponse) ([]string, error) {
-	signedTxXDRs := make([]string, len(builtTxResponse.TransactionXDRs))
+// RenderResult renders a result string for a use case.
+func RenderResult(useCase *UseCase) string {
+	status := useCase.getTransactionResult.Status
+	var statusEmoji string
+	switch status {
+	case entities.SuccessStatus:
+		statusEmoji = "✅"
```
aditya1702, line 191: Haha I really like your usage of emojis

aditya1702 (commented):
Nothing major. Overall looks good. I had a question around how the integration tests are using the wallet backend code - wouldnt it be better to spin up wallet-backend container using the dockerfile and test against it using http requests to its endpoints? That way it would mimic the actual usage in real-world env?

So in that case we wont have to add a specific `wBclient` to call the functions.

aditya1702 (approved):

internal/integrationtests/fixtures.go
marcelosalloum: https://github.com/stellar/wallet-backend/pull/167/commits/2f5a3f2280cb9ec289a2224a57cb18aee64d1148

PR #180, opened 2025-05-13
{ingestor, store}: eagerly unlock channel accounts when their txHash is ingested

marcelosalloum:
### What

Eagerly unlock channel accounts when their txHash is ingested.

### Why

So channel accounts became available sooner.

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/signing/store/channel_accounts_model.go
+++ b/internal/signing/store/channel_accounts_model.go
@@ -93,9 +93,9 @@ func (ca *ChannelAccountModel) AssignTxToChannelAccount(ctx context.Context, pub
 	return nil
 }
 
-func (ca *ChannelAccountModel) UnassignTxAndUnlockChannelAccounts(ctx context.Context, txHashes ...string) error {
+func (ca *ChannelAccountModel) UnassignTxAndUnlockChannelAccounts(ctx context.Context, txHashes ...string) (int64, error) {
 	if len(txHashes) == 0 {
-		return errors.New("txHashes cannot be empty")
+		return 0, errors.New("txHashes cannot be empty")
 	}
 
 	const query = `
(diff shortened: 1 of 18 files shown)


internal/signing/store/channel_accounts_model.go
```diff
@@ -93,13 +93,30 @@ func (ca *ChannelAccountModel) AssignTxToChannelAccount(ctx context.Context, pub
...
 
-func (ca *ChannelAccountModel) UnassignTxAndUnlockChannelAccount(ctx context.Context, txHash string) error {
-	const query = `UPDATE channel_accounts SET locked_tx_hash = NULL, locked_at = NULL, locked_until = NULL WHERE locked_tx_hash = $1`
-	_, err := ca.DB.ExecContext(ctx, query, txHash)
+func (ca *ChannelAccountModel) UnassignTxAndUnlockChannelAccounts(ctx context.Context, txHashes ...string) (int64, error) {
+	if len(txHashes) == 0 {
+		return 0, errors.New("txHashes cannot be empty")
+	}
+
+	const query = `
```
aristidesstaffieri, line 101: it could help query performance here to use a cte with unnest for the txHashes, what about something like this - 
```
const query = `
	WITH txs AS (
		SELECT UNNEST($1::text[]) AS tx_hash
	)
	UPDATE channel_accounts
	SET
		locked_tx_hash = NULL,
		locked_at = NULL,
		locked_until = NULL
	FROM txs
	WHERE channel_accounts.locked_tx_hash = txs.tx_hash
`
```

aristidesstaffieri: we talked offline and decided that for this case in practice the perf would be very similar so wont make the change.

aristidesstaffieri (approved):

PR #184, opened 2025-05-15
Remove TSS

aditya1702:
### What

- Remove all the code related to TSS.
- Change the `/tss/build` endpoint to `/transactions/build`

### Why

The wallet engineering team decided to remove TSS from the backend. This was part of rethinking the features that wallet-backend should serve. Clients would be able to build transactions and then submit them themselves via self-hosted RPC or 3rd party providers.

### Known limitations

N/A

#182 

#### PR Structure

#### Thoroughness

#### Release


internal/serve/serve.go
```diff
@@ -359,30 +243,26 @@ func handler(deps handlerDeps) http.Handler {
...
+			r.Post("/create-fee-bump", accountHandler.CreateFeeBumpTransaction)
 		})
 
-		r.Route("/tss", func(r chi.Router) {
-			handler := &httphandler.TSSHandler{
-				Router:             deps.TSSRouter,
-				Store:              deps.TSSStore,
+		r.Route("/transactions", func(r chi.Router) {
+			handler := &httphandler.TransactionsHandler{
+				TransactionService: deps.TransactionService,
```
marcelosalloum, line 259: I think the prefix are redundant here:
- /tx/create-sponsored-account
- /tx/create-fee-bump
- /transactions/build

I think building and fee bumping are closely related and should likely be together.

In terms of creating a sponsored account, maybe that could move under `/accounts/...`. WDYT?

Considering the PR is already pretty big, these changes could be done in a follow-up PR.

internal/serve/serve.go
```diff
@@ -115,14 +89,6 @@ func Serve(cfg Configs) error {
 		Handler:    handler(deps),
 		OnStarting: func() {
 			log.Infof("🌐 Starting Wallet Backend server on port %d", cfg.Port)
-			go populatePools(ctx, deps.PoolPopulator)
-		},
-		OnStopping: func() {
-			log.Info("Stopping Wallet Backend server")
```
marcelosalloum, line 121: I find it useful to log when the server is stopping. It may be useful when looking at server logs

.github/workflows/go.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
         run: |
           go install golang.org/x/tools/cmd/goimports@v0.31.0
           # Find all .go files excluding paths containing 'mock' and run goimports
-          non_compliant_files=$(find . -type f -name "*.go" ! -path "*mock*" | xargs goimports -local "github.com/stellar/stellar-disbursement-platform-backend" -l)
+          non_compliant_files=$(find . -type f -name "*.go" ! -path "*mock*" | xargs goimports -local "github.com/stellar/wallet-backend" -l)
```
marcelosalloum, line 58: Thanks for catching that!

marcelosalloum (commented):
Looking good so far

internal/serve/httphandler/transactions_handler.go
```diff
@@ -0,0 +1,70 @@
...
+	var transactionXDRs []string
+	for i, transaction := range reqParams.Transactions {
+		ops, err := transactionsUtils.BuildOperations(transaction.Operations)
+		if err != nil {
+			httperror.BadRequest("", map[string]any{
+				fmt.Sprintf("transactions[%d].operations", i): err.Error(),
+			}).Render(w)
+			return
+		}
+		tx, err := t.TransactionService.BuildAndSignTransactionWithChannelAccount(ctx, ops, transaction.Timeout, transaction.SimulationResult)
```
aristidesstaffieri, line 45: I'm wondering about the case of a partial failure. If a few xdrs are built and signed with a channel account but then an xdr later in the list fails, should we be releasing the lock on the channel accounts for the ones that succeeded since there is no partial response and all xdrs will need to be re-submitted?

aristidesstaffieri: sounds good!

marcelosalloum: PR created here: https://github.com/stellar/wallet-backend/pull/186

PR #188, opened 2025-06-09
Remove TSS

aditya1702:
### What

- Remove all the code related to TSS.
- Change the `/tss/build` endpoint to `/transactions/build`

### Why

The wallet engineering team decided to remove TSS from the backend. This was part of rethinking the features that wallet-backend should serve. Clients would be able to build transactions and then submit them themselves via self-hosted RPC or 3rd party providers.

### Known limitations

N/A

#182 

#### PR Structure

#### Thoroughness

#### Release


Makefile
marcelosalloum, line 1: 👏

internal/db/migrations/2025-01-22.1-remove_tss_tables.sql
```diff
@@ -0,0 +1,37 @@
+-- +migrate Up
+
+DROP INDEX IF EXISTS idx_tx_current_status;
+DROP INDEX IF EXISTS idx_claimed_until;
+DROP TABLE IF EXISTS tss_transactions;
+
+DROP INDEX IF EXISTS idx_original_transaction_hash;
+DROP INDEX IF EXISTS idx_last_updated;
+DROP TABLE IF EXISTS tss_transaction_submission_tries;
+
```
marcelosalloum, line 10: FYI: if you were to drop the table with the `CASCADE` option, all dependent objects – including indexes, constraints, and triggers – are automatically dropped as well.

marcelosalloum (approved):
LGTM! Great job, very thorough!

PR #192, opened 2025-06-11
Add contract ID cache and DB table

### LINKED ISSUE #190: State Change Indexer: Add Contract Cache Store and DB Tables
This is an optimization where we store a DB backend in-memory cache:

```
contractID : {
      name
      symbol
      expiry
}
```

This helps store the `name` and `symbol` values for contract tokens and avoid calling the simulation endpoint on RPC each time to fetch the information. Each item will have an expiry after which we will refresh the cache again using `simulateTransaction` calls.

We will also store the mapping in the postgres table to prevent loss of in-memory cache in case of restarts/shutdowns

aditya1702:
### What

- Add a `contracts` DB table which stores the `name` and `symbol` of custom tokens.
- Add an in-memory cache that maintains the contracts in local memory to speed up lookups.

```
contractID : {
     name
     symbol
}
``` 
This mapping is stored as a hash map structure in memory. We use the [go-cache](https://github.com/patrickmn/go-cache) library for setting up the cache.

- Add relevant unit tests

For pod restarts, the cache will be initialized from the DB which acts a persistent data store.

### Why

We dont need to call the simulation function each time to get the relevant name and symbol values for custom tokens. We can simply store the information in the cache when processing token transfer events for these tokens and use the info by querying against the cache.

### Known limitations

N/A

Closes #190  #206 #207 

#### PR Structure

#### Thoroughness

#### Release


JakeUrban:
I'm not convinced the benefits we're getting by caching name and symbol is worth adding an additional component to the architecture. We want this project to be as easy to run for other organizations as possible.

internal/data/contracts.go
marcelosalloum, line 1: Can we rename Contract to TokenContract, for disambiguation? I understand will trigger renaming of multiple files and objects

internal/data/contracts.go
```diff
@@ -0,0 +1,66 @@
...
+	DB             db.ConnectionPool
+	MetricsService metrics.MetricsService
+}
+
+type Contract struct {
+	ID        string    `db:"id" json:"id"`
+	Name      string    `db:"name" json:"name"`
+	Symbol    string    `db:"symbol" json:"symbol"`
+	CreatedAt time.Time `db:"created_at" json:"createdAt"`
+	UpdatedAt time.Time `db:"updated_at" json:"updatedAt"`
```
marcelosalloum, line 22: 💡 Any thoughts about adding an expiration field to the database as well? Do you think it's unnecessary?

internal/data/contracts.go
marcelosalloum, line 1: Still missing the unit tests for `data/contracts.go`, right?

internal/db/migrations/2025-06-13.0-create_contract_table.sql
```diff
@@ -0,0 +1,18 @@
...
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+CREATE TRIGGER contracts_set_updated_at BEFORE UPDATE ON contracts FOR EACH ROW EXECUTE PROCEDURE
+  refresh_updated_at_column();
+
+-- +migrate Down
+
+DROP TRIGGER contracts_set_updated_at ON contracts;
+DROP TABLE contracts;
```
marcelosalloum, line 18: 💡 FYI: if you drop the table with the cascade option, the triggers will be automatically dropped as well.

internal/data/contracts.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	return contract, nil
+}
+
+func (m *ContractModel) Insert(ctx context.Context, tx db.Transaction, contract *Contract) error {
+	query := `
+		INSERT INTO contracts (id, name, symbol)
+		VALUES (:id, :name, :symbol)
+	`
+	_, err := tx.NamedExecContext(ctx, query, contract)
```
marcelosalloum, line 39: I see you liked the named parameters 🚀

internal/serve/serve.go
marcelosalloum, line 1: 🤔  Is the contract store needed by the `serve` package?

marcelosalloum (commented):
Looking great!

I left a few comments about code completeness and a question regarding database expiration. We can sync by DM if you want.

internal/data/contracts.go
aditya1702: @marcelosalloum I figured since all its functions are being tested indirectly through the [contracts_test.go](https://github.com/stellar/wallet-backend/pull/192/files#diff-6ce36609adf4c97fd9444e310a74a2aada2797c8be6c7a5ab9e999f7c9bbbb95). Do you still think I should add separate tests?

internal/data/contracts.go
aditya1702: Hmm, I dont see an obvious case for expiration value here. Technically, entries will not be removed from db right? We just update the details for existing contracts there.

internal/serve/serve.go
marcelosalloum: Yeah, I'm not 100% sure this will be needed when serving the API.

My initial thought was that once we ingest an asset we keep its information as is in the state_changes DB table. I'd suggest we try to avoid using this complexity in the API until/unless we get an actual request for it.

internal/data/contracts.go
marcelosalloum: I was more thinking about achieving consistency between the in-memory cache and the DB info.
I don't think it's mandatory but a nice-to-have from my PoV.

internal/data/contracts.go
marcelosalloum: I personally like to have explicit coverage in data models, especially when we have complex queries. Since these queries are relatively simple, I agree it's not mandatory as long as the edge cases are being covered in [contracts_test.go](https://github.com/stellar/wallet-backend/pull/192/files#diff-6ce36609adf4c97fd9444e310a74a2aada2797c8be6c7a5ab9e999f7c9bbbb95) as well.

aditya1702: Ok fair enough. I think all the edge cases are being covered through the contract cache tests

PR #197, opened 2025-06-18
services: Implement call to `getLedgers` and start parsing the `xdr.LedgerCloseMeta`

### LINKED ISSUE #200: Replace calls to {RPC}.getTransactions with {RPC}.getLedgers, allowing retrieval of one or multiple ledgers at a time

### LINKED ISSUE #201: Parallelize the processing while still ensuring proper order for transaction, operation and state changes.
The ingestor work can be parallelized using sync muxes or a 3rd-party library like pond.

While parallelizing, we should still be able to order things properly by using:
1. `ledger_created_at` or `ledgerSequence` can be used to order the **ledgers**
2. `toid` can be used to order **operations** and even **transactions** (as shown in the [participants_processor](https://github.com/stellar/go/blob/ef10cd0a592065357e7dff199401c38494b461bf/services/horizon/internal/ingest/processors/participants_processor.go#L114))
3. **state changes** can use a conjunction of `operation_id` (toid) and their `type` and affected `account`. This needs further discussion.

marcelosalloum:
### What

Update the indexer code to stream from `{RPC}.getLedgers` and break down the results from ledgerRange -> ledgers -> transactions.

Also, (pseudo) process the ledgers in parallel.

### Why

As part of the new indexer implementation.

### Future Work

- [ ] evaluate transactions, filter by participant account & store in the DB
- [ ] evaluate operations, filter by participant account & store in the DB
- [ ] generate state changes
- [ ] unlock channel accounts

I haven't added many tests to the ingest service yet, since this code will still evolve quite a bit.

Closes #200
Closes #201 

#### PR Structure

#### Thoroughness

#### Release


internal/services/ingest.go
```diff
@@ -155,6 +169,212 @@ func (m *ingestService) Run(ctx context.Context, startLedger uint32, endLedger u
...
+	var allTxHashes []string
+	var errs []error
+
+	// Submit tasks to the pool
+	for _, ledger := range getLedgersResponse.Ledgers {
+		ledger := ledger // Create a new variable to avoid closure issues
+		pool.Submit(func() {
+			txHashes, err := m.processLedger(ctx, ledger)
+			mu.Lock()
+			if err != nil {
```
marcelosalloum, line 310: I plan to use toid to order the transactions in the database. It's calculated like [this](https://github.com/stellar/go/blob/ef10cd0a592065357e7dff199401c38494b461bf/processors/operation/operation.go#L49).

internal/services/ingest.go
```diff
@@ -155,6 +169,212 @@ func (m *ingestService) Run(ctx context.Context, startLedger uint32, endLedger u
...
+// - the ledgers available in the RPC,
+// - the latest ledger synced by the ingestion service,
+// - the max ledger window to ingest.
+//
+// The returned ledger sequence range is inclusive of the start and end ledgers.
+func getLedgerSeqRange(rpcOldestLedger, rpcNewestLedger, latestLedgerSynced uint32) (ledgerRange LedgerSeqRange, inSync bool) {
+	if latestLedgerSynced >= rpcNewestLedger {
+		return LedgerSeqRange{}, true
+	}
+	ledgerRange.StartLedger = max(latestLedgerSynced+1, rpcOldestLedger)
```
aditya1702, line 288: If we do a `max()` here would that mean if wallet-backend is at 5 and rpc's oldest ledger is at 8, it would skip ingesting 6, 7? If we are ingesting using `GetLedgers` by getting ledgers in bulk, do we want to allow WB  catch up to RPC instead of skipping ledgers?

internal/services/ingest.go
```diff
@@ -155,6 +169,212 @@ func (m *ingestService) Run(ctx context.Context, startLedger uint32, endLedger u
...
+	ledgerRange.Limit = getLedgersLimit
+
+	return ledgerRange, false
+}
+
+func (m *ingestService) processLedgerResponse(ctx context.Context, getLedgersResponse GetLedgersResponse) error {
+	// Create a worker pool with
+	const poolSize = 16
```
aditya1702, line 296: nit: lets move this with other constants at the top

internal/services/ingest.go
```diff
@@ -155,6 +169,212 @@ func (m *ingestService) Run(ctx context.Context, startLedger uint32, endLedger u
...
+	transactions, err := m.getLedgerTransactions(ctx, xdrLedgerCloseMeta)
+	if err != nil {
+		return nil, fmt.Errorf("getting ledger transactions: %w", err)
+	}
+
+	// TODO: process transactions
+	log.Ctx(ctx).Debugf("🚧 Got %d transactions for ledger %d", len(transactions), xdrLedgerCloseMeta.LedgerSequence())
+	txHashes := make([]string, 0, len(transactions))
+	for _, tx := range transactions {
+		txHashes = append(txHashes, tx.Hash.HexString())
```
aditya1702, line 349: Since this is still in TODO, I am assuming this will return a list of `types.Transaction` structs instead of tx hash strings? Because what would be the use of the tx hashes?

internal/services/ingest.go
marcelosalloum: https://github.com/stellar/wallet-backend/pull/197/commits/21b716821ddc575dde8beb42091132ef60522136

aditya1702 (approved):

PR #221, opened 2025-06-25
[WIP]: Implement watched accounts' bloom filter

### LINKED ISSUE #217: Cache watched accounts to reduce memory usage and improve processing time

marcelosalloum:
### What

Implement watched accounts bloom filter

### Why

So we can quickly identify which accounts the wallet backend is watching and only process transactions, operations, and state changes pertaining to those accounts.

This should save both processing time and RAM when processing the transactions metadata.

### Future work

Apply this to the transactions, operations, and state changes processors

Closes #217 

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/indexer/watched_accounts_cache.go
+++ b/internal/indexer/watched_accounts_cache.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"sync"
 	"sync/atomic"
 	"syscall"
 	"time"
@@ -20,6 +21,7 @@ type WatchedAccountsCache struct {
 	reloadInterval time.Duration
 	reloadFn       ReloadFn
 	quit           chan struct{}
+	closeOnce      sync.Once
 }
 
 func NewWatchedAccountsCache(ctx context.Context, reloadInterval time.Duration, reloadFn ReloadFn) *WatchedAccountsCache {
@@ -68,6 +70,7 @@ func (wc *WatchedAccountsCache) periodicRebuild(ctx context.Context) {
 func (wc *WatchedAccountsCache) DoRebuild(ctx context.Context) {
 	accounts, err := wc.reloadFn(ctx)
 	if err != nil {
+		// If the rebuild fails, we cannot rely on the cache anymore, so we panic to avoid a situation where the cache is in an inconsistent state.
 		panic(fmt.Errorf("failed to reload watched accounts: %w", err))
 	}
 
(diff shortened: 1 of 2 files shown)


internal/indexer/watched_accounts_cache.go
```diff
@@ -0,0 +1,94 @@
...
+
+		case <-wc.quit:
+			log.Ctx(ctx).Infof("watched accounts cache stopped due to quit signal")
+			return
+
+		case <-ticker.C:
+			wc.DoRebuild(ctx)
+		}
+	}
+}
```
marcelosalloum, line 64: The bloom filter will periodically rebuild using the provided method. We'll use `AccountsModel.GetAll` to rebuild it.

internal/indexer/watched_accounts_cache_test.go
```diff
@@ -0,0 +1,180 @@
...
+				defer mu.Unlock()
+				falsePositives++
+			}
+		}()
+	}
+	wg.Wait()
+
+	falsePositiveRatio := float64(falsePositives) / float64(testCount)
+	t.Logf("False positives: %d/%d = %.4f (expected ~0.01)", falsePositives, testCount, falsePositiveRatio)
+	assert.InDelta(t, 0.01, falsePositiveRatio, 0.005) // Should be between [0.005, 0.015]
```
marcelosalloum, line 119: This test ensures the cache can be used concurrently and that the false positive ratio is roughly the configured amount of ~1%.

internal/indexer/watched_accounts_cache.go
```diff
@@ -0,0 +1,99 @@
...
+	wc := &WatchedAccountsCache{
+		reloadInterval: reloadInterval,
+		reloadFn:       reloadFn,
+		quit:           make(chan struct{}),
+	}
+	go wc.periodicRebuild(ctx)
+	return wc
+}
+
+// periodicRebuild is a goroutine that periodically reloads the watched accounts and updates the bloom filter.
```
JakeUrban, line 37: Why do we need to rebuild the filter by querying the DB? I figured the filter would be updated using the following flow:

- on startup, get all accounts from the DB and build the filter
- when a new account is registered, immediately add it to the bloom filter and insert it into the DB
- when an account is deregistered, immediately remove it from the filter and delete it from the DB

Rebuilding from the DB implies that accounts are getting added to the DB without getting added to the filter at the same time.

JakeUrban: I can see how if the ingestor and API were running within different processes or machines, we would need some kind of shared data storage like postgres or redis, but I think we should try to support running both the ingestor and API (or other services we build in the future, like webhooks) from within the same process first. We'll probably need to support splitting them into different processes eventually for scalability purposes, but maybe we can keep it as one process for now?

We should be able to do this by running these services in different goroutines and then using channels to send messages (ex. when a new account is registered).

PR #240, opened 2025-07-12
Process and ingest state changes in the DB

aditya1702:
### What

### Why

### Known limitations

[TODO or N/A]

#### PR Structure

#### Thoroughness

#### Release


internal/db/migrations/2025-06-10.4-create_indexer_table_state_changes.sql
```diff
@@ -9,8 +9,8 @@ CREATE TABLE state_changes (
     ledger_created_at TIMESTAMPTZ NOT NULL,
     ledger_number INTEGER NOT NULL,
     account_id TEXT NOT NULL REFERENCES accounts(stellar_address),
-    operation_id BIGINT NOT NULL REFERENCES operations(id),
-    tx_hash TEXT NOT NULL REFERENCES transactions(hash),
```
marcelosalloum, line 13: Shouldn't we keep referencing the tx_hash here?

internal/indexer/indexer.go
```diff
@@ -43,13 +43,17 @@ func (i *Indexer) ProcessTransaction(transaction ingest.LedgerTransaction) error
...
+		// 	return fmt.Errorf("creating data operation: %w", err)
+		// }
 
 		for participant := range opParticipants.Participants.Iter() {
-			i.IndexerBuffer.PushParticipantOperation(participant, *dataOp, *dataTx)
+			i.IndexerBuffer.PushParticipantOperation(participant, opID, opParticipants.Operation, transaction, opParticipants.OperationIdx)
+
+			// Even though the participant is not part of a transaction, we still want to index the transaction since the participant
+			// is part of one of its operations.
+			i.IndexerBuffer.PushParticipantTransaction(participant, transaction)
```
marcelosalloum, line 56: 💡  The method `PushParticipantOperation` already injects the transaction as a collateral through the method `pushParticipantTransactionUnsafe`.

internal/services/ingest.go
```diff
@@ -389,33 +397,55 @@ func (m *ingestService) ingestProcessedData(ctx context.Context, ledgerIndexer *
...
 				}
-				stellarAddressesByTxHash[tx.Hash].Add(participant)
+				stellarAddressesByTxHash[dataTx.Hash].Add(participant)
+
+				tokenTransferChanges, err := m.tokenTransferProcessor.ProcessTransaction(ctx, tx)
+				if err != nil {
+					return fmt.Errorf("processing token transfer changes: %w", err)
+				}
+
+				stateChanges = append(stateChanges, tokenTransferChanges...)
```
marcelosalloum, line 423: You may be duplicating things here because this transaction could have been porcesed beforehand. The reason being we are looping through participants and then tx-per-participant, and often we have the same transaction connected to multiple participants.

When I designed the indexer, my intention was to process everything there (once), and then we filter things off here in the `ingestProcessedData` method.

internal/services/ingest.go
```diff
@@ -389,33 +397,55 @@ func (m *ingestService) ingestProcessedData(ctx context.Context, ledgerIndexer *
 		opByID := make(map[int64]types.Operation)
 		stellarAddressesByOpID := make(map[int64]set.Set[string])
 
+		stateChanges := make([]types.StateChange, 0)
+
 		// 1. Build the data structures needed for the DB insertions
 		for participant := range indexerBuffer.Participants.Iter() {
-			if !indexerBuffer.Participants.Contains(participant) {
-				continue
-			}
```
marcelosalloum, line 396: If we process everything in the indexer, we can re-add this filter here to decrease the data set being sent to the SQL queries.

marcelosalloum (commented):
I know it's still in draft, but I'm adding some comments here so we can recalibrate before you invest more time on the task.

internal/db/migrations/2025-06-10.4-create_indexer_table_state_changes.sql
aditya1702: @marcelosalloum So we are running into an issue here where state changes for contract invocations are processed but we did not insert those transactions because we dont identify the contract participants correctly. So this is leading to an error when we insert the state change but the txn hash has not been inserted.

And for the operation ID, fee events dont have an operation associated with them so we get an error when inserting those fee state changes

internal/indexer/indexer.go
aditya1702: Ahhh! Completely didnt notice that. Will remove this

internal/services/ingest.go
aditya1702: Ohh yeah good catch! Looks like it works for the txns and participants maps becuase they use a set so we can add duplicates without any issue. But let me resolve this

internal/services/ingest.go
aditya1702: I think the IF condition should be a check with our future bloom filter/in-memory account cache

PR #252, opened 2025-07-22
indexer: identify `DEPLOY` state changes

marcelosalloum:
### What

Identify `DEPLOY` state changes.

### Why

This was still missing. in our state changes mapping

#### PR Structure

#### Thoroughness

#### Release


internal/indexer/indexer.go
```diff
@@ -88,6 +91,13 @@ func (i *Indexer) ProcessTransaction(ctx context.Context, transaction ingest.Led
 			return fmt.Errorf("processing effects state changes: %w", err)
 		}
 		i.Buffer.PushStateChanges(effectsStateChanges)
+
+		// 2.2. Index contract deploy state changes
+		contractDeployStateChanges, err = i.contractDeployProcessor.ProcessOperation(ctx, opParticipants.OpWrapper)
```
aditya1702, line 96: Small design idea: since we are using the same interface here, we can instead have a list with these two processors and use a for loop to push the state changes. So something like this:

```
operationProcessors []OperationProcessorInterface

for _, processor: range b.operationProcessors:
     stateChanges, err := processor.ProcesssOperation...
     ....
     i.Buffer.PushStateChanges(stateChanges)
```

internal/indexer/mocks.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/stellar/wallet-backend/internal/indexer/types"
 )
 
-type EffectsProcessorInterface interface {
+type OperationProcessorInterface interface {
 	ProcessOperation(ctx context.Context, opWrapper *operation_processor.TransactionOperationWrapper) ([]types.StateChange, error)
```
aditya1702, line 16: nit: I actually forgot to move this to indexer.go file but normally interfaces should be defined in the place they are used. Can you please move this there with the other interfaces

internal/indexer/processors/contract_deploy.go
```diff
@@ -0,0 +1,110 @@
...
+
+	opID := op.ID()
+	builder := NewStateChangeBuilder(op.Transaction.Ledger.LedgerSequence(), op.LedgerClosed.Unix(), op.Transaction.Hash.HexString()).
+		WithOperationID(opID).
+		WithCategory(types.StateChangeCategoryContract).
+		WithReason(types.StateChangeReasonDeploy)
+
+	deployedContractsMap := map[string]types.StateChange{}
+
+	processCreate := func(fromAddr xdr.ContractIdPreimageFromAddress) error {
```
aditya1702, line 37: I am just curious - what is the recommended way here: do we create a private function as a variable or do we assign it to the ContractDeployProcessor struct. Normally I have always done the latter 🤔

internal/indexer/processors/contract_deploy_test.go
```diff
@@ -0,0 +1,259 @@
...
+		WithCategory(types.StateChangeCategoryContract)
+
+	type TestCase struct {
+		name             string
+		op               *operation_processor.TransactionOperationWrapper
+		wantStateChanges []types.StateChange
+	}
+	testCases := []TestCase{}
+
+	for _, withSubinvocations := range []bool{false, true} {
```
aditya1702, line 52: Really good comprehensive tests 👏 . Would you mind adding some comments here for someone who is not well versed with the XDR objects and contract ops. That would be helpful

aditya1702 (approved):
LGTM. It would be nice to have @JakeUrban review the contract parsing once just to make sure it looks correct.

internal/indexer/indexer.go
marcelosalloum: I agree with that change but I think it's outside the scope of this PR. It should be done in a dedicated PR.

marcelosalloum: I opened a PR [here](https://github.com/stellar/wallet-backend/pull/254), but in reality this won't scale for more custom contract processors.

[This discussion](https://stellarfoundation.slack.com/archives/C030Z9EHVQE/p1753204684005249) points to the necessity of calling RPC to assert contract interfaces and likely adding a special cache for contracts. This cache can expand a lot and we'll probably need some space-efficient filter, like BloomFilter.

marcelosalloum: I'm leaving that PR in draft. Feel free to either close it or build on top of it.

internal/indexer/processors/contract_deploy.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			return fmt.Errorf("deployer address to string: %w", err)
+		}
+
+		deployedContractsMap[contractID] = builder.Clone().WithAccount(contractID).WithDeployer(deployerAddr).Build()
+		return nil
+	}
+
+	var walkInvocation func(inv xdr.SorobanAuthorizedInvocation) error
+	walkInvocation = func(inv xdr.SorobanAuthorizedInvocation) error {
```
JakeUrban, line 52: I didn't realize this `walkInvocation` pattern was [provided by the JS SDK](https://stellar.github.io/js-stellar-sdk/global.html#walkInvocationTree). I wonder if we can request this to be added to the txnbuild golang package?

internal/indexer/processors/contract_deploy.go
```diff
@@ -0,0 +1,110 @@
...
+func (p *ContractDeployProcessor) ProcessOperation(_ context.Context, op *operation_processor.TransactionOperationWrapper) ([]types.StateChange, error) {
+	if op.OperationType() != xdr.OperationTypeInvokeHostFunction {
+		return nil, ErrInvalidOpType
+	}
+	invokeHostOp := op.Operation.Body.MustInvokeHostFunctionOp()
+
+	opID := op.ID()
+	builder := NewStateChangeBuilder(op.Transaction.Ledger.LedgerSequence(), op.LedgerClosed.Unix(), op.Transaction.Hash.HexString()).
+		WithOperationID(opID).
+		WithCategory(types.StateChangeCategoryContract).
```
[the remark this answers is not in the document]
marcelosalloum: Screenshot 2025-07-23 at 5 07 35 PM

(G) Account creation is generating these state changes, but for contracts there's no concept of DEBIT/CREDIT in play, it's just fees and deployer so I think the states being changed are different even though they can both be interpreted as account creation.

I'm still inclined to keep some differentiation in G-account and C-account creation because the nature of the accounts are different.

internal/indexer/processors/contract_deploy.go
marcelosalloum: `walkInvocation` is a common pattern to traverse node-like structures like sub-invocations.

internal/indexer/processors/contract_deploy.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+	case xdr.HostFunctionTypeHostFunctionTypeUploadContractWasm, xdr.HostFunctionTypeHostFunctionTypeInvokeContract:
+		// no-op
+	}
+
+	for _, auth := range invokeHostOp.Auth {
+		if err := walkInvocation(auth.RootInvocation); err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
marcelosalloum: > Just to confirm we're on the same page, you have the walkInvocation calls outside of the switch because both InvokeContract and CreateContractV2 can deploy many contracts, right?

Yes, and since there are (possibly) nested calls, it needs to be able to call itself for the child nodes.

internal/indexer/processors/contract_deploy.go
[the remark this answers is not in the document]
JakeUrban: Ok, well even if we keep a “Contract”/“Deploy” classification/reason for contract account creation, we don’t have a state change for stellar account creation. Meaning, a client querying for state changes wouldn’t be able to tell when the account was created. They’d have to inspect the operation or transaction. If we have a state change representing account creation for contracts, I think we should have one for classic accounts. Wdyt?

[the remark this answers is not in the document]
marcelosalloum: Yeah, that makes sense to me 👍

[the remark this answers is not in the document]
JakeUrban: SGTM!

internal/indexer/indexer.go
```diff
@@ -89,6 +97,13 @@ func (i *Indexer) ProcessTransaction(ctx context.Context, transaction ingest.Led
...
+		if err != nil && !errors.Is(err, processors.ErrInvalidOpType) {
+			return fmt.Errorf("processing contract deploy state changes: %w", err)
+		}
```
JakeUrban, line 105: What if `errors.Is(err, processors.ErrInvalidOpType)` is true?

internal/indexer/processors/contract_deploy.go
```diff
@@ -0,0 +1,110 @@
...
+		deployedContractsMap[contractID] = builder.Clone().WithAccount(contractID).WithDeployer(deployerAddr).Build()
+		return nil
+	}
+
+	var walkInvocation func(inv xdr.SorobanAuthorizedInvocation) error
+	walkInvocation = func(inv xdr.SorobanAuthorizedInvocation) error {
+		switch inv.Function.Type {
+		case xdr.SorobanAuthorizedFunctionTypeSorobanAuthorizedFunctionTypeCreateContractHostFn:
+			cc := inv.Function.MustCreateContractHostFn()
+			if cc.ContractIdPreimage.Type == xdr.ContractIdPreimageTypeContractIdPreimageFromAddress {
```
JakeUrban, line 56: What other preimage types are there and should we just skip them? Or can we safely assume the type will always be `ContractIdPreimageTypeContractIdPreimageFromAddress`?

internal/indexer/types/types.go
```diff
@@ -164,13 +167,12 @@ type StateChange struct {
...
 	TokenID            sql.NullString `json:"tokenId,omitempty" db:"token_id"`
 	Amount             sql.NullString `json:"amount,omitempty" db:"amount"`
-	ClaimableBalanceID sql.NullString `json:"claimableBalanceId,omitempty" db:"claimable_balance_id"`
-	LiquidityPoolID    sql.NullString `json:"liquidityPoolId,omitempty" db:"liquidity_pool_id"`
 	OfferID            sql.NullString `json:"offerId,omitempty" db:"offer_id"`
 	SignerAccountID    sql.NullString `json:"signerAccountId,omitempty" db:"signer_account_id"`
 	SpenderAccountID   sql.NullString `json:"spenderAccountId,omitempty" db:"spender_account_id"`
 	SponsoredAccountID sql.NullString `json:"sponsoredAccountId,omitempty" db:"sponsored_account_id"`
 	SponsorAccountID   sql.NullString `json:"sponsorAccountId,omitempty" db:"sponsor_account_id"`
+	DeployerAccountID  sql.NullString `json:"deployerAccountId,omitempty" db:"deployer_account_id"`
```
JakeUrban, line 175: If we have a deployer address, should we have a `funder` address for the create account op case?

internal/indexer/processors/contract_deploy.go
aristidesstaffieri: there is also `contractIdPreimageFromAsset`

JakeUrban: Oh ok well that’s for SAC’s which should ever be registered as contract accounts.

internal/indexer/indexer.go
marcelosalloum: We chose to send ops of different types to multiple processors, and if that processor is not able to process the incoming type (checked in the first lines of its implementation), then the processor returns early with that error.

The alternative would be adding routers or switch/cases to many parts of the code, but that would add unnecessary complexity, given the current code base.

marcelosalloum (commented):
LGTM!

internal/indexer/indexer.go
aditya1702: Yeah +1 to what Marcelo said. We simply fail the processing and it exits out of the ingestion process.

internal/indexer/processors/contract_deploy.go
aditya1702: @JakeUrban I am guessing you meant "never" registered as contract accounts

JakeUrban: Yes, sorry typo

internal/indexer/types/types.go
aditya1702: @JakeUrban Hmm, I am not strongly opinionated about it. We can add it for consistency with deployer address

PR #253, opened 2025-07-22
Move `/accounts/` REST APIs to GraphQL

akcays:
### What

Moves `/accounts/` APIs from REST to GraphQL.

### Why

As the wallet-backend indexer introduces GraphQL as part of its API design, we are converting the current REST endpoints to GraphQL for uniformity and ease of use.

### Known limitations

N/A

[#231 ](https://github.com/stellar/wallet-backend/issues/231)

#### PR Structure

#### Thoroughness

#### Release


internal/data/accounts.go
```diff
@@ -34,14 +34,23 @@ func (m *AccountModel) Get(ctx context.Context, address string) (*types.Account,
...
 
+	// Check if any rows were affected to determine if account already exists
+	rowsAffected, err := result.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("checking rows affected for address %s: %w", address, err)
+	}
+	if rowsAffected == 0 {
+		return fmt.Errorf("account %s is already registered", address)
+	}
+
```
marcelosalloum, line 53: :wave: The query has `ON CONFLICT DO NOTHING`, so the idea when applying that to the query is to ignore the error.

If you want to surface the error, I'd recommend a few changes here:
1. Remove ~`ON CONFLICT DO NOTHING`~
2. Update the error handling:
    1. Identify the DB conflict error, like done [here](https://github.com/stellar/stellar-disbursement-platform-backend/blob/3663ea2228d72227669caba01e20e0d9871c7010/internal/data/url_shortener.go#L94-L98)
    2. Return a custom error, like this one:
    https://github.com/stellar/wallet-backend/blob/11a92dce4742c79220292a6d2db93b36e6ea8e36/internal/signing/store/keypairs_model.go#L16
    3.  In the Handler/mutationResolver, you should use `errors.Is` to identify if your custom error is in the stack, and then return a specific error response, instead of returning a `500`. I think a `409 - conflict` would be a good call here.

marcelosalloum: :wave: `RowsAffected` would probably be a good call if you want to return 404 when trying to delete a non-existing account. Then you'll need to apply error handling similar to what I described above for the insertion.

internal/data/accounts_test.go
```diff
@@ -22,27 +22,53 @@ func TestAccountModelInsert(t *testing.T) {
...
+		address := keypair.MustRandom().Address()
+		
+		// First insert should succeed
+		err = m.Insert(ctx, address)
+		require.NoError(t, err)
+
+		// Second insert should fail
+		err = m.Insert(ctx, address)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "already registered")
```
marcelosalloum, line 70: 💡  you can also assert the error string with `assert.ErrorContains`. The result is virtually the same, but it has some extra checks, like ensuring `err != nil`

internal/serve/serve.go
```diff
@@ -256,18 +256,6 @@ func handler(deps handlerDeps) http.Handler {
...
-				AccountService:            deps.AccountService,
-				AccountSponsorshipService: deps.AccountSponsorshipService,
-				SupportedAssets:           deps.SupportedAssets,
-				AppTracker:                deps.AppTracker,
-			}
-
-			r.Post("/{address}", handler.RegisterAccount)
-			r.Delete("/{address}", handler.DeregisterAccount)
-		})
-
```
marcelosalloum, line 270: 💡 we can also delete the account handler objects.

marcelosalloum (commented):
The main thing missing here is proper error handling and surfacing it in the GraphQL response.

Also, I can see some `gofmt` errors. Usually IDEs Respect the formatters set in `.golangci.yml`. If you use vscode (or its forks) I can help with configuration.

internal/serve/graphql/resolvers/mutations.resolvers.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"context"
+	"fmt"
+
+	graphql1 "github.com/stellar/wallet-backend/internal/serve/graphql/generated"
+)
+
+// RegisterAccount is the resolver for the registerAccount field.
+func (r *mutationResolver) RegisterAccount(ctx context.Context, input graphql1.RegisterAccountInput) (*graphql1.RegisterAccountPayload, error) {
+	err := r.accountService.RegisterAccount(ctx, input.Address)
```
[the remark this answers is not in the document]
JakeUrban: I think you can use [strkey.IsValidContractAddress()](https://github.com/stellar/go/blob/master/strkey/main.go#L278)

internal/services/account_service.go
```diff
@@ -34,7 +38,16 @@ func NewAccountService(models *data.Models, metricsService metrics.MetricsServic
 	}, nil
 }
 
+// isValidStellarAddress validates that the address is either a valid Stellar public key or contract address
+func isValidStellarAddress(address string) bool {
+	return strkey.IsValidEd25519PublicKey(address) || strkey.IsValidContractAddress(address)
+}
```
JakeUrban, line 44: Can we rename to `isValidAddress`? "Stellar address" implies the G account format

internal/serve/graphql/resolvers/mutations.resolvers.go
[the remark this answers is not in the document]
JakeUrban: I don't think we should support muxed accounts. Their state is not onchain, so its up to the downstream application to index subaccounts.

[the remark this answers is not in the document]
akcays: I don't have enough context around the implications of supporting muxed accounts vs not, so will leave the decision to you. 

If we end up deciding to support them, I'll add `strkey.IsValidMuxedAccountEd25519PublicKey(address)` (as similar to @marcelosalloum's PR)

akcays:
> Also, I can see some gofmt errors. Usually IDEs Respect the formatters set in .golangci.yml. If you use vscode (or its forks) I can help with configuration.

@marcelosalloum thanks for your ^ comment, I could definitely use some help with the configuration. So far, I'm not getting any errors locally when I do `make check`, yet `Go / check (pull_request)` has been consistently failing for all of my commits (except for one, somehow).

internal/serve/graphql/resolvers/mutations.resolvers.go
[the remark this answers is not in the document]
JakeUrban: Marcelo and I spoke on a call, we're not going to included muxed accounts

internal/services/account_service.go
```diff
@@ -34,9 +38,21 @@ func NewAccountService(models *data.Models, metricsService metrics.MetricsServic
...
+	if !isValidAddress(address) {
+		return ErrInvalidAddress
+	}
+
 	err := s.models.Account.Insert(ctx, address)
 	if err != nil {
+		if errors.Is(err, data.ErrAccountAlreadyExists) {
+			return data.ErrAccountAlreadyExists
+		}
 		return fmt.Errorf("registering account %s: %w", address, err)
```
marcelosalloum, line 56: [Here](https://go.dev/blog/go1.13-errors)'s go spec on `errors.Is`, `errors.As`, and error wrapping.

In a nutshell:
- Always wrap errors using the `%w` verb, like in `fmt.Errorf("{message}: %w", err)` ✅ 
- it doesn't matter how many layers of wrapping you have, when you call `errors.Is`, Go will iterate over the nested wrapped erros searching for that error object, and return true if it's found
- the same happens for `errors.As`, the difference is that `errors.As` are used to parse a custom error object **type** (note that `error` is an interface), and allow you to access that custom type (as you did with the postgres error).

That said, I'd like to point to this statement:
```go
if errors.Is(err, data.ErrAccountAlreadyExists) {
	return data.ErrAccountAlreadyExists
}
```

What you're doing there is just stripping off the other errors from the stack (if any), which usually is not beneficial. You don't need those lines of code and can just proceed with returning the wrapped error (line 56). Whoever has called this method will still be able to call `if errors.Is(data.ErrAccountAlreadyExists` id it needs to handle it in a specific way.

internal/data/accounts_test.go
```diff
@@ -22,27 +23,53 @@ func TestAccountModelInsert(t *testing.T) {
...
+		address := keypair.MustRandom().Address()
+
+		// First insert should succeed
+		err = m.Insert(ctx, address)
+		require.NoError(t, err)
+
+		// Second insert should fail
+		err = m.Insert(ctx, address)
+		require.Error(t, err)
+		assert.True(t, errors.Is(err, ErrAccountAlreadyExists))
```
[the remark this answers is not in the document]
akcays: Thanks for pointing it out! I've updated to utilize `assert.ErrorIs` as it seems like it's specifically designed for error comparison and is more concise.

internal/services/account_service.go
akcays: Thank you so much for this explanation and the link to the Go blog post, these kinds of insights are really helpful for learning Go best practices.

It looks like I was unnecessarily stripping the error context. I've changed it to simply return the wrapped error, which preserves the full error chain while still allowing the GraphQL resolvers to use `errors.Is()` to identify specific error types.

PR #260, opened 2025-07-30
Move `/transactions/build/` from REST to GraphQL

akcays:
### What

Moves /transactions/build/ from REST to GraphQL.

### Why

As the wallet-backend indexer introduces GraphQL as part of its API design, we are converting the current REST endpoints to GraphQL for uniformity and ease of use.

### Known limitations

N/A

https://github.com/stellar/wallet-backend/issues/235

#### PR Structure

#### Thoroughness

#### Release


pkg/wbclient/client.go
```diff
@@ -35,9 +61,91 @@ func NewClient(baseURL string, requestSigner auth.HTTPRequestSigner) *Client {
...
 func (c *Client) BuildTransactions(ctx context.Context, transactions ...types.Transaction) (*types.BuildTransactionsResponse, error) {
-	buildTxRequest := types.BuildTransactionsRequest{Transactions: transactions}
+	if len(transactions) == 0 {
+		return nil, fmt.Errorf("at least one transaction is required")
+	}
+
+	// We only support building one transaction at a time
+	if len(transactions) > 1 {
+		return nil, fmt.Errorf("GraphQL buildTransactions mutation supports only one transaction at a time")
+	}
```
JakeUrban, line 71: Why not just update the schema to only support one transaction?

internal/serve/graphql/resolvers/resolver.go
```diff
@@ -10,6 +10,7 @@ package resolvers
 import (
 	"github.com/stellar/wallet-backend/internal/data"
 	"github.com/stellar/wallet-backend/internal/services"
+	txservices "github.com/stellar/wallet-backend/internal/transactions/services"
```
JakeUrban, line 13: Why is `TransactionService` in `/transactions/services` rather than `/services` like `AccountService`? Maybe worth a refactor in another PR.

internal/serve/graphql/schema/mutations.graphqls
```diff
@@ -25,3 +28,30 @@ type DeregisterAccountPayload {
...
+# Input types for transaction mutations
+input BuildTransactionsInput {
+    transaction: TransactionInput!
+}
+
+input TransactionInput {
+    operations: [String!]!
+    timeout: Int!
+    simulationResult: SimulationResultInput
+}
```
JakeUrban, line 41: I'd like to suggest a pretty impactful change to this mutation:

Right now, this mutation accepts a subset of all attributes of a transaction. See the complete list of attributes [here](https://pkg.go.dev/github.com/stellar/go/xdr#Transaction). This means that we don't support building transactions with memos or ledger time bounds, for example. Its also awkward for clients to create a transaction, simulate it, then pull out the attributes required for the mutation.

Why don't we simply accept a full transaction? This would make it easier for clients, because they wouldn't have to construct a more complex request body, and it would ensure that the wallet backend can build the transaction including all of its attributes.

The wallet backend would take that transaction and rebuild it using our own channel account (and its sequence number).

pkg/wbclient/client.go
```diff
@@ -35,9 +61,91 @@ func NewClient(baseURL string, requestSigner auth.HTTPRequestSigner) *Client {
...
+		if transaction.SimulationResult.LatestLedger != 0 {
+			simulationResult["latestLedger"] = transaction.SimulationResult.LatestLedger
+		}
+
+		if transaction.SimulationResult.Error != "" {
+			simulationResult["error"] = transaction.SimulationResult.Error
+		}
+
+		variables["input"].(map[string]interface{})["transaction"].(map[string]interface{})["simulationResult"] = simulationResult
+	}
```
JakeUrban, line 141: We can remove all of this if we accept transaction XDR in the request rather than a subset of its attributes.

internal/serve/serve.go
```diff
@@ -283,16 +283,7 @@ func handler(deps handlerDeps) http.Handler {
 			r.Post("/create-fee-bump", accountHandler.CreateFeeBumpTransaction)
 		})
 
-		r.Route("/transactions", func(r chi.Router) {
-			handler := &httphandler.TransactionsHandler{
```
JakeUrban, line 287: I don't see the `TransactionHandler` used elsewhere in this PR, can we remove it? Or is it used by another REST endpoint?

internal/serve/graphql/resolvers/mutations.resolvers.go
```diff
@@ -84,6 +91,110 @@ func (r *mutationResolver) DeregisterAccount(ctx context.Context, input graphql1
...
+					Extensions: map[string]interface{}{
+						"code": "INVALID_TRANSACTION_DATA",
+					},
+				}
+			}
+			simulationResult.TransactionData = txData
+		}
+	}
+
+	tx, err := r.transactionService.BuildAndSignTransactionWithChannelAccount(ctx, ops, int64(transaction.Timeout), simulationResult)
```
JakeUrban, line 140: If we support accepting a full transaction in the gql request, we won't have to construct an `entities.RPCSimulatedTransactionResult` type and can just pass the transaction. 

We'll need to update the `BuildAndSignTransactionWithChannelAccount` implementation though. There are some bugs we need to fix in the current implementation though anyway.

internal/serve/graphql/resolvers/mutations.resolvers.go
```diff
@@ -84,6 +91,110 @@ func (r *mutationResolver) DeregisterAccount(ctx context.Context, input graphql1
...
+	if err != nil {
+		switch {
+		case errors.Is(err, transactionservices.ErrInvalidArguments):
+			return nil, &gqlerror.Error{
+				Message: err.Error(),
+				Extensions: map[string]interface{}{
+					"code": "INVALID_ARGUMENTS",
+				},
+			}
```
JakeUrban, line 149: Can we have a more specific error message/code here? The client won't know which arguments are invalid.

internal/serve/graphql/resolvers/mutations.resolvers.go
```diff
@@ -84,6 +91,110 @@ func (r *mutationResolver) DeregisterAccount(ctx context.Context, input graphql1
...
+		case errors.Is(err, signing.ErrUnavailableChannelAccounts):
+			return nil, &gqlerror.Error{
+				Message: err.Error(),
+				Extensions: map[string]interface{}{
+					"code": "UNAVAILABLE_CHANNEL_ACCOUNTS",
+				},
+			}
```
JakeUrban, line 156: Whats the difference between `UNAVAILABLE_CHANNEL_ACCOUNTS` and `NO_IDLE_CHANNEL_ACCOUNT`?

internal/serve/graphql/resolvers/mutations.resolvers.go
```diff
@@ -84,6 +91,110 @@ func (r *mutationResolver) DeregisterAccount(ctx context.Context, input graphql1
...
+	// Convert transaction to XDR string
+	txXdrStr, err := tx.Base64()
+	if err != nil {
+		return nil, &gqlerror.Error{
+			Message: fmt.Sprintf("Failed to encode transaction to XDR: %s", err.Error()),
+			Extensions: map[string]interface{}{
+				"code": "XDR_ENCODING_FAILED",
+			},
+		}
+	}
```
JakeUrban, line 190: If we validated the transaction provided in the request, and our building logic is without bugs, then this shouldn't fail, so I don't know if we should handle the error here.

internal/serve/graphql/resolvers/mutations.resolvers.go
akcays: They represent different stages of the same problem:
* `NO_IDLE_CHANNEL_ACCOUNT` - Immediate failure: No channel accounts are available right now (all are busy). This triggers the retry mechanism in the signature client.
* `UNAVAILABLE_CHANNEL_ACCOUNTS` - Final failure: After exhausting all retry attempts over time, we still couldn't get any channel accounts.

The flow:
1. First attempt fails with `NO_IDLE_CHANNEL_ACCOUNT`-> sleep and retry
2. After all retries exhausted -> return `UNAVAILABLE_CHANNEL_ACCOUNTS`

However, there's a logical inconsistency in the current code. The `GetAccountPublicKey` method can return `NO_IDLE_CHANNEL_ACCOUNT` in two different contexts:
  1. Inside retry loop (handled by continuing)
  2. After retry loop (returned to caller)

This makes the same error mean different things. We should consider renaming these to be clearer:
  * `CHANNEL_ACCOUNTS_BUSY` (temporary - retry)
  * `CHANNEL_ACCOUNTS_EXHAUSTED_RETRIES` (permanent - after all retries failed)

Additionally, I've spotted a bug where `UNAVAILABLE_CHANNEL_ACCOUNTS` can be returned immediately on database errors ([here](https://github.com/stellar/wallet-backend/blob/graphql-main/internal/signing/channel_account_db_signature_client.go#L90)), even though the error message says "total time spent waiting" when no time was actually spent retrying.

Would you like me to create a follow-up issue to clean up this error handling logic?

JakeUrban: The client is requesting their transaction be built with a channel account provided by the wallet backend. The client doesn't care about our implementation, so the differentiation between these error codes seems unnecessary to me. We should have one error that indicates the wallet backend could not provide a channel account because they were all busy. Its implementation can decide how long to wait or how many retries to make, but we shouldn't expose those details to the client.

pkg/wbclient/client.go
```diff
@@ -34,10 +60,81 @@ func NewClient(baseURL string, requestSigner auth.HTTPRequestSigner) *Client {
 	}
 }
 
-func (c *Client) BuildTransactions(ctx context.Context, transactions ...types.Transaction) (*types.BuildTransactionsResponse, error) {
-	buildTxRequest := types.BuildTransactionsRequest{Transactions: transactions}
+func (c *Client) BuildTransaction(ctx context.Context, transaction types.Transaction) (*types.BuildTransactionResponse, error) {
+	query := `
```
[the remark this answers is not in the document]
akcays: Addressed here https://github.com/stellar/wallet-backend/pull/260/commits/d6b8e79daf3ae25ef321341ddc3e9f3f8eb97419

internal/serve/graphql/schema/mutations.graphqls
```diff
@@ -25,3 +28,30 @@ type DeregisterAccountPayload {
     success: Boolean!
     message: String
 }
+
+# Input types for transaction mutations
+input BuildTransactionsInput {
+    transaction: TransactionInput!
+}
```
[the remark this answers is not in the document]
akcays: Addressed here https://github.com/stellar/wallet-backend/pull/260/commits/4bfb3f6dae6f2683acf20c6d3728650d27843696

internal/serve/graphql/resolvers/resolver.go
akcays: Addressed here https://github.com/stellar/wallet-backend/pull/260/commits/a51636591a883e18b025daf9a15d5d72a4c70411

pkg/wbclient/client.go
akcays: Created an issue for this to be addressed when we change the input type here https://github.com/stellar/wallet-backend/issues/278

pkg/wbclient/client.go
```diff
@@ -34,10 +68,64 @@ func NewClient(baseURL string, requestSigner auth.HTTPRequestSigner) *Client {
...
+
+	if transaction.SimulationResult.LatestLedger != 0 {
+		simulationResult["latestLedger"] = transaction.SimulationResult.LatestLedger
+	}
+
+	if transaction.SimulationResult.Error != "" {
+		simulationResult["error"] = transaction.SimulationResult.Error
+	}
+
+	variables["input"].(map[string]interface{})["transaction"].(map[string]interface{})["simulationResult"] = simulationResult
```
JakeUrban, line 121: yikes, is there a more idiomatic way to do this? @aditya1702 what do you think?

JakeUrban (approved):
I'll defer to Aditya on the details but this LGTM!

PR #261, opened 2025-08-01
Build dynamic sql query based on columns requested in GraphQL

aditya1702:
### What

- Read the passed graphql fields from the request context
- Use the passed fields to construct sql queries dynamically to only fetch those columns
- Refactor the dataloaders and resolvers to use this new functionality
- Update the unit tests

### Why

This helps optimize requests to only fetch the required columns that are requested by the user.

### Known limitations

N/A

#262 

#### PR Structure

#### Thoroughness

#### Release


internal/indexer/types/types.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 type Account struct {
-	StellarAddress string    `json:"stellarAddress,omitempty" db:"stellar_address"`
+	StellarAddress string    `json:"address,omitempty" db:"stellar_address"`
```
JakeUrban, line 14: Any reason not to change this to `address` within the struct rather than just the json representation?

JakeUrban (commented):
Awesome work.

internal/indexer/types/types.go
aditya1702: @JakeUrban We are reading the json tag and mapping it to db tags so that is what needs to be changed. However, I do want to change the struct variable to Address too but thats outside the scope of this PR. Will do it in a separate one

PR #268, opened 2025-08-08
Use a composite key as primary key for state changes

aditya1702:
### What

Use a composite key tuple of transaction/operation order (toid) and the state change order within an operation as the primary key for a state change.

- Adds a new db column:`toID` which represents the operationID if not a fee state change otherwise its a transactionID. This is because fee state changes are not associated with an operation but a tx. The [toID](https://pkg.go.dev/github.com/stellar/go/toid) is an ID that represents the ledger, a transaction's order within that ledger and an operation's order within that tx
- Adds a new column: `StateChangeOrder` which represents the order of a state change within an operation.
- Establishes a deterministic order of state changes within an operation based on a sort key: `toID > category > reason > account ID > .....`.
- Add the primary key for state changes table:  `(to_id, state_change_order)`
- Update unit tests to account for the change

### Why

We need a way to uniquely identify a state change and this new composite key of toID and the state change's order within an operation helps us do this. The sort order makes sure that the order value we calculate will always be deterministic and produce the same result if we do reprocessing/backfilling.

The `state_change_order` is also helpful for pagination queries for state changes where we can use the tuple (`to_id`, `state_change_order`) as a cursor.

### Known limitations

N/A

#269 

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/db/migrations/2025-06-10.4-create_indexer_table_state_changes.sql
+++ b/internal/db/migrations/2025-06-10.4-create_indexer_table_state_changes.sql
@@ -2,7 +2,8 @@
 
 -- Table: state_changes
 CREATE TABLE state_changes (
-    id TEXT PRIMARY KEY,
+    to_id BIGINT NOT NULL,
+    state_change_order BIGINT NOT NULL,
     state_change_category TEXT NOT NULL,
     state_change_reason TEXT,
     ingested_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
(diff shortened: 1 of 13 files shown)


internal/db/migrations/2025-06-10.4-create_indexer_table_state_changes.sql
```diff
@@ -23,7 +24,8 @@ CREATE TABLE state_changes (
     spender_account_id TEXT,
     sponsored_account_id TEXT,
     sponsor_account_id TEXT,
-    thresholds JSONB
+    thresholds JSONB,
+    PRIMARY KEY (to_id, state_change_order)
```
akcays, line 28: 👍

Issue #271, opened 2025-08-08
Paginate through an account's transactions and operations


aditya1702:
Support following type of queries:

```
query {
   account {
         transactions(first: 2) { .... }
         operations(first: 2) { .... }
   }
}
```

PR #279, opened 2025-08-11
Paginate through an account's transactions and operations

aditya1702:
### What

Add ability to paginate through an account's transactions and operations. The design follows the standard specifications set forth in this doc - https://relay.dev/graphql/connections.htm - which is a standardization defined for GraphQL APIs. Using this, the following type of queries are now possible for wallet-backend's API:

```
query {
	account(address: "GBKFMWTIBN5S7CT5QC2OYIAALZYKCTZKZ2MNG4ODXR6BAMOEWH2NSK43") {
		transactions(first: 2, after: "MzYzMTM1MTg5OTExMTQzMQ==") {
			edges {
				node {
					hash
				}
			}
			pageInfo {
				endCursor
				hasNextPage
			}
		}
		transactions(last: 2, before: "MzYzMTM1MTg5OTExMTQzMQ==") {
			edges {
				node {
					hash
				}
			}
			pageInfo {
				endCursor
				hasNextPage
			}
		}	
		operations(last: 2, before: "MzYzMTM1MTg5OTExMTQzMQ==") {
			edges {
				node {
					id
					operationType
				}
			}
			pageInfo {
				endCursor
				hasNextPage
			}
		}
		operations(first: 2, after: "MzYzMTM1MTg5OTExMTQzMQ==") {
			edges {
				node {
					id
					operationType
				}
			}
			pageInfo {
				endCursor
				hasNextPage
			}
		}	
}
```

Some important notes on the implementation:

#### Cursor Column
We use the `to_id` column for transactions and the `id` column for operations as the cursor. Since the ID is formed from the ledger number and the transaction/op order, it automatically encodes temporal information and is a good value for the cursor.

#### Response Ordering
The results are **always** returned in descending order of the cursor i.e. the recent ledger transactions/ops at the top. This follows the guideline stated in the relay doc that the returning order should stay consistent irrespective of using `first/after` or `last/before`.

For e.g. lets assume we have 10 txns in our db: `tx1, tx2, tx3,...., tx10` 
So given our default ordering is `latest txns/ops -> earlier txns/ops`, when a user passes `first: 2, after: tx10`, we return the first 2 transactions after `tx10` in the decreasing order i.e. `tx9, tx8` and one can continue paging through txns in the descending order of ledgers using `first/after`.

When `last: 2, before: tx1` is passed, we return the last 2 txns that come after `tx1` when ordered in descending order i.e. `tx2, tx3` and keep on paging backwards using `last/before`.

Basically, the forwards/backwards direction is based on the canonical ordering we decide which is descending order of ledgers.

### Why

We want to support paging through txns and operations for an account.

#271

### DIFF
--- a/internal/data/query_utils.go
+++ b/internal/data/query_utils.go
@@ -57,7 +57,7 @@ func buildGetByAccountAddressQuery(config paginatedQueryConfig) (string, []any) 
 	// Add cursor condition if provided
 	if config.Cursor != nil {
 		// When paginating in descending order, we are going from greater cursor id to smaller cursor id
-		if config.OrderBy == "DESC" {
+		if config.OrderBy == DESC {
 			queryBuilder.WriteString(fmt.Sprintf(` AND %s.%s < $%d`, config.TableName, config.CursorColumn, argIndex))
 		} else {
 			queryBuilder.WriteString(fmt.Sprintf(` AND %s.%s > $%d`, config.TableName, config.CursorColumn, argIndex))
@@ -67,7 +67,7 @@ func buildGetByAccountAddressQuery(config paginatedQueryConfig) (string, []any) 
 	}
 
 	// Add ordering
-	if config.OrderBy == "DESC" {
+	if config.OrderBy == DESC {
 		queryBuilder.WriteString(fmt.Sprintf(" ORDER BY %s.%s DESC", config.TableName, config.CursorColumn))
 	} else {
 		queryBuilder.WriteString(fmt.Sprintf(" ORDER BY %s.%s ASC", config.TableName, config.CursorColumn))
@@ -83,7 +83,7 @@ func buildGetByAccountAddressQuery(config paginatedQueryConfig) (string, []any) 
 
 	// For backward pagination, wrap query to reverse the final order
 	// This ensures we always display the oldest items first in the output
-	if config.OrderBy == "DESC" {
+	if config.OrderBy == DESC {
 		query = fmt.Sprintf(`SELECT * FROM (%s) AS %s ORDER BY %s.cursor ASC`,
 			query, config.TableName, config.TableName)
 	}


aditya1702:
> > We use the to_id column for transactions
> 
> This isn't ideal from a usability perspective, clients are more likely to want to use the transaction's hash don't you think? Why can't we allow clients to specify the hash as the `before`/`after` value while still using the transaction's associated `to_id` field for ordering?

@JakeUrban That is not how these pagination parameters are supposed to work in a typical graphql request. You request an account's transactions: 

```
account(address: "GBKFMWTIBN5S7CT5QC2OYIAALZYKCTZKZ2MNG4ODXR6BAMOEWH2NSK43") {
		transactions(first: 10) {
			edges {
				node {
					hash
				}
			}
			pageInfo {
				endCursor
				hasNextPage
			}
		}
}
```

Then use the `endCursor` from the responses to keep paging through the transactions with the `after` argument. The specification defines that the cursors be encoded and all business logic of what these cursors are be hidden. So my PR is just strictly adding the minimum requirements for a simple relay based pagination.

What you are describing is supporting requests where clients would want to page from arbitrary txns based on the hash/time which we would add as a separate feature. WDYT?

internal/serve/graphql/resolvers/utils.go
```diff
@@ -50,3 +167,39 @@ func getColumnMap(model any) map[string]string {
...
+		cursor = after
+		limit = *first
+		isForward = true
+	} else if last != nil {
+		cursor = before
+		limit = *last
+		isForward = false
+	} else {
+		limit = defaultLimit
+	}
```
JakeUrban, line 193: If I omit `first` and `last`, is `isForward` `false` because its not explicitly defined?

internal/data/query_utils.go
```diff
@@ -35,3 +36,92 @@ func PrepareNamedQuery(ctx context.Context, connectionPool db.ConnectionPool, na
...
+		argIndex++
+	}
+
+	// Add ordering
+	// Forward: newer to older (DESC), Backward: older to newer (ASC)
+	if config.Forward {
+		queryBuilder.WriteString(fmt.Sprintf(" ORDER BY %s.%s DESC", config.TableName, config.CursorColumn))
+	} else {
+		queryBuilder.WriteString(fmt.Sprintf(" ORDER BY %s.%s ASC", config.TableName, config.CursorColumn))
+	}
```
JakeUrban, line 99: Why not use `Order`? `Forward` seems to me like a new concept you have to wrap your head around.

internal/serve/graphql/dataloaders/transaction_loaders.go
```diff
@@ -17,31 +17,6 @@ type TransactionColumnsKey struct {
 	Columns       string
 }
 
-// TransactionsByAccountLoader creates a dataloader for fetching transactions by account address
-// This prevents N+1 queries when multiple accounts request their transactions
-// The loader batches multiple account addresses into a single database query
-func TransactionsByAccountLoader(models *data.Models) *dataloadgen.Loader[TransactionColumnsKey, []*types.Transaction] {
```
JakeUrban, line 23: Why are these removed? Does this mean we're susceptible to the N+1 problem for transactions-for-account queries? Same for the `OperationsByAccountLoader`.

internal/serve/graphql/resolvers/utils.go
aditya1702: @JakeUrban Yes it would be false. However, now that I think about it, it should be true so by default its returning from decreasing -> increasing transaction IDs 🤔

internal/data/query_utils.go
aditya1702: Changed it to `isDescending` in https://github.com/stellar/wallet-backend/pull/279/commits/bbbb2ddd3afca517268e786b01ac05098109513f

internal/serve/graphql/resolvers/utils.go
aditya1702: Also changed the default order to descending

internal/data/query_utils.go
```diff
@@ -81,7 +81,7 @@ func buildGetByAccountAddressQuery(config paginatedQueryConfig) (string, []any)
 
 	// Add cursor condition if provided
 	if config.Cursor != nil {
-		if config.Forward {
+		if config.IsDescending {
 			queryBuilder.WriteString(fmt.Sprintf(` AND %s.%s < $%d`, config.TableName, config.CursorColumn, argIndex))
```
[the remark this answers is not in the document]
JakeUrban: If I understand correctly, what we need is:

- `first`/`last` can't be used together
- `after`/`before` can't be used together
- `first/before` can't be used together
- `after`/`last` can't be used together
- If `after` is specified, `cursor value > cursor provided`
- if `before` is specified`, `cursor value < cursor provided`
- if `first` is specified, `ORDER BY ASC`
- if `last` is specified, `ORDER BY DESC`
- if neither `first` or `last` is specified, `ORDER BY DESC`

PR #284, opened 2025-08-13
Remove /payments/ endpoint, models, and ingestion logic

akcays:
### What

### Why

### Known limitations

[TODO or N/A]

#### PR Structure

#### Thoroughness

#### Release


internal/data/query_utils.go
```diff
@@ -1,37 +1,8 @@
...
-	"fmt"
-
-	"github.com/jmoiron/sqlx"
-
-	"github.com/stellar/wallet-backend/internal/db"
-)
-
 type SortOrder string
 
 const (
```
aditya1702, line 5: are these constants used somewhere? It makes more sense to move them elsewhere where they are used (assuming its just a single file?) instead of keeping them in a separate file

internal/db/migrations/2024-05-22.0-ingest_payments.sql
```diff
@@ -31,4 +31,4 @@ CREATE INDEX to_address_idx ON ingest_payments (to_address);
 
 DROP TABLE ingest_payments;
 
-DROP TABLE ingest_store;
+DROP TABLE ingest_store;
```
aditya1702, line 34: Add a newline

internal/db/migrations/2025-08-13.0-drop_ingest_payments_table.sql
```diff
@@ -0,0 +1,26 @@
+-- +migrate Up
```
aditya1702, line 1: Normally, when removing prior tables, we would introduce a drop migration but in this case, since we have not officially released WB, I was thinking of just deleting the migration files for unused tables. This way we release wallet backend with a proper migration with only the tables that are used. What are your thoughts? @JakeUrban @akcays

internal/serve/httphandler/request_params_validator.go
```diff
@@ -22,15 +22,6 @@ func DecodeJSONAndValidate(ctx context.Context, req *http.Request, reqBody inter
 	return ValidateRequestParams(ctx, reqBody, appTracker)
 }
 
-func DecodeQueryAndValidate(ctx context.Context, req *http.Request, reqQuery interface{}, appTracker apptracker.AppTracker) *httperror.ErrorResponse {
```
aditya1702, line 25: Ohh weird, was this request validation only used for payments? 🤔

openapi/main.yaml
```diff
@@ -440,4 +276,4 @@ paths:
               example:
                 status: 500
                 error: An error occurred while processing this request.
-x-original-swagger-version: '2.0'
+x-original-swagger-version: '2.0'
```
aditya1702, line 279: Add newline

internal/db/migrations/2025-08-13.0-drop_ingest_payments_table.sql
JakeUrban: Yea I'm ok with that approach. We should reach out to known users of the wallet backend prior to merging to main though so they have the opportunity to fork the repo if they want to keep their own applications stable.

internal/serve/httphandler/request_params_validator.go
akcays: Yup- it was only being used [here](https://github.com/stellar/wallet-backend/pull/284/files#diff-79bec5c3f3d3b400ed3052934528a98f6aa9e823636998ec89345d0faf6a1686L37)

internal/data/query_utils.go
akcays: Actually they aren't even being used anymore- I've just fully removed the file

PR #286, opened 2025-08-13
Paginate through an account's state changes

aditya1702:
### What

Add ability to paginate through an account's state changes. The overall design is similar to paginating an account's operations and transactions in https://github.com/stellar/wallet-backend/pull/279. 

The difference is in the cursor design for a state change since it is a combination of a txn/op's ID and the state change's order within an operation. This leads to a secondary layer of pagination where we paginate through ledgers using the `to_id` but also within an operation's state changes, using the `state_change_order`.

```
query {
	account(address: "GBKFMWTIBN5S7CT5QC2OYIAALZYKCTZKZ2MNG4ODXR6BAMOEWH2NSK43") {
		stateChanges(first: 2, after: "MzYzMTM1MTg5OTExMTQzMQ==") {
			edges {
				node {
					hash
				}
			}
			pageInfo {
				endCursor
				hasNextPage
			}
		}
		stateChanges(last: 2, before: "MzYzMTM1MTg5OTExMTQzMQ==") {
			edges {
				node {
					hash
				}
			}
			pageInfo {
				endCursor
				hasNextPage
			}
		}
}
```

#### Cursor Column
We use combination of `to_id, state_change_order` as the cursor for a state change, since that tuple uniquely identifies a state change in our table.

#### Response Ordering
The results are ordered in the descending order of `to_id` and then within the same operation, its ordered in descending order of state changes `state_change_order`

### Why

We want to support paging through an account's state changes.

#282


internal/data/statechanges.go
```diff
@@ -17,26 +18,56 @@ type StateChangeModel struct {
...
+
+	if isForward {
+		queryBuilder.WriteString(" ORDER BY to_id DESC, state_change_order DESC")
+	} else {
+		queryBuilder.WriteString(" ORDER BY to_id ASC, state_change_order ASC")
+	}
+
+	if limit != nil && *limit > 0 {
+		queryBuilder.WriteString(fmt.Sprintf(" LIMIT %d", *limit))
+	}
```
JakeUrban, line 56: Why are we checking if limit is non-zero? Shouldn't this validation have been done prior?

PR #295, opened 2025-08-19
Bugfix: race condition caused by processing taking too long

marcelosalloum:
### What

Only call the ingestor's manualTriggerChannel if it's empty

### Why

When the processing time takes longer than 5s, which is common in production, since even downloading the ledgers from the RPC takes ~5s for 10 ledgers, the code was entering a deadlock.

Considering the processing time is 11s, here's the sequence of events that lead to a deadlock:

- [**T=0s**] RPC Health Check: channel now contains 1 health update
https://github.com/stellar/wallet-backend/blob/8a12000c478e72169413ef2b47085415b5c1fdb7/internal/services/rpc_service.go#L284
- [**T=0s**] Ingestion Loop Starts. Channel will be emptied and the processing step will take 11s
https://github.com/stellar/wallet-backend/blob/8a12000c478e72169413ef2b47085415b5c1fdb7/internal/services/ingest.go#L226
- [**T=5s**] RPC Health Check (BLOCKS!): RPC health goroutine STOPS executing since it gets blocked until `r.heartbeatChannel` is consumed.
https://github.com/stellar/wallet-backend/blob/8a12000c478e72169413ef2b47085415b5c1fdb7/internal/services/rpc_service.go#L284
- [**T=11s**]: Ingestion Loop Finishes (BLOCKS!): the manual trigger is called, but since the RPC health check is already halted, this will make the ingestor halt as well, resulting in a deadlock.

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/services/ingest.go
+++ b/internal/services/ingest.go
@@ -176,8 +176,12 @@ func (m *ingestService) DeprecatedRun(ctx context.Context, startLedger uint32, e
 			m.metricsService.ObserveIngestionDuration(totalIngestionPrometheusLabel, time.Since(start).Seconds())
 
 			// immediately trigger the next ingestion the wallet-backend is behind the RPC's latest ledger
-			if resp.LatestLedger-ingestLedger > 1 && len(manualTriggerChannel) == 0 {
-				manualTriggerChannel <- true
+			if resp.LatestLedger-ingestLedger > 1 {
+				select {
+				case manualTriggerChannel <- true:
+				default:
+					// do nothing if the channel is full
+				}
 			}
 
 			ingestLedger++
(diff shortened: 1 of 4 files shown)


internal/services/ingest.go
```diff
@@ -38,7 +38,7 @@ var ErrAlreadyInSync = errors.New("ingestion is already in sync")
 const (
 	advisoryLockID                          = int(3747555612780983)
 	ingestHealthCheckMaxWaitTime            = 90 * time.Second
-	getLedgersLimit                         = 50 // NOTE: cannot be larger than 200
+	getLedgersLimit                         = 10 // NOTE: cannot be larger than 200. In production set it to 10 if your RPC has MAX_GET_LEDGERS_EXECUTION_DURATION="5s" or unset.
```
aditya1702, line 41: I think you need to merge your changes from this PR to use the flag default value: https://github.com/stellar/wallet-backend/pull/296. Right?

internal/services/ingest.go
```diff
@@ -171,7 +171,7 @@ func (m *ingestService) DeprecatedRun(ctx context.Context, startLedger uint32, e
 			m.metricsService.ObserveIngestionDuration(totalIngestionPrometheusLabel, time.Since(start).Seconds())
 
 			// immediately trigger the next ingestion the wallet-backend is behind the RPC's latest ledger
-			if resp.LatestLedger-ingestLedger > 1 {
+			if resp.LatestLedger-ingestLedger > 1 && len(manualTriggerChannel) == 0 {
```
aditya1702, line 174: I am a little confused so can you revisit why the `manualTriggerChannel` is needed? Wouldnt the `rpcHeartbeatChannel` keep the loop running and we would catchup to the tip if we fall behind?

aditya1702 (approved):
LGTM. Good idea using non blocking channel sends. Can you also add a test to make sure this is working as expected ✌️

PR #309, opened 2025-09-04
services: complete the channel account service so it can create/delete any number of accounts

marcelosalloum:
### What

Complete the channel account service, so it can create/delete any number of accounts.

### Why

Before this change, it wouldn't be able to delete any accounts and would only be able to create at most 19 accounts. Such behavior is not enough for a prod-ready application.

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/services/channel_account_service.go
+++ b/internal/services/channel_account_service.go
@@ -2,13 +2,14 @@ package services
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	"math"
 	"time"
 
 	"github.com/stellar/go/keypair"
 	"github.com/stellar/go/support/log"
 	"github.com/stellar/go/txnbuild"
+	"github.com/stellar/go/xdr"
 
 	"github.com/stellar/wallet-backend/internal/db"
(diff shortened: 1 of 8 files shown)


internal/services/channel_account_service.go
```diff
@@ -48,14 +50,44 @@ func (s *channelAccountService) EnsureChannelAccounts(ctx context.Context, numbe
...
+		return s.deleteChannelAccountsInBatches(ctx, int(numAccountsToDelete))
+	} else {
+		// Create additional accounts
+		numOfChannelAccountsToCreate := number - currentChannelAccountNumber
+		log.Ctx(ctx).Infof("⏳ Creating %d additional channel accounts...", numOfChannelAccountsToCreate)
+		return s.createChannelAccountsInBatches(ctx, int(numOfChannelAccountsToCreate))
+	}
+}
+
+// createChannelAccountsInBatches creates channel accounts in batches of MaximumCreateAccountOperationsPerStellarTx (19).
```
aristidesstaffieri, line 71: looks like this comment is stale

PR #317, opened 2025-09-18
Prevent ingestion from stopping for state change processing errors

aditya1702:
### What

Instead of returning error and stopping the state change processing and halting ingestion, we log the error with proper details and continue processing the rest of the events.

### Why

A single event/state change can halt the ingestion completely. We want to keep processing the rest of the state changes and log the errors for debugging.

### Known limitations

NA


aditya1702:
> should we add tests for effects that makes sure that we can continue processing after a failure?

@aristidesstaffieri Hmm I thought we already tested that. But I realized we dont for the effects processor. I will add the relevant tests

aristidesstaffieri (approved):

Issue #278, opened 2025-08-11
Accept full transaction object in `BuildTransaction` mutation instead of subset of attributes


akcays:
Currently, the `BuildTransaction` GraphQL mutation only accepts a subset of transaction attributes (`operations`, `timeout`, and `simulationResult`), this limits building transactions with additional attributes like memos, time bounds, etc. and also makes it more complex for clients.

**Problems:**

  1. Limited functionality: We don't support building transactions with memos, ledger time bounds, or other transaction attributes available [here](https://pkg.go.dev/github.com/stellar/go/xdr#Transaction)
  2. Awkward client workflow: Clients have to create a transaction, simulate it, then extract only the supported attributes for our mutation.
  3. Missing features: Important transaction features are not accessible through the wallet backend.

**Proposed Solution:**

Change the `BuildTransaction` mutation to accept a full transaction object instead of just the current subset of attributes (`operations`, `timeout`, `simulationResult`).

**Context:** This limitation was introduced when working on contract simulation features, but for general transaction building use cases, accepting the full transaction makes more sense.

https://github.com/stellar/wallet-backend/pull/260#discussion_r2254905537
https://github.com/stellar/wallet-backend/pull/260#discussion_r2263945873

PR #322, opened 2025-09-24
Directly pass the transaction XDR in build transaction endpoint

aditya1702:
### What

- Clients now pass the transaction XDR string to the `BuildTransaction` endpoint.
- Refactored the `BuildAndSignTransactionWithChannelAccount` to take in the XDR string and do the validation checks
- Refactored the unit and integration tests with this change
- Simplified the graphql codes to send to the client in case of an error in build transaction endpoint.

### Why

It is much easier to pass the XDR strings instead of building separate operations and passing the array of operations. The only drawback is that users will need to use a dummy account to build the XDR and then wallet backend will rebuild the same txn with its channel account and corresponding sequence.

### Known limitations

N/A

#278


internal/transactions/services/transaction_service.go
```diff
@@ -134,16 +134,53 @@ func (t *transactionService) BuildAndSignTransactionWithChannelAccount(ctx conte
...
+	}
+	if clientPreconditions.MinSequenceNumberAge != 0 {
+		finalPreconditions.MinSequenceNumberAge = clientPreconditions.MinSequenceNumberAge
+	}
+	if clientPreconditions.MinSequenceNumberLedgerGap != 0 {
+		finalPreconditions.MinSequenceNumberLedgerGap = clientPreconditions.MinSequenceNumberLedgerGap
+	}
+	if len(clientPreconditions.ExtraSigners) > 0 {
+		finalPreconditions.ExtraSigners = clientPreconditions.ExtraSigners
+	}
```
marcelosalloum, line 173: Do you need to check for empty values before assigning it? I believe the result is the same if you don't check.

marcelosalloum (approved):

internal/transactions/services/transaction_service_test.go
```diff
@@ -244,7 +244,7 @@ func TestBuildAndSignTransactionWithChannelAccount(t *testing.T) {
 	require.NoError(t, outerErr)
 
 	t.Run("🔴timeout_must_be_smaller_than_max_timeout", func(t *testing.T) {
```
aristidesstaffieri, line 246: should we also expand the suite with some tests for using client preconditions?

aditya1702: I updated tests since we are now expecting to recieve entire XDR string

internal/transactions/services/transaction_service.go
```diff
@@ -134,22 +140,39 @@ func (t *transactionService) BuildAndSignTransactionWithChannelAccount(ctx conte
...
 			AccountID: channelAccountPublicKey,
 			Sequence:  channelAccountSeq,
 		},
-		Operations: operations,
-		BaseFee:    t.BaseFee,
-		Preconditions: txnbuild.Preconditions{
-			TimeBounds: txnbuild.NewTimeout(timeoutInSecs),
-		},
+		Operations:           operations,
+		BaseFee:              t.BaseFee,
```
aristidesstaffieri, line 164: should this optionally come from the `clientTx` too? Is there another way to increase the base fee?

PR #327, opened 2025-10-01
Add GraphQL API documentation to README

aditya1702:
### What

- Make graphql query complexity configurable via env flags
- Some housekeeping changes to graphql enums to remove unused state changes enums.
- Add documentation for wallet backend's graphql API.

### Why

Part of the graphql API design work.

### Known limitations

N/A

https://github.com/stellar/wallet-backend/issues/326


internal/serve/graphql/schema/enums.graphqls
```diff
@@ -43,9 +43,8 @@ enum StateChangeCategory {
   METADATA
   FLAGS
   TRUSTLINE
-  SPONSORSHIP
```
aristidesstaffieri, line 46: this seems unrelated, is this intentional?

aditya1702: @aristidesstaffieri I mentioned this in the PR description. Looks like the graphql enum was not updated to match this change from `SPONSORSHIP` -> `RESERVES` so I changed it in this PR.

PR #369, opened 2025-11-13
Use checkpoint ledger to populate account token cache in redis

### LINKED ISSUE #358: Use checkpoint ledger to populate redis cache with account token list

aditya1702:
### What

- Add account-tokens service that reads ledger entries from latest checkpoint reader and populates the token cache.
- Add a contract validator service that validates whether a contract spec is SEP41 or not.
      - It reads the contract code ledger entries and extracts the wasm hash and code for each entry.
      - With the contract code, we parse the contract spec and validate against know SEP41 spec.
      - The validation method is to validate all function names and respective input/output signatures
- Relevant unit tests
- Update integration test suite:
     - Add redis container
     - Change rpc and core docker images to v24
     - Update wallet ingest and api containers to reference the redis container host and port

### Why

### Known limitations

N/A

Closes #358 #363

### DIFF
--- a/internal/services/contract_validator.go
+++ b/internal/services/contract_validator.go
@@ -46,7 +46,7 @@ var scSpecTypeNames = map[xdr.ScSpecType]string{
 // sep41FunctionSpec defines the expected signature for a SEP-41 token function.
 type sep41FunctionSpec struct {
 	name            string
-	expectedInputs  map[string]any
+	expectedInputs  map[string]string
 	expectedOutputs []string
 }
 
@@ -55,32 +55,32 @@ type sep41FunctionSpec struct {
 var sep41RequiredFunctions = []sep41FunctionSpec{
 	{
 		name:            "balance",
-		expectedInputs:  map[string]any{"id": "Address"},
+		expectedInputs:  map[string]string{"id": "Address"},
 		expectedOutputs: []string{"i128"},
 	},
 	{
 		name:            "allowance",
-		expectedInputs:  map[string]any{"from": "Address", "spender": "Address"},
+		expectedInputs:  map[string]string{"from": "Address", "spender": "Address"},
 		expectedOutputs: []string{"i128"},
 	},
 	{
 		name:            "decimals",
-		expectedInputs:  map[string]any{},
+		expectedInputs:  map[string]string{},
 		expectedOutputs: []string{"u32"},
 	},
 	{
 		name:            "name",
-		expectedInputs:  map[string]any{},
+		expectedInputs:  map[string]string{},
 		expectedOutputs: []string{"String"},
 	},
 	{
 		name:            "symbol",
-		expectedInputs:  map[string]any{},
+		expectedInputs:  map[string]string{},
 		expectedOutputs: []string{"String"},
 	},
 	{
 		name: "approve",
-		expectedInputs: map[string]any{
+		expectedInputs: map[string]string{
 			"from":              "Address",
 			"spender":           "Address",
 			"amount":            "i128",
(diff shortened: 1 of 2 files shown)


internal/services/contract_validator.go
```diff
@@ -0,0 +1,398 @@
...
+func (v *contractValidator) Close(ctx context.Context) error {
+	if err := v.runtime.Close(ctx); err != nil {
+		return fmt.Errorf("closing contract spec validator: %w", err)
+	}
+	return nil
+}
+
+func (v *contractValidator) isContractCodeSEP41(contractSpec []xdr.ScSpecEntry) bool {
+	/*
+		For a contract to be SEP-41, it must atleast have the following functions and inputs/outputs:
```
aristidesstaffieri: several of these signatures have changed since CAP67 to support muxed destinations. We should check for both signatures but likely you will see the muxed address destination going forward.

internal/services/ingest.go
```diff
@@ -271,10 +304,23 @@ func (m *ingestService) Run(ctx context.Context, startLedger uint32, endLedger u
 			return fmt.Errorf("processing ledger response: %w", err)
 		}
 
-		// update cursor
-		startLedger = getLedgersResponse.Ledgers[len(getLedgersResponse.Ledgers)-1].Sequence
-		if err := m.models.IngestStore.UpdateLatestLedgerSynced(ctx, m.ledgerCursorName, startLedger); err != nil {
-			return fmt.Errorf("updating latest synced ledger: %w", err)
+		// update cursors
+		err = db.RunInTransaction(ctx, m.models.DB, nil, func(dbTx db.Transaction) error {
```
aristidesstaffieri: why run this in a transaction but not use the transaction?

internal/services/account_tokens.go
```diff
@@ -0,0 +1,515 @@
...
+	key := s.buildTrustlineKey(accountAddress)
+	trustlines, err := s.redisStore.SMembers(ctx, key)
+	if err != nil {
+		return nil, fmt.Errorf("getting trustlines for account %s: %w", accountAddress, err)
+	}
+	return trustlines, nil
+}
+
+// GetAccountContracts retrieves all Stellar Asset Contract (SAC) balance contract IDs for an account from Redis.
+func (s *accountTokenService) GetAccountContracts(ctx context.Context, accountAddress string) ([]string, error) {
```
piyalbasu: Is the idea here that this will basically be the contract id's for the list of trustlines the account has? Trying to think of the use-case where a wallet would prefer this over the trustlines. If I were to show my classic/sac balances, I would probably just want to use the account trustlines

aditya1702: @piyalbasu My bad, I should have updated the func docstring with latest (Updated in https://github.com/stellar/wallet-backend/pull/369/commits/79ce45245aaea17ed6f719b1fa4bafd99a63de8a) . So `GetAccountContracts` will return the following:
1. G-address: all custom token IDs (non-SAC)
2. C-address: all custom and SAC token IDs

We will still call the `GetAccountTrustlines` for a G-address to fetch the classic assets

internal/services/ingest.go
aditya1702: Ah good catch - added in https://github.com/stellar/wallet-backend/pull/369/commits/591a858526d5493429c8b708f535b16fd69148ca

internal/indexer/types/types.go
```diff
@@ -41,6 +41,15 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
+type ContractType string
+
+const (
+	ContractTypeNative  ContractType = "NATIVE"
```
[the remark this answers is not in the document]
aditya1702: Ah this got added in this PR by mistake, its used in the graphql resolver: https://github.com/stellar/wallet-backend/pull/372/files#diff-4f5944448cab4bbf7619d82033d3a9e69b2b9ee22abfce48632909a2f4d8fb72R376

internal/services/contract_validator.go
```diff
@@ -0,0 +1,324 @@
...
+			"amount":            "i128",
+			"expiration_ledger": "u32",
+		},
+		expectedOutputs: []string{},
+	},
+	{
+		name: "transfer",
+		expectedInputs: map[string]any{
+			"from":   set.NewSet("Address", "MuxedAddress"), // Support the new MuxedAddress type change introduced in CAP-67
+			"to":     "Address",
```
aristidesstaffieri: its not from that can accept a muxed address, it is the `to_muxed` parameter.
https://github.com/stellar/soroban-examples/blob/main/token/src/contract.rs#L120

internal/indexer/types/types.go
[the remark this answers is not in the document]
aristidesstaffieri: looks like this is set up for the next PR, resolving.

internal/services/contract_validator.go
aditya1702: @aristidesstaffieri Ah! I dont know how I confused this with the `from` address - fixed in https://github.com/stellar/wallet-backend/pull/369/commits/de5d9fe2ec421fc80b56241b1a2a95c3b5b6db06

PR #372, opened 2025-11-13
Add graphql resolver for balances query

### LINKED ISSUE #362: Add balances query resolver for the graphql API

aditya1702:
### What

Adds the relevant resolver for the account balances query.

### Why

### Known limitations

N/A

Closes #362

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -184,8 +184,8 @@ func GetContractDataEntryLedgerKey(holderAddress, contractAddress string) (strin
 
 // IsContractAddress determines if the given address is a contract address (C...) or account address (G...)
 func IsContractAddress(address string) bool {
-	// Contract addresses start with 'C' and account addresses start with 'G'
-	return len(address) > 0 && address[0] == 'C'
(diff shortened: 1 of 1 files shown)


internal/utils/utils.go
```diff
@@ -62,6 +62,132 @@ func GetAccountLedgerKey(address string) (string, error) {
...
+		return "", fmt.Errorf("marshalling ledger key: %w", err)
+	}
+
+	return keyXdr, nil
+}
+
+// IsContractAddress determines if the given address is a contract address (C...) or account address (G...)
+func IsContractAddress(address string) bool {
+	// Contract addresses start with 'C' and account addresses start with 'G'
+	return len(address) > 0 && address[0] == 'C'
```
aristidesstaffieri: can improve this by checking if you can `StrKey.decodeContract(address[0])`

aristidesstaffieri: but the helper isn't correct, a string that starts with C isn't guaranteed to be a contract address.
I guess if we plan to only use this in the context of a G or C address then its good enough but this feels incomplete to me generally.

PR #379, opened 2025-11-17
Store details for SAC, SEP41 tokens during checkpoint population

### LINKED ISSUE #343: Define the indexer types in common file and reuse them everywhere

aditya1702:
### What

- When populating tokens from checkpoint, make simulation calls for SEP41 compliant tokens to get the `name`, `symbol` and `decimals`
- For `SAC`, we extract the code and issuer from the name and store those. For non-SAC tokens, we store the data as is.
- Update the contract_tokens table to store all the additional information
- Remove the contracts in-memory cache since we will get the required data from the new contract_tokens table. The ID is the primary key and is indexed to get us the performance
- Add `BatchInsert` method to insert the contract tokens details

### Why

### Known limitations

N/A

Closes #343

### DIFF
--- a/internal/data/contract_tokens.go
+++ b/internal/data/contract_tokens.go
@@ -22,9 +22,9 @@ type Contract struct {
 	Type      string    `db:"type" json:"type"`
 	Code      *string   `db:"code" json:"code"`
 	Issuer    *string   `db:"issuer" json:"issuer"`
-	Name      string    `db:"name" json:"name"`
-	Symbol    string    `db:"symbol" json:"symbol"`
-	Decimals  int16     `db:"decimals" json:"decimals"`
+	Name      *string   `db:"name" json:"name"`
+	Symbol    *string   `db:"symbol" json:"symbol"`
+	Decimals  uint32    `db:"decimals" json:"decimals"`
 	CreatedAt time.Time `db:"created_at" json:"createdAt"`
 	UpdatedAt time.Time `db:"updated_at" json:"updatedAt"`
 }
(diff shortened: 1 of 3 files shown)


internal/data/contract_tokens.go
```diff
@@ -0,0 +1,119 @@
...
+	DB             db.ConnectionPool
+	MetricsService metrics.MetricsService
+}
+
+type Contract struct {
+	ID        string    `db:"id" json:"id"`
+	Type      string    `db:"type" json:"type"`
+	Code      *string   `db:"code" json:"code"`
+	Issuer    *string   `db:"issuer" json:"issuer"`
+	Name      string    `db:"name" json:"name"`
```
aristidesstaffieri: should `name and `symbol` also be nullable? Looks like they are in the db schema.

internal/services/account_tokens.go
```diff
@@ -537,24 +558,289 @@ func (s *accountTokenService) extractHolderAddress(key xdr.ScVal) (string, error
...
+		}
+
+		contracts = append(contracts, &data.Contract{
+			ID:       contractID,
+			Type:     string(metadata.Type),
+			Code:     code,
+			Issuer:   issuer,
+			Name:     metadata.Name,
+			Symbol:   metadata.Symbol,
+			Decimals: int16(metadata.Decimals),
```
aristidesstaffieri: I guess this can technically overflow since decimals are `uint32` but it wouldn't really make sense to have a decimal value that high, just wanted to point it out though.

internal/services/account_tokens.go
```diff
@@ -537,24 +558,289 @@ func (s *accountTokenService) extractHolderAddress(key xdr.ScVal) (string, error
...
+		for _, contractID := range contractIDsBatch {
+			group.Submit(func() {
+				data, err := s.fetchAllcontractMetadata(ctx, contractID)
+				if err != nil {
+					log.Ctx(ctx).Warnf("Failed to fetch metadata for contract %s: %v", contractID, err)
+					return
+				}
+
+				// Only store if we successfully fetched at least name and symbol
+				if data.Name != "" && data.Symbol != "" {
```
aristidesstaffieri: why not check decimals here? Without the decimals, clients have no way to know how to display a balance.

aditya1702: I changed it to uint32. Fixed in https://github.com/stellar/wallet-backend/pull/379/commits/f7e5a58dbda661c5170323e3f6473136d9c3eeb4

PR #389, opened 2025-11-21
[internal] Uses LedgerBackend interface for ingestion

### LINKED ISSUE #388: Use LedgerBackend interface for ingestion
Using LedgerBackend interface will allow us to ingest from any data source that supports it.

Refactor from http calls to rpc, to using LedgerBackend for rpc.

None

aristidesstaffieri:
Closes #388 

### What
Adds ledgerbackend and uses LedgerBackend interface for ingestion, now supports datastore. No support for direct captive core ledger backend at this point.
Adds `ledger-backend-type` and `datastore-config` options.

### Why
Adds flexibility to ingestion in order to support any source that produces a compatible response(data lake)

### Known limitations
N/A

#388 

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/services/ingest.go
+++ b/internal/services/ingest.go
@@ -6,10 +6,7 @@ import (
 	"fmt"
 	"hash/fnv"
 	"io"
-	"os"
-	"os/signal"
 	"sort"
-	"syscall"
 	"time"
 
 	"github.com/alitto/pond/v2"
@@ -83,37 +80,6 @@ func NewIngestService(
 	networkPassphrase string,
 	archive historyarchive.ArchiveInterface,
 ) (*ingestService, error) {
-	if models == nil {
-		return nil, errors.New("models cannot be nil")
-	}
-	if ledgerCursorName == "" {
-		return nil, errors.New("ledgerCursorName cannot be nil")
-	}
-	if accountTokensCursorName == "" {
-		return nil, errors.New("accountTokensCursorName cannot be nil")
-	}
-	if appTracker == nil {
-		return nil, errors.New("appTracker cannot be nil")
-	}
-	if rpcService == nil {
-		return nil, errors.New("rpcService cannot be nil")
-	}
-	if ledgerBackend == nil {
-		return nil, errors.New("ledgerBackend cannot be nil")
-	}
-	if chAccStore == nil {
-		return nil, errors.New("chAccStore cannot be nil")
-	}
-	if metricsService == nil {
-		return nil, errors.New("metricsService cannot be nil")
-	}
-	if getLedgersLimit <= 0 {
-		return nil, errors.New("getLedgersLimit must be greater than 0")
-	}
-	if archive == nil {
-		return nil, errors.New("archive cannot be nil")
-	}
-
 	// Create worker pool for the ledger indexer (parallel transaction processing within a ledger)
 	ledgerIndexerPool := pond.NewPool(0)
 	metricsService.RegisterPoolMetrics("ledger_indexer", ledgerIndexerPool)
@@ -186,10 +152,6 @@ func (m *ingestService) Run(ctx context.Context, startLedger uint32, endLedger u
 		}
 	}
 
-	signalChan := make(chan os.Signal, 1)
-	signal.Notify(signalChan, syscall.SIGTERM, syscall.SIGINT, syscall.SIGQUIT)
-	defer signal.Stop(signalChan)
-
 	// Prepare backend range
 	err = m.prepareBackendRange(ctx, startLedger, endLedger)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/ingest.go
```diff
@@ -67,6 +68,22 @@ func (c *ingestCmd) Command() *cobra.Command {
...
+		{
+			Name:        "ledger-backend-type",
+			Usage:       "Type of ledger backend to use for fetching ledgers. Options: 'rpc' (default) or 'datastore'",
+			OptType:     types.String,
+			ConfigKey:   &ledgerBackendType,
+			FlagDefault: string(ingest.LedgerBackendTypeRPC),
+			Required:    false,
+		},
+		{
+			Name:        "datastore-config",
```
aditya1702: nit: `datastore-config-path`

internal/ingest/ingest.go
```diff
@@ -83,6 +106,47 @@ func setupDeps(cfg Configs) (services.IngestService, error) {
 	if err != nil {
 		return nil, fmt.Errorf("instantiating rpc service: %w", err)
 	}
+
+	// Initialize ledger backend based on configuration
+	var ledgerBackend ledgerbackend.LedgerBackend
```
aditya1702: lets move this whole section to a separate function for readability

internal/ingest/ingest.go
```diff
@@ -83,6 +106,47 @@ func setupDeps(cfg Configs) (services.IngestService, error) {
...
+	if cfg.LedgerBackendType == LedgerBackendTypeDatastore {
```
aditya1702: lets use a switch-case here

internal/services/ingest.go
```diff
@@ -351,23 +364,59 @@ func (m *ingestService) Run(ctx context.Context, startLedger uint32, endLedger u
...
-	return getLedgersResponse, nil
+	// Prepare the range based on backend type:
+	// - RPCLedgerBackend: Use UnboundedRange and prepare only once (continuously buffers)
+	// - BufferedStorageBackend: Use BoundedRange and prepare for each batch
+	_, isRPCBackend := m.ledgerBackend.(*ledgerbackend.RPCLedgerBackend)
+
+	if isRPCBackend {
+		// For RPC backend, prepare unbounded range only once
+		if !m.rpcBackendPrepared {
+			ledgerRange := ledgerbackend.UnboundedRange(ledgerSeqRange.StartLedger)
```
aditya1702: Will this also work if we have prepared the range once, then ingestion pod dies and we restart again?

internal/services/ingest.go
```diff
@@ -517,18 +565,15 @@ func (m *ingestService) mergeLedgerBuffers(ledgerBuffers []*indexer.IndexerBuffe
...
-func (m *ingestService) processLedgerResponse(ctx context.Context, getLedgersResponse GetLedgersResponse) error {
+func (m *ingestService) processLedgerResponse(ctx context.Context, ledgers []xdr.LedgerCloseMeta) error {
 	// Phase 1: Collect all transaction data and participants from all ledgers in parallel
-	start := time.Now()
-	ledgerDataList, err := m.collectLedgerTransactionData(ctx, getLedgersResponse)
+	ledgerDataList, err := m.collectLedgerTransactionData(ctx, ledgers)
 	if err != nil {
 		return fmt.Errorf("collecting ledger transaction data: %w", err)
 	}
-	collectDuration := time.Since(start)
```
aditya1702: Why did we remove this?

internal/services/ingest.go
```diff
@@ -448,6 +492,10 @@ func (m *ingestService) collectLedgerTransactionData(ctx context.Context, getLed
 	if len(errs) > 0 {
 		return nil, fmt.Errorf("collecting ledger data: %w", errors.Join(errs...))
 	}
+	phaseDuration := time.Since(phaseStart).Seconds()
+	m.metricsService.ObserveIngestionPhaseDuration("collect_transaction_data", phaseDuration)
+	log.Ctx(ctx).Infof("🚧 Done collecting data from %d ledgers in %vs", len(ledgers), phaseDuration)
```
aditya1702: I removed this log line and moved the phase duration to the main function. Any reason for moving it back?

internal/services/ingest.go
aristidesstaffieri: afaik this should be done per instance of the ledgerBackend so the in memory flag should work but let me know if I'm missing something.

internal/ingest/ingest.go
aristidesstaffieri: refactored in a179cf3328872b65cd01421cec8d5f18745de2fa

internal/services/ingest.go
```diff
@@ -228,344 +150,217 @@ func (m *ingestService) Run(ctx context.Context, startLedger uint32, endLedger u
...
+			// If start ledger is some value less than latest ingested ledger, we go into backfilling mode. In this mode
+			// we dont update the account token cache (since it is already populated with recent checkpoint ledger) and we dont
+			// update the latest ledger ingested cursor.
+			// NOTE: Currently we dont have the functionality of detecting gaps and intelligently backfilling so we would process the same
+			// ledgers again during backfilling. However the db insertions have ON CONFLICT DO NOTHING, so we would not do repeated insertions
+			m.backfillMode = true
 		}
 	}
 
 	signalChan := make(chan os.Signal, 1)
```
tamirms: I don't think it is appropriate to set up the signals here. You probably want to do this at the very top level when running the application. The context which is passed into Run() will be sufficient as a termination singal. You can use https://pkg.go.dev/os/signal#NotifyContext to create a context which is done once it receives SIGTERM, SIGINT, etc

PR #400, opened 2025-12-04
Allow operators to skip storing `TXMETA`

### LINKED ISSUE #434: Add an env flag for operators to skip storing `TX_META`

aditya1702:
### What

- Add an env flag `SKIP_TX_META` to control storing txmeta in our transactions table.
- Update unit tests
- Update the integration tests
- Update the Go client methods

### Why

TXMETA occupies a lot of storage. We already extract the relevant account activity data in the form of state changes, so operators can simply skip storing these. Having the flag also gives a choice if we want to start storing it for future features.

### Known limitations

N/A

Closes #434

### DIFF
--- a/cmd/ingest.go
+++ b/cmd/ingest.go
@@ -70,7 +70,7 @@ func (c *ingestCmd) Command() *cobra.Command {
 		},
 		{
 			Name:        "ledger-backend-type",
-			Usage:       "Type of ledger backend to use for fetching ledgers. Options: 'rpc' (default) or 'datastore'",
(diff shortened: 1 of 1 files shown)


cmd/ingest.go
```diff
@@ -73,15 +73,23 @@ func (c *ingestCmd) Command() *cobra.Command {
 			Usage:       "Type of ledger backend to use for fetching ledgers. Options: 'rpc' (default) or 'datastore'",
```
aristidesstaffieri: looks like default is now datastore

PR #416, opened 2025-12-10
Add new graphql endpoint for multi-account balances call

### LINKED ISSUE #415: Add balances endpoint to fetch balances for multiple accounts in 1 call

aditya1702:
### What

- Add a new endpoint `balancesByAccountAddresses` which processes multiple accounts and fetches their balances in 1 call.
- Add a config variable `MaxAccountsPerBalances` to configure the maximum allowed account addresses for this query.
- Add unit tests
- Add integration tests
- Update README with the new endpoint details
- Update the Go client with this new endpoint.

### Why

This reduces API calls to get balances for multiple accounts

### Known limitations

N/A

Closes #415

### DIFF
--- a/internal/serve/graphql/resolvers/resolver.go
+++ b/internal/serve/graphql/resolvers/resolver.go
@@ -35,6 +35,7 @@ const (
 // ResolverConfig holds configuration values for the GraphQL resolver.
 type ResolverConfig struct {
 	MaxAccountsPerBalancesQuery int
+	MaxWorkerPoolSize           int
 }
 
 var ErrNotStateChange = errors.New("object is not a StateChange")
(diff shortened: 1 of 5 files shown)


internal/serve/graphql/resolvers/queries.resolvers.go
```diff
@@ -201,201 +248,349 @@ func (r *queryResolver) BalancesByAccountAddress(ctx context.Context, address st
...
+					"address": addr,
+				},
 			}
+		}
+	}
 
-			contractID, ok := contractDataEntry.Contract.GetContractId()
-			if !ok {
-				continue
+	// Deduplicate addresses while preserving order
```
aristidesstaffieri: NIT: deduplication can happen before validation in order to only validate each member once.

internal/serve/graphql/resolvers/resolver.go
```diff
@@ -64,6 +74,8 @@ func NewResolver(models *data.Models, accountService services.AccountService, tr
 		rpcService:          rpcService,
 		accountTokenService: accountTokenService,
 		metricsService:      metricsService,
+		pool:                pond.NewPool(0),
```
aristidesstaffieri: Even if this seems ok to be unbounded given our `max-accounts-per-balances-query`, we should still provide a config variable to limit the workers for deployments that have different resource constraints or rate limits, can be default `0`.

internal/serve/graphql/resolvers/queries.resolvers.go
aditya1702: Ah good catch!

PR #418, opened 2025-12-11
Fix balances bug

aditya1702:
### What

- Update docker file to install dlv so we can attach debugger to our kube pods
- Fix the bug where we incorrectly map ledger keys to the ledger entry results.


piyalbasu:
@aditya1702 Looks like the `resolvers` tests are failing now. We should also make sure we have a test for the scenario we saw (some empty balances being misattributed to the wrong address)

piyalbasu (approved):

PR #419, opened 2025-12-15
Optimize account token cache design

### LINKED ISSUE #422: Optimize account token cache structure design to cut down storage usage

aditya1702:
### What

Optimizes Redis storage for the Account Token Cache to significantly reduce memory usage through multiple compression strategies.
1. PostgreSQL Asset ID Mapping - Store compact integer IDs in Redis instead of full CODE:ISSUER strings. Asset definitions are stored once in a new trustline_assets PostgreSQL table.
2. Bucket Sharding - Distribute accounts across 12,000 Redis hash buckets using [FNV-1a hashing](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function) instead of one key per account. This reduces key overhead and improves Redis memory efficiency.
3. VARINT Binary Encoding - Encode asset IDs using [variable-length integer format](https://en.wikipedia.org/wiki/Variable-length_quantity) instead of ASCII  
4. Frequency-Based ID Assignment - Sort assets by frequency during initial population so common assets (USDC, EURC) get lowest IDs (1, 2, 3...), which encode to fewer bytes in VARINT format.

With all these optimizations the memory usage went from **3 GB -> 387 MB**.

### Why

Optimizing memory usage for our token cache is important since the account keys will have an unbounded growth and this could lead to increased storage.

### Known limitations

N/A

Closes #422

### DIFF
--- a/internal/data/trustline_assets_test.go
+++ b/internal/data/trustline_assets_test.go
@@ -226,8 +226,8 @@ func TestTrustlineAssetModel_BatchGetOrInsert(t *testing.T) {
 	t.Run("handles mix of existing and new assets", func(t *testing.T) {
 		cleanUpDB()
 		mockMetricsService := metrics.NewMockMetricsService()
-		mockMetricsService.On("ObserveDBQueryDuration", "BatchGetOrCreateIDs", "trustline_assets", mock.Anything).Return()
-		mockMetricsService.On("IncDBQuery", "BatchGetOrCreateIDs", "trustline_assets").Return()
+		mockMetricsService.On("ObserveDBQueryDuration", "BatchGetOrInsert", "trustline_assets", mock.Anything).Return()
+		mockMetricsService.On("IncDBQuery", "BatchGetOrInsert", "trustline_assets").Return()
 		defer mockMetricsService.AssertExpectations(t)
 
 		m := &TrustlineAssetModel{


internal/services/account_tokens.go
```diff
@@ -489,32 +601,74 @@ func (s *accountTokenService) enrichContractTypes(
...
+// which results in smaller varint encoding and reduced Redis memory usage.
+// Contract addresses are stored directly as full strings.
 func (s *accountTokenService) storeAccountTokensInRedis(
 	ctx context.Context,
-	trustlines map[string][]string,
-	contracts map[string][]string,
+	trustlinesByAccountAddress map[string]set.Set[wbdata.TrustlineAsset],
+	contractsByAccountAddress map[string]set.Set[string],
+	trustlineFrequency map[wbdata.TrustlineAsset]int,
+	_ []string, // uniqueContractTokens - no longer needed, contracts stored directly
```
aristidesstaffieri: any reason to not remove this from the signature?

docker-compose.yaml
```diff
@@ -55,6 +55,10 @@ services:
       - 6379:6379
     volumes:
       - redis-data:/data
+    command: >
+      redis-server
+      --hash-max-listpack-entries 1000
```
aristidesstaffieri: would operators want to tune these as well, since they can tune the bucket size?

internal/services/account_tokens.go
aditya1702: removed it

PR #425, opened 2025-12-18
Implement live + parallel worker backfill ingestion using datastore ledger backend interface

### LINKED ISSUE #427: Implement parallel backfilling functionality using datastore ledger backend interface

aditya1702:
### What

- Add functionality to implement backfilling using datastore ledger backend interface. We now have two modes for ingestion: `LIVE`, `BACKFILL`. Once we do the migrations using live instance, we can have multiple backfill instances that process historical ledger ranges.
- Add ability to calculate gaps in our ledger range - this helps make sure we always process empty ranges and dont insert over already processed ledgers
- Add relevant unit tests
- Update integration tests to spin up parallel backfill ingest container along with the live one and assert succesfull ingestion of transactions and operations.
       - So our setup is this: we spin up RPC, core, fund accounts, create trustlines etc... and then we spin up a live WB ingest container and a backfill ingest container and then assert everything.

### Why

Part of the https://github.com/stellar/wallet-backend/milestone/3. This also allows us to do TPS testing, test different PoCs since we can retroactively backfill old history over different ledger ranges.

### Known limitations

N/A

Closes #427

### DIFF
--- a/internal/services/ingest.go
+++ b/internal/services/ingest.go
@@ -29,12 +29,7 @@ import (
 	"github.com/stellar/wallet-backend/internal/utils"
 )
 
-var ErrAlreadyInSync = errors.New("ingestion is already in sync")
-
 const (
-	// HistoricalBufferLedgers is the number of ledgers to keep before latestRPCLedger
-	// to avoid racing with live finalization during parallel processing.
-	HistoricalBufferLedgers uint32 = 5
 	// maxLedgerFetchRetries is the maximum number of retry attempts when fetching a ledger fails.
 	maxLedgerFetchRetries = 10
 	// maxRetryBackoff is the maximum backoff duration between retry attempts.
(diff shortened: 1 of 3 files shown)


internal/data/ingest_store.go
```diff
@@ -54,3 +59,40 @@ func (m *IngestStoreModel) Update(ctx context.Context, dbTx db.Transaction, curs
 
 	return nil
 }
+
+func (m *IngestStoreModel) UpdateMin(ctx context.Context, dbTx db.Transaction, cursorName string, ledger uint32) error {
+	const query = `
+		UPDATE ingest_store
```
aristidesstaffieri: can this be an upsert to guard against a missing cursor?

internal/services/ingest.go
```diff
@@ -31,6 +30,86 @@ import (
 
 var ErrAlreadyInSync = errors.New("ingestion is already in sync")
 
+const (
+	// BackfillBatchSize is the number of ledgers processed per parallel batch during backfill.
+	BackfillBatchSize uint32 = 250
```
aristidesstaffieri: this seems like a setting that could be configurable for different deployments, wdyt?

internal/data/ingest_store.go
aditya1702: We always initialize this cursor during live ingestion so there will never be the scenario where it is not present

PR #497, opened 2026-02-05
`operations` table: Store the `operation_xdr` as BYTEA

### LINKED ISSUE #492: `operations_accounts`: Store `account_id` as BYTEA

aditya1702:
### What

Updates the operations.operation_xdr storage to use PostgreSQL BYTEA (raw XDR bytes) while preserving the GraphQL API contract by continuing to expose operationXdr as a base64-encoded String.

Changes:

Change operations.operation_xdr from TEXT to BYTEA in the schema/migration and update DB write paths accordingly.
Introduce types.XDRBytea to scan/store raw bytes and encode to base64 when presented as a string.
Update GraphQL schema/resolvers and tests to reflect the new storage/encoding behavior.

### Why

Storage optimization for our DB

### Known limitations

N/A

Closes #492

### DIFF
--- a/internal/data/statechanges.go
+++ b/internal/data/statechanges.go
@@ -532,12 +532,16 @@ func (m *StateChangeModel) BatchGetByToID(ctx context.Context, toID int64, colum
 	if cursor != nil {
 		if sortOrder == DESC {
 			queryBuilder.WriteString(fmt.Sprintf(`
-				AND (to_id, operation_id, state_change_order) < (%d, %d, %d)
-			`, cursor.ToID, cursor.OperationID, cursor.StateChangeOrder))
+				AND (to_id, operation_id, state_change_order) < ($%d, $%d, $%d)
+			`, argIndex, argIndex+1, argIndex+2))
+			args = append(args, cursor.ToID, cursor.OperationID, cursor.StateChangeOrder)
+			argIndex += 3
 		} else {
 			queryBuilder.WriteString(fmt.Sprintf(`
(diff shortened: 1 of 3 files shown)


aristidesstaffieri:
### Code Review — Thresholds and SignerWeights resolvers silently map NULL to 0

**Severity: Medium (incorrect data for GraphQL clients)**

The `Thresholds` resolver never checks `.Valid` on `sql.NullInt16` fields:

https://github.com/stellar/wallet-backend/blob/b84442d95d973a902d27fe20589d370c7e3bb366/internal/serve/graphql/resolvers/statechange.resolvers.go#L275-L280

When a threshold is first set (no prior state), `ThresholdOld` is NULL (`Valid=false`, `Int16=0`). Output: `{"old": "0", "new": "5"}` — clients cannot distinguish "was 0" from "did not exist."

The `SignerWeights` resolver has a similar issue — it returns nil only when *both* fields are NULL, but when one is NULL and the other valid, the NULL one renders as `0`:

https://github.com/stellar/wallet-backend/blob/b84442d95d973a902d27fe20589d370c7e3bb366/internal/serve/graphql/resolvers/statechange.resolvers.go#L238-L247

The correct pattern exists in `TrustlineLimit` (same file), which checks `.Valid` per-field and uses `"null"` for absent values:

https://github.com/stellar/wallet-backend/blob/b84442d95d973a902d27fe20589d370c7e3bb366/internal/serve/graphql/resolvers/statechange.resolvers.go#L349-L364

PR #503, opened 2026-02-13
feature-design: data migrations

### LINKED ISSUE #468: SPIKE: Data Migrations strategy
We need to use a spike to produce a strategy document for wallet backend's data migration. 

The goals are to understand how to add support for new contracts and protocols to our "first class" processors in order to elevate the UX around specific standards(sep-50) and deployments(blnd for example).

aristidesstaffieri:
Closes #468 

### What
Adds a design document for the data migrations feature.

### Why
Document the feature for implementation and receive feedback from stake holders.

### Known limitations
N/A

#468 

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/docs/feature-design/data-migrations.md
+++ b/docs/feature-design/data-migrations.md
@@ -102,13 +102,13 @@ CREATE TABLE protocol_contracts (
 );
 ```
 
-### known_wasms
+### protocol_wasms
 
 A cache for all known WASM blobs. This acts as a filter for the classification process
 to reduce the overhead of classifying new contract instances that use the same WASM code.
 
 ```sql
-CREATE TABLE known_wasms (
+CREATE TABLE protocol_wasms (
     wasm_hash TEXT PRIMARY KEY,
     protocol_id TEXT REFERENCES protocols(id),  -- NULL if unknown/unclassified
     created_at TIMESTAMPTZ DEFAULT NOW()
@@ -172,7 +172,7 @@ Adding a new protocol requires four coordinated processes:
 
 | Step | Requires | Produces |
 |------|----------|----------|
-| **1. protocol-setup** | Protocol migration SQL file, protocol implementation in code | Protocol in DB, `known_wasms`, `protocol_contracts`, `classification_status = success`, both cursors initialized |
+| **1. protocol-setup** | Protocol migration SQL file, protocol implementation in code | Protocol in DB, `protocol_wasms`, `protocol_contracts`, `classification_status = success`, both cursors initialized |
 | **2. ingest (live)** | `classification_status = success`, processor registered | State changes after history convergence (history cursor). Current state after current-state convergence (current-state cursor). |
 | **3a. protocol-migrate history** | `classification_status = success` | Protocol state changes within retention window, through convergence with live ingestion |
 | **3b. protocol-migrate current-state** | `classification_status = success` | Current state from `start_ledger` through convergence with live ingestion |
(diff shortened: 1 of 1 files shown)


docs/feature-design/data-migrations.md
```diff
@@ -0,0 +1,1139 @@
...
+    created_at TIMESTAMPTZ DEFAULT NOW(),
+    updated_at TIMESTAMPTZ DEFAULT NOW()
+);
+
+-- migration_status values:
+-- 'pending'              - Pre-registration, initial state
+-- 'classification_ready' - Checkpoint population is done
+-- 'in_progress'          - Migration in progress
+-- 'ready'                - Migration complete, data is complete
+-- 'failed'               - Migration failed
```
JakeUrban: I'm not sure `pending` is the best name for the initial state. Without understanding the migration process in detail, I can see it getting confused with `in_progress`.

docs/feature-design/data-migrations.md
```diff
@@ -0,0 +1,1139 @@
...
+├── 001_sep50.sql
+├── 002_blend.sql
+└── 003_aqua.sql
+```
+
+These migrations are tracked separately from the main schema migrations, allowing `protocol-setup` to run them independently.
+
+### Implicit Protocol Selection
+
+The command **automatically discovers** which protocols to set up by finding all protocols where `migration_status = 'pending'` in the database.
```
JakeUrban: Supporting a protocol is opt-in though, isn't it? If so, then we'll run discovery for protocols the operator may not have any intention of enabling.

docs/feature-design/data-migrations.md
```diff
@@ -0,0 +1,1139 @@
...
+┌─────────────────────────────────────────────────────────────────────────────┐
+│                        BACKFILL MIGRATION FLOW                              │
+└─────────────────────────────────────────────────────────────────────────────┘
+
+                    ┌────────────────────────────┐
+                    │ ./wallet-backend           │
+                    │ protocol-migrate           │
+                    │ --protocol-id SEP50 ...    │
+                    │ --start-ledger 1           │
+                    │ --end-ledger 5             │
```
JakeUrban: `--start-ledger` needs to be optional right? I'm thinking about when the WB has a sliding history retention, and the protocol being supported launched prior to the earliest ledger in the retention window.

I think right now we don't support a sliding history retention window, right? Either way, it should be optional and default to the earliest ledger in the DB.

docs/feature-design/data-migrations.md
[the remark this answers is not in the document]
JakeUrban: > The protocol-setup command also runs the protocol migrations on initialization, so if you want to opt out of a protocol you can use a protocol migration to set enabled to false at that point, which will cause the discovery logic to skip that one.

I'm not sure I understand, what do you mean by "use a protocol migration"? How does an operator set enabled to false?

docs/feature-design/data-migrations.md
JakeUrban: So in the case where `--start-ledger` is outside the retention window, what happens?

aristidesstaffieri: In your example, if you ran from N to N+1000 then you would have state changes and current state for some ledgers that are before your retention window(N to N+100), but this could be needed in order to produce the current state for a protocol. Current state will often rely on replaying history for the life of the protocol, which could be outside of your history retention window. 

We need to decouple history retention from the migrations processing range, otherwise we would force all operators to have a retention period that covers the entire range of history for all protocols, which would in practice often be all of Soroban's history and will grow over time.

The purpose of the retention window is only to define how much data to retain for history, while the migrations backfill range defines the ledger range needed in order to produce accurate state for the protocol.

aristidesstaffieri: Got it, yeah I agree that we can avoid storing the state changes outside of the current retention window and only need them to process the current state in the batch itself.

So the additional input into the processor should be the start of the retention window at the time of the migration in order to decide to store the state changes or not. If we support a rolling window then the process to slide the window forward will also truncate older data as the window moves forward.

docs/feature-design/data-migrations.md
```diff
@@ -0,0 +1,1139 @@
...
+                                       │
+                  ┌────────────────────┴────────────────────┐
+                  │                                         │
+                  ▼                                         ▼
+┌─────────────────────────────────────┐   ┌─────────────────────────────────────┐
+│         1. CLASSIFICATION           │   │        2. STATE PRODUCTION          │
+│                                     │   │                                     │
+│  Watch for contract deployments/    │   │  Run protocol processors on         │
+│  upgrades in ledger changes         │   │  transactions in ledger             │
+└─────────────────┬───────────────────┘   └─────────────────┬───────────────────┘
```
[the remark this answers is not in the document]
aristidesstaffieri: I've updated this diagram to better reflect the dependency between classification and state production during live ingestion.
a7bf39790390691ccd806f50cfd03daf71e8ae71

aristidesstaffieri:
Here is the milestone created to represent his feature: https://github.com/stellar/wallet-backend/milestone/16

PR #540, opened 2026-03-13
feat(internal): protocol setup

### LINKED ISSUE #506: Classification step #1: Protocol Setup
The first step in a data migration is to run the `protocol-setup` process.
`protocol-setup` is a standalone top level command that uses the `known_wasms` table to compare existing WASM uploads(more specifically their contract interfaces) to a new interface for a protocol defined in the data-migration.

A new command called `protocol-setup`
A module that uses an interface definition(s) for a protocol to compare against known WASMs on the network
Produces updates to `known_wasms` that assign a protocol ID to an entry in `known_wasms` when the matching conditions are met.

N/A

Desing Document for more context: https://github.com/stellar/wallet-backend/pull/503

aristidesstaffieri:
Closes #506 

### What
Add the protocol-setup CLI command that classifies WASM bytecodes against registered protocol validators.

- Introduces a protocol migration system (internal/db/migrations/protocols/) where adding a new protocol only requires a SQL file + validator — no Go constant maintenance
- Implements ProtocolSetupService that fetches unclassified WASM bytecodes via RPC getLedgerEntries (batched at 200/call), extracts contract specs using wazero, validates them against registered ProtocolValidator implementations, and persists matches back to protocol_wasms
- Adds ProtocolValidator interface, WasmSpecExtractor, and a dynamic validator registry for extensibility
- Includes unit tests for all new service/data code and integration tests using real WASM bytecodes against a real DB with mocked RPC

### Why
This is the first step of the protocol data migration pipeline. Before the wallet-backend can index protocol-specific data (e.g., SEP-41 token balances), it needs to know which WASM contracts on the network implement which protocols. The protocol-setup command performs this one-time classification(per protocol) by comparing on-chain WASM bytecodes against protocol interface definitions, tagging each matching WASM in the database. Subsequent pipeline steps (history migration, current-state migration) depend on this classification to know which contracts to index.

### Known limitations
N/A

https://github.com/stellar/wallet-backend/issues/506

#### PR Structure

#### Thoroughness

#### Release


cmd/protocol_setup.go
```diff
@@ -0,0 +1,139 @@
...
+func (c *protocolSetupCmd) Run(databaseURL, rpcURL, networkPassphrase string, protocolIDs []string) error {
+	ctx := context.Background()
+
+	// Build validators from protocol IDs using the dynamic registry
+	var validators []services.ProtocolValidator
+	for _, pid := range protocolIDs {
+		factory, ok := services.GetValidator(pid)
+		if !ok {
+			return fmt.Errorf("unknown protocol ID %q — no validator registered", pid)
+		}
```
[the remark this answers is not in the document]
aristidesstaffieri: this is expected since this task is only for the protocol-setup.

internal/db/migrations/protocols/main.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+
+		content, readErr := fs.ReadFile(protocolFS, entry.Name())
+		if readErr != nil {
+			return executed, fmt.Errorf("reading protocol SQL file %s: %w", entry.Name(), readErr)
+		}
+
+		if _, execErr := sqlDB.ExecContext(ctx, string(content)); execErr != nil {
+			return executed, fmt.Errorf("executing protocol SQL file %s: %w", entry.Name(), execErr)
+		}
```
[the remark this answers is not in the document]
aristidesstaffieri: this is expected and the file will be removed when the first real migration is added.

cmd/protocol_setup.go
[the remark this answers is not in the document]
aristidesstaffieri: added in beac58de44fe91f4b0319781b0eefc4eb21a392a

internal/services/validator_registry.go
```diff
@@ -0,0 +1,15 @@
...
+
+// RegisterValidator registers a validator factory for a protocol ID.
+func RegisterValidator(protocolID string, factory func() ProtocolValidator) {
+	validatorRegistry[protocolID] = factory
+}
+
+// GetValidator returns the validator factory for a protocol ID, if registered.
+func GetValidator(protocolID string) (func() ProtocolValidator, bool) {
+	factory, ok := validatorRegistry[protocolID]
+	return factory, ok
```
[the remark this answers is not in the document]
aristidesstaffieri: added the mutex guard in 5ef78539a2a934a278f263ab15d8103694e626f5

internal/services/contract_validator.go
```diff
@@ -313,6 +313,25 @@ func (v *contractValidator) extractContractSpecFromWasmCode(ctx context.Context,
...
+}
+
+func (v *SEP41ProtocolValidator) ProtocolID() string { return "SEP41" }
+
+func (v *SEP41ProtocolValidator) Validate(specEntries []xdr.ScSpecEntry) bool {
+	return v.cv.isContractCodeSEP41(specEntries)
+}
+
+func (v *SEP41ProtocolValidator) Close(ctx context.Context) error {
+	return v.cv.Close(ctx)
```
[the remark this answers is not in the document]
aristidesstaffieri: e9844d16fd4815ba991acb1f6374c87f4ceb78e3 makes the existing legacy SEP-41 validator stateless

internal/services/protocol_setup.go
```diff
@@ -0,0 +1,277 @@
...
+		}
+	}
+
+	// Log warnings for hashes not found (expired/evicted)
+	for _, hexHash := range hexHashes {
+		if _, found := foundKeys[hexHash]; !found {
+			log.Ctx(ctx).Warnf("WASM hash %s not found in ledger (possibly expired/evicted)", hexHash)
+		}
+	}
+
```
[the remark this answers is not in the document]
aristidesstaffieri: this should not be a frequent occurence

internal/services/protocol_setup.go
```diff
@@ -0,0 +1,277 @@
...
+			log.Ctx(ctx).Warnf("Failed to extract spec from WASM %s: %v", wasmHash, extractErr)
+			continue
+		}
+
+		for _, pid := range protocolIDs {
+			validator := validatorsByProtocol[pid]
+			if validator.Validate(specEntries) {
+				matchedHashes[pid] = append(matchedHashes[pid], types.HashBytea(wasmHash))
+				break // A WASM matches at most one protocol
+			}
```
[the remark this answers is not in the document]
aristidesstaffieri: this is not a concern today, we are looking at roughly ~1000 WASM hashes only.

cmd/protocol_setup.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Run protocol migrations to ensure protocols are registered in the DB
+	if _, err := db.RunProtocolMigrations(ctx, dbPool); err != nil {
+		return fmt.Errorf("running protocol migrations: %w", err)
+	}
+
+	// Create models
+	sqlxDB, err := dbPool.SqlxDB(ctx)
+	if err != nil {
+		return fmt.Errorf("getting sqlx DB: %w", err)
```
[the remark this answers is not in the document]
aristidesstaffieri: this is expected, the actual migrations will be introduced in follow up work.

cmd/protocol_setup.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		validators = append(validators, v)
+	}
+
+	// Open DB connection
+	dbPool, err := db.OpenDBConnectionPool(databaseURL)
+	if err != nil {
+		return fmt.Errorf("opening database connection: %w", err)
+	}
+	defer internalutils.DeferredClose(ctx, dbPool, "closing dbPool in protocol setup")
```
aditya1702, line 96: Why not using defer dbPool.Close() here? FYI I dont know how DeferredClose works

internal/data/protocols.go
```diff
@@ -0,0 +1,123 @@
...
+	StatusSuccess    = "success"
+	StatusFailed     = "failed"
+)
+
+// Protocols represents a row in the protocols table.
+type Protocols struct {
+	ID                          string    `db:"id"`
+	ClassificationStatus        string    `db:"classification_status"`
+	HistoryMigrationStatus      string    `db:"history_migration_status"`
+	CurrentStateMigrationStatus string    `db:"current_state_migration_status"`
```
aditya1702, line 28: What is the difference between `HistoryMigrationStatus` and `CurrentStateMigrationStatus`. If we are doing migration, wouldnt it be on historical data? I dont remember the nuance between the two now

internal/db/migrations/2026-03-09.2-protocol_wasms_fk.sql
```diff
@@ -0,0 +1,6 @@
+-- +migrate Up
+ALTER TABLE protocol_wasms
```
aditya1702: Until the WB is not in production, I would suggest directly changing the migration without adding alter migration files.

internal/db/migrations/protocols/000_placeholder.sql
```diff
@@ -0,0 +1,2 @@
+-- Placeholder file required by go:embed *.sql directive.
```
aditya1702, line 1: A little confused here: What do these individual protocol migration files track here? Because we have the separate `protocols.sql` too right?

internal/services/protocol_validator.go
```diff
@@ -0,0 +1,85 @@
+package services
```
aditya1702, line 1: The two files `contract_validator.go` and `protocol_validator.go` have **duplicated WASM extraction logic** and overlapping validation concerns from parallel evolution — the ingestion pipeline built its own validator, then protocol-setup introduced a more extensible design without refactoring the original.

  ### What each file provides

  | | `contract_validator.go` (older) | `protocol_validator.go` (newer) |
  |---|---|---|
  | **WASM extraction** | `contractValidator.extractContractSpecFromWasmCode()` — private method | `WasmSpecExtractor.ExtractSpec()` — public interface |
  | **SEP-41 validation** | `isContractCodeSEP41()` + `ContractValidator.ValidateFromContractCode()` | `SEP41ProtocolValidator.Validate()` (delegates to
  same `isContractCodeSEP41`) |
  | **wazero runtime** | Owns one instance | Owns a **separate** instance |
  | **Design** | Tightly coupled (extract + validate in one call) | Separated concerns, supports multiple protocols via registry |

  ### Who uses what

  - **`contract_validator.go`** → ingestion pipeline (`ingest.go` → `CheckpointService` → `processContractCode`)
  - **`protocol_validator.go`** → protocol-setup command (`cmd/protocol_setup.go` → `protocolSetupService`)

  ### The duplication

  The WASM extraction logic is **copy-pasted** almost verbatim:
  - `contractValidator.extractContractSpecFromWasmCode()`
  - `wasmSpecExtractor.ExtractSpec()`

  Both compile the WASM module with wazero, find the `contractspecv0` custom section, and unmarshal XDR `ScSpecEntry` entries. Additionally, `SEP41ProtocolValidator.Validate()` just delegates to the same `isContractCodeSEP41()` function that `ContractValidator` uses internally.

  ### Suggested consolidation

  **Delete `contract_validator.go` entirely.** Migrate to the newer, extensible design with one file per protocol:

  internal/services/
    protocol_validator.go          # WasmSpecExtractor + ProtocolValidator interface (keep as-is)
    validator_registry.go          # Registry glue (keep as-is)
    sep41_validator.go             # SEP-41 spec, validation logic, type maps (new — extracted from contract_validator.go)
    sep41_validator_test.go        # SEP-41 tests (new — extracted from contract_validator_test.go)

  Specific changes:
  1. **Create `sep41_validator.go`** — move `SEP41ProtocolValidator`, `isContractCodeSEP41()`, `sep41RequiredFunctions`, `sep41FunctionSpec`,
  `scSpecTypeNames`, `getTypeName()`, and `validateFunctionInputsAndOutputs()` here
  2. **Delete `contract_validator.go`** — the `ContractValidator` interface and `contractValidator` struct are no longer needed
  3. **Update `CheckpointService`** — replace `ContractValidator` dependency with `WasmSpecExtractor` + `[]ProtocolValidator`:
     ```go
     // Before:
     contractValidator.ValidateFromContractCode(ctx, wasmBytes)

     // After:
     specs, err := specExtractor.ExtractSpec(ctx, wasmBytes)
     // then run validators against specs
  4. Update ingest.go — wire WasmSpecExtractor and validators into CheckpointService instead of ContractValidator

  This pattern scales cleanly — adding a new protocol (e.g., SEP-50) means adding one file + one registry call, no existing files modified.

internal/services/validator_registry.go
```diff
@@ -0,0 +1,32 @@
...
+import "sync"
+
+// registryMu guards concurrent access to validatorRegistry.
+var registryMu sync.RWMutex
+
+// validatorRegistry holds factory functions keyed by protocol ID.
+var validatorRegistry = map[string]func() ProtocolValidator{}
+
+// RegisterValidator registers a validator factory for a protocol ID.
+func RegisterValidator(protocolID string, factory func() ProtocolValidator) {
```
aditya1702: Where do we call `RegisterValidator` from? I do not see where a new validator is registered

internal/services/validator_registry.go
```diff
@@ -0,0 +1,32 @@
+package services
+
+import "sync"
+
+// registryMu guards concurrent access to validatorRegistry.
+var registryMu sync.RWMutex
+
+// validatorRegistry holds factory functions keyed by protocol ID.
+var validatorRegistry = map[string]func() ProtocolValidator{}
```
aditya1702: ## Registry Simplification: Store Instances, Not Factories

  ### Current code

  `validator_registry.go` stores factory functions:

  - `map[string]func() ProtocolValidator`
  - `RegisterValidator(protocolID string, factory func() ProtocolValidator)`
  - `GetValidator(protocolID string) (func() ProtocolValidator, bool)`

  ### Problem

  The factory pattern adds indirection that isn't justified. Protocol validators are **stateless** — `SEP41ProtocolValidator` is an empty struct whose `Validate()` is a pure function over spec entries. There's no per-use state, no expensive construction, no resource lifecycle that would warrant lazy instantiation via factories.

  ### Suggested change

  Store validator instances directly where `ProtocolValidator` is an interface and each new validator implements it:

  - `map[string]ProtocolValidator`
  - `RegisterValidator(protocolID string, v ProtocolValidator)`
  - `GetValidator(protocolID string) (ProtocolValidator, bool)`

  Self-registration in `sep41_validator.go` becomes:

```go  
func init() {
      RegisterValidator("SEP41", NewSEP41ProtocolValidator())
  }
```

  Removes an unnecessary layer of indirection. If a future validator ever needs per-use instantiation, it can be refactored then.

cmd/protocol_setup.go
aristidesstaffieri: DeferredClose is just a thin wrapper around Close() that logs the error instead of silently discarding it. It's a common Go pattern for deferred cleanup of `io.Closer` resources.

DeferredClose ensures that if closing fails, the error gets logged — which is helpful for diagnosing connection leak issues or unclean shutdowns.

internal/data/protocols.go
aristidesstaffieri: no, data migrations involve migrating both history and current state. In order to produce current state, you need an individual ledger range to build up the state through the protocol's history, and migrating history requires a different ledger range, since you only need to produce history for your retention period or desired range. 

This distinction is detailed in deeper detail in [the design document](https://github.com/stellar/wallet-backend/blob/feature/data-migrations/docs/feature-design/data-migrations.md) if you need a reminder of the architecture.

internal/db/migrations/protocols/000_placeholder.sql
aristidesstaffieri: each new migration will need new entries into the protocols table, as well as potentially new tables to track state. `protocols.sql` only adds the protocols table itself which is why it is in the regular migrations for schema. Migrations for a specific protocol are added at the time that protocol is implemented and they are run during the `protocol-setup` process.

internal/db/migrations/2026-03-09.2-protocol_wasms_fk.sql
aristidesstaffieri: sounds good, updated in 998bf7f6caee6df4749ea5a46f69cd4869bc91ee

internal/services/protocol_validator.go
aristidesstaffieri: I extracted the extractor and validator flows as described here in 6fc45354afdc1ccecf4aa60df605e0390ed9fc42

Then as we discussed I also removed all of the SEP-41 related code and flows. I left the validator in code so that we can use it when we implement the actual data migration for it. b2ac7ce334c4f672d4009e610380b160a10ae54d

internal/services/validator_registry.go
aristidesstaffieri: The intended pattern is a Go init() function in the protocol's package.

So when a new protocols is implemented you would do - 
```
func init() {                                                   
      RegisterValidator("SEP41", NewSEP41ProtocolValidator())
  }
```

That will run at import time, since `cmd/protocol_setup.go` already imports `internal/services`, any `init()` in that package would execute before the CLI command runs, populating the registry.

internal/services/validator_registry.go
aristidesstaffieri: This change is already applied in 6fc45354afdc1ccecf4aa60df605e0390ed9fc42 and referenced in a previous comment, and the registration is already setup to work with an `init` call.

internal/db/migrations/protocols/000_placeholder.sql
aditya1702: @aristidesstaffieri makes sense - one question: will all new protocol specfic migrations added here be related to current state?

aristidesstaffieri: no not just current state but also adding the protocol itself to the protocols table, but yeah I expect those are the 2 types of sql that will live here.

internal/services/protocol_validator.go
```diff
@@ -0,0 +1,87 @@
...
+	"fmt"
+
+	"github.com/stellar/go-stellar-sdk/support/log"
+	"github.com/stellar/go-stellar-sdk/xdr"
+	"github.com/tetratelabs/wazero"
+)
+
+const contractSpecV0SectionName = "contractspecv0"
+
+// ProtocolValidator is the contract between protocol-setup and individual protocol validators.
```
aditya1702, line 15: "ProtocolValidator is the contract"? Is this a typo?

internal/services/sep41_validator.go
```diff
@@ -189,7 +145,7 @@ func (v *contractValidator) Close(ctx context.Context) error {
 //   - transfer_from: (spender: Address, from: Address, to: Address, amount: i128) -> ()
 //   - burn: (from: Address, amount: i128) -> ()
 //   - burn_from: (spender: Address, from: Address, amount: i128) -> ()
-func (v *contractValidator) isContractCodeSEP41(contractSpec []xdr.ScSpecEntry) bool {
+func isContractCodeSEP41(contractSpec []xdr.ScSpecEntry) bool {
```
aditya1702: nit: we can just move this implemention to Validate() function we implemented. No need to keep `isContractCodeSEP41` since this is explicitly a SEP41 validator now

aditya1702 (approved):
Just 2 nit comments but overall LGTM

internal/services/protocol_validator.go
aristidesstaffieri: Seems correct to me. ProtocolValidator is the interface that protocolSetupService depends on (the "contract"), and individual validators like SEP41ProtocolValidator implement it. The protocol-setup service holds a []ProtocolValidator slice and calls Validate() on each one to classify WASM contracts by protocol.

internal/services/sep41_validator.go
aristidesstaffieri: yeah good call, I updated this in https://github.com/stellar/wallet-backend/pull/540/changes/ddf483d4439eab5551926e9d76e62210edc6e373

Issue #514, opened 2026-02-26
State Production: Live Ingestion State Production with Per-Protocol Dual CAS Gating


aristidesstaffieri:
## Summary

Add protocol state production to the live ingestion pipeline and implement per-protocol dual CAS gating in `PersistLedgerData` to coordinate with migration processes.

## Scope

- Add state production step to live ingestion (after classification, before persist) — run all registered protocol processors on each ledger's transactions
- Implement dual CAS gating within `PersistLedgerData` (`internal/services/ingest_live.go`), for each registered protocol at ledger N:
  - **State changes gating**: Read `protocol_{ID}_history_cursor`. If cursor >= N-1, CAS from N-1→N. On success: write state changes. On CAS fail: skip (migration wrote them). If cursor < N-1: skip (migration hasn't caught up).
  - **Current state gating**: Read `protocol_{ID}_current_state_cursor`. Same CAS logic, writes current state instead.
- Per-protocol independence — different protocols can be at different migration stages simultaneously
- Reason for gating state changes: `BatchCopy` (pgx COPY) doesn't support ON CONFLICT, so duplicates must be prevented

**Key files**: `internal/services/ingest.go`, `internal/services/ingest_live.go`

## Dependencies

- #512 (CompareAndSwap Method on IngestStoreModel)
- #513 (Protocol State Processor Interface)

Issue #513, opened 2026-02-26
State Production: Protocol State Processor Registry


aristidesstaffieri:
## Summary

Define the interface that protocol-specific processors implement to produce state changes and current state from ledger data, plus a registration mechanism for looking up processors by protocol ID.

## Scope

- Interface definition: accepts ledger transaction data + protocol's classified contracts, produces protocol-specific state changes and/or current state
- Processor registry: maps protocol ID → processor implementation (so live ingestion and migrations can look up which processor to run for each protocol)
- The interface must support producing both state changes (operation enrichment) and current state (accumulated state like balances) — processors may produce one or both
- No concrete protocol implementation yet (that comes with the first protocol, e.g., SEP-41 state producer) — just the interface and registry

## Why Separate

This is the contract that live ingestion (state production gating) and both migration commands depend on. Defining it first ensures a consistent interface across all consumers.

## Dependencies

None (pure interface definition)

## Reference

`docs/feature-design/data-migrations.md` (State Production section, line 619+)

PR #542, opened 2026-03-18
feat(internal, cmd): live ingest state production

### LINKED ISSUE #512: State Production: CompareAndSwap Method on IngestStoreModel
## Summary

Add a CAS (Compare-And-Swap) method to `IngestStoreModel` for atomic cursor advancement. This is the foundational primitive for all convergence logic between migrations and live ingestion.

## Scope

- New method on `IngestStoreModel` (`internal/data/ingest_store.go`): `CompareAndSwap(ctx, dbTx, cursorName, expectedValue, newValue) (bool, error)`
- Executes: `UPDATE ingest_store SET value = $new WHERE key = $cursor_name AND value = $expected`, returns `true` if `rows_affected = 1`
- Must work within a `pgx.Tx` (used inside `PersistLedgerData`'s DB transaction and migration batch commits)
- Unit tests for success (value matches), failure (value changed), and key-not-found cases

## Why Separate

Every subsequent state production task depends on CAS. Small, well-defined, independently testable.

## Dependencies

None (existing `ingest_store` table and model)

## Reference

`docs/feature-design/data-migrations.md` (State Production section, line 619+)

aristidesstaffieri:
Closes #512 #513 #514 

### What
- Add a ProtocolProcessor interface and processor registry that allows protocol-specific state producers to be plugged into live ingestion                                 
- Implement dual compare-and-swap (CAS) cursor gating so live ingestion only persists protocol history/current-state when the corresponding migration cursor is caught up, preventing duplicate writes during migration backfill                                                                                                                      
- Add a `protocolContractCache` that periodically refreshes classified contracts from the DB (every 100 ledgers), avoiding per-ledger queries
- Add `CompareAndSwap` to `IngestStoreModel` for atomic cursor advancement within the persist transaction                                                                      
- Add `GetClassified` and `GetByProtocolID` queries to the `protocols` and `protocol_contracts` data models                                                                        
- Add Prometheus metrics for protocol state processing duration, contract cache hit/miss, and cache refresh latency                                                        
- Consolidate integration test coverage into a `DataMigrationTestSuite` that validates the full protocol-setup-to-live-ingestion pipeline, including three CAS gating scenarios: cursors ready (processes), cursors missing (skips), and cursors lagging (skips) 
- Unit tests: `CompareAndSwap`, processor registry, ingest service protocol state production, CAS gating logic                                                               
- Integration tests: `TestProtocolSetupClassifiesIngestedWasms`, `TestLiveIngestionProcessesSetupClassifiedSEP41WhenCursorsReady`, `TestLiveIngestionSkipsSetupClassifiedSEP41WithoutProtocolCursors`, `TestLiveIngestionSkipsSetupClassifiedSEP41WhenProtocolCursorsLag`                                         
- Verify no regressions in existing live ingestion flow (protocol processors are nil by default)
### Known limitations
Since the migrate history/current-state work is not implemented yet, nothing sets the initial cursors right now. Live ingestion in this case will follow the CAS fail scenario until that work is implemented. Tests step around this by manually managing the cursors right now.
No protocols are implemented yet, so the processor registry remains empty.

https://github.com/stellar/wallet-backend/issues/512
https://github.com/stellar/wallet-backend/issues/513
https://github.com/stellar/wallet-backend/issues/514

#### PR Structure

#### Thoroughness

#### Release


internal/services/processor_registry_test.go
```diff
@@ -0,0 +1,95 @@
...
+		for i := range n {
+			id := fmt.Sprintf("PROTO_%d", i)
+			go func() {
+				defer wg.Done()
+				RegisterProcessor(id, func() ProtocolProcessor { return nil })
+			}()
+			go func() {
+				defer wg.Done()
+				GetProcessor(id)
+			}()
```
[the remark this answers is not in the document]
aristidesstaffieri: The project uses Go 1.24, and for i := range n (integer range) has been valid since Go 1.22

internal/services/ingest.go
```diff
@@ -137,6 +142,19 @@ func NewIngestService(cfg IngestServiceConfig) (*ingestService, error) {
 	backfillPool := pond.NewPool(backfillWorkers)
 	cfg.MetricsService.RegisterPoolMetrics("backfill", backfillPool)
 
+	// Build protocol processor map from slice
+	ppMap := make(map[string]ProtocolProcessor, len(cfg.ProtocolProcessors))
+	for _, p := range cfg.ProtocolProcessors {
+		ppMap[p.ProtocolID()] = p
```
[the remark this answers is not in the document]
aristidesstaffieri: hardened in 3040b1adbd15a7fbd796bd799ccf7d276b9eebc2

internal/services/ingest_live.go
```diff
@@ -94,6 +144,60 @@ func (m *ingestService) PersistLedgerData(ctx context.Context, ledgerSeq uint32,
...
+					}
+					if swapped {
+						start := time.Now()
+						persistErr := processor.PersistCurrentState(ctx, dbTx)
+						m.metricsService.ObserveProtocolStateProcessingDuration(protocolID, "persist_current_state", time.Since(start).Seconds())
+						if persistErr != nil {
+							return fmt.Errorf("persisting current state for %s at ledger %d: %w", protocolID, ledgerSeq, persistErr)
+						}
+					}
+				}
```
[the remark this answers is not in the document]
aristidesstaffieri: removed redundant Gets in e3a249c6a840c7f18f889e827890d112f4f6f52d

internal/services/ingest_live.go
```diff
@@ -234,6 +349,85 @@ func (m *ingestService) ingestLiveLedgers(ctx context.Context, startLedger uint3
...
+		if err != nil {
+			log.Ctx(ctx).Warnf("Error refreshing protocol contract cache for %s: %v", protocolID, err)
+			continue
+		}
+		newMap[protocolID] = contracts
+	}
+	m.protocolContractCache.contractsByProtocol = newMap
+	m.protocolContractCache.lastRefreshLedger = currentLedger
+	m.metricsService.ObserveProtocolContractCacheRefreshDuration(time.Since(start).Seconds())
+	log.Ctx(ctx).Infof("Refreshed protocol contract cache at ledger %d", currentLedger)
```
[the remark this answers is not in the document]
aristidesstaffieri: this is fixed in 12250367e9ee430f0385152fa9b3253b7c27c741

internal/services/ingest_test.go
```diff
@@ -2909,3 +2910,363 @@ func Test_ingestService_processBackfillBatchesParallel_BothModes(t *testing.T) {
...
+	t.Parallel()
+
+	ctx := context.Background()
+	mockMetrics := metrics.NewMockMetricsService()
+	mockMetrics.On("ObserveProtocolStateProcessingDuration", "selected", "process_ledger", mock.Anything).Return().Once()
+	mockMetrics.On("IncProtocolContractCacheAccess", "selected", "hit").Return().Once()
+
+	selectedProcessor := NewProtocolProcessorMock(t)
+	skippedProcessor := NewProtocolProcessorMock(t)
+	expectedContracts := []data.ProtocolContracts{{ContractID: types.HashBytea(txHash1), WasmHash: types.HashBytea(txHash2)}}
```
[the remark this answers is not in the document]
aristidesstaffieri: renamed and removed the skippedProcessor in 251590827e19be1e083e20af63183169faf26b90

internal/services/ingest_live.go
```diff
@@ -233,6 +338,104 @@ func (m *ingestService) ingestLiveLedgers(ctx context.Context, startLedger uint3
 	}
 }
 
+// produceProtocolState runs all registered protocol processors against a ledger.
+func (m *ingestService) produceProtocolState(ctx context.Context, ledgerMeta xdr.LedgerCloseMeta, ledgerSeq uint32) error {
```
aditya1702: This function is only used in tests, we should move it to the test file

internal/services/ingest_live.go
```diff
@@ -233,6 +338,104 @@ func (m *ingestService) ingestLiveLedgers(ctx context.Context, startLedger uint3
...
+	// Snapshot previous entries for fallback on partial failure
+	prevContracts := m.protocolContractCache.contractsByProtocol
+	m.protocolContractCache.mu.Unlock()
+
+	// 2. Fetch new data outside the lock
+	start := time.Now()
+	newMap := make(map[string][]data.ProtocolContracts, len(m.protocolProcessors))
+	allSucceeded := true
+	for protocolID := range m.protocolProcessors {
+		contracts, err := m.models.ProtocolContracts.GetByProtocolID(ctx, protocolID)
```
aditya1702: Although the number of protocols is small, we can optimize by using a `BatchGetByProtocolIDs` function here to get everything in 1 call

internal/services/ingest_live.go
```diff
@@ -233,6 +338,104 @@ func (m *ingestService) ingestLiveLedgers(ctx context.Context, startLedger uint3
...
+	m.protocolContractCache.mu.RLock()
+	defer m.protocolContractCache.mu.RUnlock()
+	return m.protocolContractCache.contractsByProtocol[protocolID]
+}
+
+// refreshProtocolContractCache reloads all protocol contracts from the DB.
+// The write lock is held only to check staleness and swap the new data in,
+// keeping DB queries outside the lock to avoid blocking concurrent readers.
+func (m *ingestService) refreshProtocolContractCache(ctx context.Context, currentLedger uint32) {
+	// 1. Check staleness under write lock, copy previous data for fallback
```
aditya1702: Isnt this function called because the cache is detected stale in `getProtocolContracts`? Why are we checking staleness again here?

internal/services/ingest_live.go
```diff
@@ -233,6 +338,104 @@ func (m *ingestService) ingestLiveLedgers(ctx context.Context, startLedger uint3
...
+	}
+	return nil
+}
+
+// getProtocolContracts returns cached contracts for a protocol, refreshing if stale.
+func (m *ingestService) getProtocolContracts(ctx context.Context, protocolID string, currentLedger uint32) []data.ProtocolContracts {
+	if m.protocolContractCache == nil {
+		return nil
+	}
+	m.protocolContractCache.mu.RLock()
```
aditya1702: Why do we need write locks here if live ingestion is single-threaded and this cache will only be called by 1 goroutine?

internal/services/ingest_live.go
```diff
@@ -21,8 +24,55 @@ const (
 	maxIngestProcessedDataRetries      = 5
 	maxIngestProcessedDataRetryBackoff = 10 * time.Second
 	oldestLedgerSyncInterval           = 100
+	protocolContractRefreshInterval    = 100
```
aditya1702, line 26: should we make this a configurable env variable?

internal/services/ingest_live.go
aristidesstaffieri: moved to test file in https://github.com/stellar/wallet-backend/pull/542/changes/56adea9e8c49ec897c6acbeaa17eb55183356cf8

internal/services/ingest_live.go
aristidesstaffieri: added the batch call in c4182e22da147686cfc5a8757012bc01dc3a65f4

internal/services/ingest_live.go
aristidesstaffieri: 1. getProtocolContracts (line 369-370) checks staleness under a read lock and calls refreshProtocolContractCache if stale.                                                
2. refreshProtocolContractCache (line 394-395) re-checks staleness under a write lock before doing the actual refresh.                                                    
                                                                                                                                                                          
Between step 1 (read lock released at line 371) and step 2 (write lock acquired at line 393), another goroutine could have already acquired the write lock and completed the refresh. Without the re-check, you'd get redundant DB queries — multiple goroutines that all saw "stale" under the read lock would each do a full refresh sequentially.

But yeah I see that this isn't ever going to be concurrent so this is not really needed.
I simplified this in 69fc578575495b9a52d3fb89e6098fbab7f934dd

internal/services/ingest_live.go
aristidesstaffieri: it was added under the same pretense as the comment above, but we don't actually need this. Removed in 69fc578575495b9a52d3fb89e6098fbab7f934dd

internal/services/ingest_live.go
aristidesstaffieri: do you think we will actually need this to be configurable? I don't see any operational reason why this would need to be tuned differently for different deployments. `oldestLedgerSyncInterval` is also hardcoded.

aditya1702 (approved):
LGTM

PR #546, opened 2026-03-20
feature(internal, cmd): protocol-migrate history

### LINKED ISSUE #515: State Production: protocol-migrate history Command
## Summary

Implement the history migration subcommand that backfills protocol state changes within the retention window, converging with live ingestion via the shared history CAS cursor.

## Scope

- CLI: `./wallet-backend protocol-migrate history --protocol-id SEP50 SEP41`
- **Phase 1 — Validate**: Verify protocol exists, `classification_status = 'success'`. Set `history_migration_status = 'in_progress'`. Read `oldest_ledger_cursor`. Initialize `protocol_{ID}_history_cursor = oldest_ledger_cursor - 1`.
- **Phase 2 — Process batches to tip** (loop): Read `latest_ledger_cursor` as target. Split `[cursor+1, target]` into batches. Process batches sequentially (parallel comes in Task 6). Each batch commit: CAS-advance history cursor + write state changes. On CAS fail → handoff detected, exit loop. After all batches: re-read latest cursor, repeat if more ledgers; if caught up, block on RPC for next ledger (~5s).
- **Phase 3 — Complete**: Verify cursor at or past attempted ledger. Set `history_migration_status = 'success'`. Clean up.
- Error handling: set `history_migration_status = 'failed'`, log error. On retry: resume from cursor + 1.
- All processed ledgers are within retention — no "process but discard" logic needed.

## Dependencies

- #512 (CompareAndSwap Method on IngestStoreModel)
- #513 (Protocol State Processor Interface)

## Reference

`docs/feature-design/data-migrations.md` (State Production section, line 619+)

aristidesstaffieri:
Closes #515

### What
Implements the `protocol-migrate history` CLI subcommand, backfills protocol history for historical ledgers within the retention window, synchronizing with live ingestion through a shared per-protocol history CAS cursor.                                                           
   
- ProtocolMigrateHistoryService — walks forward from the oldest ingest cursor to the latest cursor, calling `PersistHistory` at each ledger with batch processing and CAS-based progress tracking. On CAS failure, detects handoff from live ingestion and exits cleanly. Tracks `history_migration_status` lifecycle (in_progress → success/failed).
- `protocol-data-migrate` CLI command (cmd/protocol_data_migrate.go) — accepts -`-protocol-id` flags to select which protocols to backfill.
- Shared helpers extracted to ingest_helpers.go — `getLedgerWithRetry`, `buildProtocolProcessorMap`, and cursor name formatters deduplicate logic between history migration and live ingestion.
- `UpdateHistoryMigrationStatus` added to `ProtocolsModel` for migration lifecycle tracking.                                                                                   
- Configurable cursor names — history migration respects `--latest-ledger-cursor-name` / `--oldest-ledger-cursor-name` overrides, matching live ingestion behavior.            
- Convergence poll — distinguishes poll-deadline-exceeded (converged) from transient RPC errors (retry) to avoid premature success marking during network blips.       
- Bulk StatusFailed — excludes protocols already handed off to live ingestion via CAS failure, preventing re-processing conflicts.
- Integration tests covering history-to-live handoff, asymmetric cursor CAS paths (history ready but current_state lags), and BoundedRange→UnboundedRange transitions on shared ledger backends.

### Why
Protocol state production needs historical data within the retention window before live ingestion can provide complete coverage. Without this command, protocols only have history state from the point live ingestion started using their processors — all prior ledgers are missing. This subcommand lets operators backfill per-protocol, with safe handoff to live
ingestion via CAS cursors ensuring no gaps or double-processing at the migration boundary.

### Known limitations
N/A

https://github.com/stellar/wallet-backend/issues/515

#### PR Structure

#### Thoroughness

#### Release


internal/services/protocol_migrate_history.go
```diff
@@ -0,0 +1,449 @@
...
+
+// validate checks that all protocol IDs are valid and ready for history migration.
+// Returns the list of protocol IDs that need processing (excludes already-success ones).
+func (s *protocolMigrateHistoryService) validate(ctx context.Context, protocolIDs []string) ([]string, error) {
+	// Check each protocol has a registered processor
+	for _, pid := range protocolIDs {
+		if _, ok := s.processors[pid]; !ok {
+			return nil, fmt.Errorf("no processor registered for protocol %q", pid)
+		}
+	}
```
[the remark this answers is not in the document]
aristidesstaffieri: added order preserving de-dupe in 07b0e6faf3db7a5245876ce680f4c0f0bba4b215

internal/services/protocol_migrate_history.go
```diff
@@ -0,0 +1,449 @@
...
+				minCursor = t.cursorValue
+				first = false
+			}
+		}
+
+		startLedger := minCursor + 1
+		if startLedger > latestLedger {
+			log.Ctx(ctx).Infof("All protocols at or past tip %d, migration complete", latestLedger)
+			return handedOffProtocolIDs(trackers), nil
+		}
```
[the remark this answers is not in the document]
aristidesstaffieri: The two code paths serve different purposes and the immediate exit is correct:                                                                                             
                                                                                                                                                                             
1. startLedger > latestLedger (immediate exit) — All protocol cursors have already advanced past the latest ingested ledger. There's nothing to process. The cursors got ahead via live ingestion's CAS updates, so migration is done.                                                                                                  
2. Convergence poll (post-processing) — Migration has processed all ledgers in [startLedger, latestLedger] and caught up exactly to the tip. The poll gives RPC 5 seconds to produce the next ledger to confirm whether live ingestion has truly taken over or there's more to process.                                                              
                                                                                                                                                                           
These are semantically different situations:                                                                                                                               
- Case 1: Cursors are already ahead of the tip — migration is provably complete, polling would be pointless (waiting for a ledger the cursors already passed).             
- Case 2: Cursors just reached the tip during this run — a brief poll distinguishes "caught up and done" from "tip is still advancing, keep going."

internal/services/protocol_migrate_history.go
```diff
@@ -0,0 +1,449 @@
...
+					swapped, casErr = s.ingestStore.CompareAndSwap(ctx, dbTx, t.cursorName, expected, next)
+					if casErr != nil {
+						return fmt.Errorf("CAS history cursor for %s: %w", t.protocolID, casErr)
+					}
+					if swapped {
+						return t.processor.PersistHistory(ctx, dbTx)
+					}
+					return nil
+				}); err != nil {
+					return handedOffProtocolIDs(trackers), fmt.Errorf("persisting ledger %d for protocol %s: %w", seq, t.protocolID, err)
```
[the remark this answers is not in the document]
aristidesstaffieri: batch optimizations are for ledger sharing across protocols, PrepareRange prefetch, and COPY FROM bulk inserts.

The per ledger transaction is deliberate and necessary for the CAS handoff to work properly.

cmd/protocol_migrate.go
```diff
@@ -0,0 +1,162 @@
...
+		if !ok {
+			return fmt.Errorf("unknown protocol ID %q — no processor registered", pid)
+		}
+		p := factory()
+		if p == nil {
+			return fmt.Errorf("processor factory for protocol %q returned nil", pid)
+		}
+		processors = append(processors, p)
+	}
+
```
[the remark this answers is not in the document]
aristidesstaffieri: this is expected, no protocols will be implemented yet.

cmd/protocol_migrate.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+type protocolMigrateCmd struct{}
+
+func (c *protocolMigrateCmd) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "protocol-migrate",
+		Short: "Data migration commands for protocol state",
+		Long:  "Parent command for protocol data migrations. Use subcommands to run specific migration tasks.",
+		Run: func(cmd *cobra.Command, args []string) {
```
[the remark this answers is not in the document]
aristidesstaffieri: 80a2d83500c83ca3c426d612e7c7a84e91cb1dbf

internal/data/protocol_contracts.go
```diff
@@ -122,45 +121,3 @@ func (m *ProtocolContractsModel) GetByProtocolID(ctx context.Context, protocolID
 	m.MetricsService.IncDBQuery("GetByProtocolID", "protocol_contracts")
 	return contracts, nil
 }
-
-// BatchGetByProtocolIDs returns all contracts for the given protocol IDs in a single query,
-// grouped by protocol ID.
-func (m *ProtocolContractsModel) BatchGetByProtocolIDs(ctx context.Context, protocolIDs []string) (map[string][]ProtocolContracts, error) {
```
aditya1702, line 128: Why did we remove this function (didnt we introduce in earlier PR?)

internal/services/ingest.go
```diff
@@ -130,6 +131,12 @@ type ingestService struct {
 	eligibleProtocolProcessors map[string]ProtocolProcessor
 }
 
+// SetEligibleProtocolProcessorsForTest sets the eligible protocol processors for testing.
+// In production, this is set by ingestLiveLedgers before each PersistLedgerData call.
+func (m *ingestService) SetEligibleProtocolProcessorsForTest(processors map[string]ProtocolProcessor) {
```
aditya1702: I think we should not have methods on receivers just for tests to use. We should find some way to do this in the test file. I think this might be against Go best practices. Can you double check this one?

internal/services/ingest_live.go
```diff
@@ -337,6 +347,11 @@ func (m *ingestService) ingestLiveLedgers(ctx context.Context, startLedger uint3
 	}
 }
 
+// produceProtocolState runs all registered protocol processors against a ledger.
+func (m *ingestService) produceProtocolState(ctx context.Context, ledgerMeta xdr.LedgerCloseMeta, ledgerSeq uint32) error {
```
aditya1702: Same check for this function too - this function is only used in tests so is this fine adding a method for that on ingestService reciever?

internal/services/ingest_live.go
```diff
@@ -27,8 +29,8 @@ const (
 )
 
 // protocolContractCache caches classified protocol contracts to avoid per-ledger DB queries.
-// Only accessed from the single-threaded live ingestion loop, so no mutex is needed.
 type protocolContractCache struct {
+	mu                  sync.RWMutex
```
aditya1702: Is this going to be shared by goroutines?

internal/services/protocol_migrate_history.go
```diff
@@ -0,0 +1,477 @@
...
+			continue
+		}
+		active = append(active, pid)
+	}
+
+	return active, nil
+}
+
+// processAllProtocols runs history migration for all protocols using ledger-first iteration.
+// Each ledger is fetched once and processed by all eligible protocols, avoiding redundant RPC calls.
```
aditya1702: Since this is a historical ledger processing (migration), why not use the datastore backend? Because that is better suited for backfilling, with RPC we will have to ensure the historical range is supported by its history retention

internal/services/protocol_migrate_history.go
```diff
@@ -0,0 +1,477 @@
...
+			return handedOffProtocolIDs(trackers), nil
+		}
+
+		latestLedger, err := s.ingestStore.Get(ctx, s.latestLedgerCursorName)
+		if err != nil {
+			return handedOffProtocolIDs(trackers), fmt.Errorf("reading latest ingest ledger: %w", err)
+		}
+
+		// Find minimum cursor among non-handed-off trackers
+		var minCursor uint32
```
aditya1702: Why are we tracking the minCursor here?

internal/services/protocol_migrate_history.go
```diff
@@ -0,0 +1,477 @@
...
+		}
+		active = append(active, pid)
+	}
+
+	return active, nil
+}
+
+// processAllProtocols runs history migration for all protocols using ledger-first iteration.
+// Each ledger is fetched once and processed by all eligible protocols, avoiding redundant RPC calls.
+func (s *protocolMigrateHistoryService) processAllProtocols(ctx context.Context, protocolIDs []string) ([]string, error) {
```
aditya1702, line 213: Can we break this function in smaller sub functions for readability. It is very long and hard to understand on all the moving parts

internal/services/protocol_migrate_history_test.go
```diff
@@ -0,0 +1,1191 @@
+package services
```
aditya1702, line 1: Awesome test suite 👏 

I think these 2 are missing:

1. **Context cancellation during processing** — There's a select { case <-ctx.Done() } check at line 304 but no test cancels a context mid-batch. A test with
  context.WithCancel that cancels after N ledgers would verify the cleanup path returns correct handedOffProtocolIDs.
  2. **oldest_ingest_ledger is 0** — The guard at line 226 (return nil, fmt.Errorf("ingestion has not started yet")) has no test.

internal/services/protocol_migrate_history.go
```diff
@@ -0,0 +1,477 @@
...
+
+		startLedger := minCursor + 1
+		if startLedger > latestLedger {
+			log.Ctx(ctx).Infof("All protocols at or past tip %d, migration complete", latestLedger)
+			return handedOffProtocolIDs(trackers), nil
+		}
+
+		log.Ctx(ctx).Infof("Processing ledgers %d to %d for %d protocol(s)", startLedger, latestLedger, len(protocolIDs))
+
+		if err := s.ledgerBackend.PrepareRange(ctx, ledgerbackend.BoundedRange(startLedger, latestLedger)); err != nil {
```
aditya1702: I think I found a valid bug here based on my knowledge of ledger backend. I used Claude to craft this reply:

## Bug: `PrepareRange` called multiple times on `RPCLedgerBackend` — fails in production

  `processAllProtocols` calls `PrepareRange` on every iteration of the outer `for {}` loop (line 299) and again during the convergence poll (line 411).
  However, `RPCLedgerBackend.PrepareRange` [returns an error if the backend is already
  prepared](https://github.com/stellar/go-stellar-sdk/blob/main/ingest/ledgerbackend/rpc_backend.go#L176-L178):

  ```go
  if b.preparedRange != nil {
      return fmt.Errorf("RPCLedgerBackend is already prepared with range [%d, %d]",
          b.preparedRange.from, b.preparedRange.to)
  }
  ```

  There is no reset mechanism — `preparedRange` is only `nil` at construction time.

  ### Affected paths

  | Path | Where | What happens |
  |------|-------|-------------|
  | Tip advances after first batch | Line 299 (second outer loop iteration) | `PrepareRange` returns error → function returns with failure |
  | Convergence poll | Line 411 | `PrepareRange(UnboundedRange(...))` returns error → misinterpreted as transient error → infinite retry loop |
  | Transient poll error retry | Line 421 → `continue` → line 299 | Same as tip-advance case |

  ### Why tests don't catch this

  All test backends (`multiLedgerBackend`, `transientErrorBackend`, `rangeTrackingBackend`) implement `PrepareRange` as a no-op that always returns `nil`.
  None of them enforce the single-prepare constraint that `RPCLedgerBackend` has.

  ### Impact

  The function works correctly **only when the entire backfill completes in a single batch** — i.e., all ledgers from `startLedger` to `latestLedger` are
  processed and the tip doesn't advance during that time. If the tip advances or the convergence poll is reached, the service errors out.

internal/services/protocol_migrate_history.go
```diff
@@ -0,0 +1,477 @@
...
+		// Check if tip has advanced
+		newLatest, err := s.ingestStore.Get(ctx, s.latestLedgerCursorName)
+		if err != nil {
+			return handedOffProtocolIDs(trackers), fmt.Errorf("re-reading latest ingest ledger: %w", err)
+		}
+		if newLatest > latestLedger {
+			continue
+		}
+
+		// At tip — poll briefly for convergence.
```
aditya1702: Following from the `PrepareRange` bug above — rather than patching the re-call sites, the entire outer loop and convergence poll can be replaced with a
  single `UnboundedRange`.

  ### Current design

  ```
  PrepareRange(BoundedRange(start, latest))    ← first call
    process ledgers start..latest
    tip advanced?
      PrepareRange(BoundedRange(next, newLatest))  ← BUG: errors on RPCLedgerBackend
    at tip?
      PrepareRange(UnboundedRange(latest+1))       ← BUG: same
      GetLedger with 5s timeout (convergence poll)
      new ledger? loop again → PrepareRange(Bounded...) ← BUG: same
  ```

  ### Proposed design

  ```
  PrepareRange(UnboundedRange(startLedger))    ← once, at the top

  for seq := startLedger; ; seq++ {
      // for each tracker, check if converged (via timeout on GetLedger)
      pollCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
      meta, err := s.ledgerBackend.GetLedger(pollCtx, seq)
      cancel()
      if errors.Is(err, context.DeadlineExceeded) {
          // no new ledger in 5s → converged at seq-1
          break
      }
      // process meta for eligible trackers...
  }
  ```

  ### Why this works with `RPCLedgerBackend`

  - `PrepareRange` is called exactly once — no conflict with the single-prepare constraint
  - `UnboundedRange` means `GetLedger` accepts any sequence number without range validation
  - `RPCLedgerBackend.GetLedger` already has built-in 2s polling when a ledger is beyond the RPC tip — so the timeout context is the only convergence
  mechanism needed
  - Bounded vs Unbounded has no behavioral difference in `RPCLedgerBackend` other than the range check at [line
  131](https://github.com/stellar/go-stellar-sdk/blob/main/ingest/ledgerbackend/rpc_backend.go#L131)

  ### What gets removed

  - The outer `for {}` loop (lines 268–441)
  - `minCursor` recomputation on each iteration
  - The 40-line convergence poll section (lines 402–440)
  - All Bounded/Unbounded mode switching
  - The `PrepareRange` bug entirely

aditya1702 (commented):
Found some issues with the migration logic

internal/services/protocol_migrate_history.go
aditya1702: Btw this same refactor should also work with the datastore backend since that also supports live ingestion of new ledgers

aristidesstaffieri: I changed to use the unbounded range in 3c7714fb58b5d19a02f30884d9db46ccd0d8cbf4

internal/services/protocol_migrate_history.go
aristidesstaffieri: this was fixed in 3c7714fb58b5d19a02f30884d9db46ccd0d8cbf4

internal/services/ingest.go
aristidesstaffieri: removed in a5a260c5a8ea9c116584b36b27435cbf165d8079

internal/services/ingest_live.go
aristidesstaffieri: removed in a5a260c5a8ea9c116584b36b27435cbf165d8079

internal/data/protocol_contracts.go
aristidesstaffieri: looks like this was an artifact of a rebase, I restored it in 9a4f29b0dfdb93425c6682e0498ccd1f4f8bd4fb

internal/services/ingest_live.go
aristidesstaffieri: no this was added defensively but is not needed, removed in 64de5eceaeb1b860f8bae2f07242d92c72853014

internal/services/protocol_migrate_history.go
aristidesstaffieri: Yeah this should be using datastore backend, I changed it in 4ee42c9d7a466e1bd0cd907f35b2c32358931e44

internal/services/protocol_migrate_history_test.go
aristidesstaffieri: these cases were added in 3c7714fb58b5d19a02f30884d9db46ccd0d8cbf4

internal/services/protocol_migrate_history.go
aristidesstaffieri: this is now much smaller since the changes in 3c7714fb58b5d19a02f30884d9db46ccd0d8cbf4

internal/services/protocol_migrate_history.go
aristidesstaffieri: this was removed as part of the refactor in 3c7714fb58b5d19a02f30884d9db46ccd0d8cbf4

internal/services/protocol_migrate_history.go
```diff
@@ -0,0 +1,509 @@
...
+		},
+	); prepErr != nil {
+		return handedOffProtocolIDs(trackers), fmt.Errorf("preparing unbounded range from %d: %w", startLedger, prepErr)
+	}
+
+	for seq := startLedger; ; seq++ {
+		if err := ctx.Err(); err != nil {
+			return handedOffProtocolIDs(trackers), fmt.Errorf("context cancelled: %w", err)
+		}
+		if allHandedOff(trackers) {
```
aditya1702, line 252: Correct me if I am wrong: as the migration proceeds migration trackers will be handed off one by one right? So why do we need the `errConverged` logic? Wont we simply exit the loop once `allHandedOff = true`?

aditya1702 (approved):

internal/services/protocol_migrate_history.go
aristidesstaffieri: this was meant to handle a case where the datastore doesnt have the ledger the migration is asking for, in practice this can't really happen and if it does we can just wait for it. 

Refactored to remove this in https://github.com/stellar/wallet-backend/pull/546/changes/ef6392b46dcb6b22b8e2babaa3b81fef44045499

PR #549, opened 2026-03-26
Remove `envelope_xdr` and `meta_xdr` from transactions table

### LINKED ISSUE #527: Remove `envelope_xdr` from `transactions` table

aditya1702:
Closes #527 #528 

We do not need these fields stored in wallet backend and will use RPC to get them on demand if needed


internal/data/transactions.go
```diff
@@ -346,11 +344,3 @@ func (m *TransactionModel) BatchCopy(
 
 	return len(txs), nil
 }
-
-// pgtypeTextFromPtr converts a *string to pgtype.Text for efficient binary COPY.
-func pgtypeTextFromPtr(s *string) pgtype.Text {
```
aristidesstaffieri, line 351: this seems like a pretty general util, any reason to not pull it out into the utils package?

aditya1702: @aristidesstaffieri but this function is not used right now so why not add it later when its need arises?

aristidesstaffieri: ah yeah apologies, I didn't notice it was removed, that sounds good.

aristidesstaffieri (approved):

PR #573, opened 2026-04-15
Remove `balancesByAccountAddresses` endpoint

### LINKED ISSUE #568: Remove `balancesByAccountAddresses` endpoint
The wallet-backend's balances endpoint is called by freighter backend V2 and we should add the multi account balance fetch logic on BE v2 side instead of wallet-backend

aditya1702:
Closes #568 

Remove the multi address balance endpoint since we will add it on freighter backend V2 side. This keeps the graphql API clean.


aristidesstaffieri:
We should still add per request bounded pool like in https://github.com/stellar/wallet-backend/pull/571 to prevent any single request from hogging the pool

aristidesstaffieri (approved):

PR #574, opened 2026-04-15
Remove `balanceByAccountAddress` endpoint and make it callable from the `accountByAddress` query

### LINKED ISSUE #570: `AccountByAddress` should support querying the balances of an account
Currently, we have a separate endpoint for querying balances: `BalanceByAccountAddress` but we should also support querying it through the `AccountByAddress` endpoint which already supports querying transactions, operations and state changes.

aditya1702:
### What

- Removed the `balancesByAccountAddress` root query from the GraphQL schema.
- Added `balances: [Balance!]` to `Account`, so balances are now fetched via `accountByAddress(address) { balances }`.
- Moved the balance-loading flow into the `Account` field resolver and shared helper logic, while keeping existing balance behavior intact for native, trustline, SAC, and SEP-41 balances.
- Updated the repo-owned GraphQL client to use the nested query shape and renamed `GetBalancesByAccountAddress` to `GetAccountBalances`.
- Regenerated gqlgen code with `make gql-generate`.
- Updated resolver tests, middleware tests, integration tests, and docs/examples to match the new schema shape.

### Why

- This makes the schema more idiomatic GraphQL by treating `accountByAddress` as the entry point and exposing account-scoped data as fields on `Account`.
- `balances` was the outlier in the current API; `transactions`, `operations`, and `stateChanges` already hang off `Account`, so this change makes the account graph consistent.
- Removing the extra root query simplifies the schema and avoids maintaining two ways to access the same account resource.
- Since the backend is not yet in production and has no external clients, this breaking cleanup can be done now without a compatibility layer.
- The change is schema-shape only; it does not change balance semantics or supported balance types.

### Known limitations

N/A

Closes #570

### DIFF
--- a/internal/serve/graphql/resolvers/account_balances.go
+++ b/internal/serve/graphql/resolvers/account_balances.go
@@ -2,7 +2,11 @@ package resolvers
 
 import (
 	"context"
+	"fmt"
+	"slices"
+	"strings"
 
+	"github.com/google/uuid"
 	"github.com/stellar/go-stellar-sdk/support/log"
 	"github.com/vektah/gqlparser/v2/gqlerror"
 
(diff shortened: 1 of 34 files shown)


internal/serve/graphql/resolvers/account_balances.go
```diff
@@ -0,0 +1,90 @@
...
+			sep41TokenIDs = append(sep41TokenIDs, contract.ContractID)
+			contractsByContractID[contract.ContractID] = contract
+		}
+	}
+
+	if len(sep41TokenIDs) == 0 {
+		return balances, nil
+	}
+
+	sep41Balances, err := getSep41Balances(ctx, address, r.contractMetadataService, sep41TokenIDs, contractsByContractID, r.pool)
```
aristidesstaffieri: This still suffers from the complexity fan out issue.

GraphQL aliasing: ~1000 aliased accountByAddress calls fit inside the default 1000-complexity budget (both accountByAddress and Account.balances carry the default weight of 1 because no custom complexity function is registered for them). Each balances selection invokes Resolver.getAccountBalances → getSep41Balances, which submits one task per SEP‑41 contract held by the target account into the shared worker pool.

Amplification: N_aliases × K_sep41_tokens tasks on the shared pool (default 100 workers). An attacker can pick or seed any address with many SEP‑41 holdings to maximize K, starving the pool.

aristidesstaffieri:
### Code review

Found 1 issue:

1. `Account.balances` has no complexity function registered, leaving it exposed to alias-based amplification. The PR adds `balances: [Balance\!]` to `Account` and wires `getAccountBalances` (DB reads for native/trustlines/SAC/contract tokens, plus sequential RPC calls per SEP-41 token), but `addComplexityCalculation` registers only `Account.Transactions/Operations/StateChanges` and leaves `Account.Balances` unset — gqlgen defaults that to a flat complexity of 1. With the default 1000 limit, ~1000 aliased `accountByAddress(address){balances}` selections fit the budget, and each one fans out into multiple DB queries + N serial RPC calls (N unbounded by `GetByAccount` and now serial after f8d81be8). The same amplification is reachable through `transactions{edges{node{accounts{balances}}}}` / `operations{...{accounts{balances}}}`. Making RPC calls sequential removes the shared-pool exhaustion vector but doesn't bound per-request latency or total cross-request load. This is the same fan-out concern previously raised inline on `account_balances.go`; it remains unaddressed at HEAD.

https://github.com/stellar/wallet-backend/blob/f8d81be8c20a424bb2db6b55b173b60ea46284a9/internal/serve/serve.go#L374-L388

https://github.com/stellar/wallet-backend/blob/f8d81be8c20a424bb2db6b55b173b60ea46284a9/internal/serve/graphql/resolvers/account_balances.go#L11-L90

aristidesstaffieri:
> ### Code review
> Found 1 issue:
> 
> 1. `Account.balances` has no complexity function registered, leaving it exposed to alias-based amplification. The PR adds `balances: [Balance\!]` to `Account` and wires `getAccountBalances` (DB reads for native/trustlines/SAC/contract tokens, plus sequential RPC calls per SEP-41 token), but `addComplexityCalculation` registers only `Account.Transactions/Operations/StateChanges` and leaves `Account.Balances` unset — gqlgen defaults that to a flat complexity of 1. With the default 1000 limit, ~1000 aliased `accountByAddress(address){balances}` selections fit the budget, and each one fans out into multiple DB queries + N serial RPC calls (N unbounded by `GetByAccount` and now serial after [f8d81be](https://github.com/stellar/wallet-backend/commit/f8d81be8c20a424bb2db6b55b173b60ea46284a9)). The same amplification is reachable through `transactions{edges{node{accounts{balances}}}}` / `operations{...{accounts{balances}}}`. Making RPC calls sequential removes the shared-pool exhaustion vector but doesn't bound per-request latency or total cross-request load. This is the same fan-out concern previously raised inline on `account_balances.go`; it remains unaddressed at HEAD.
> 
> https://github.com/stellar/wallet-backend/blob/f8d81be8c20a424bb2db6b55b173b60ea46284a9/internal/serve/serve.go#L374-L388
> 
> https://github.com/stellar/wallet-backend/blob/f8d81be8c20a424bb2db6b55b173b60ea46284a9/internal/serve/graphql/resolvers/account_balances.go#L11-L90
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

you can ignore this, I see it is resolved in #577

aristidesstaffieri:
### Code review (follow-up)

Found 1 additional issue:

1. The PR changes `paginatedQueryComplexityFunc`'s default multiplier from `10` to `int(graphqlutils.DefaultPageLimit)` = `50`, but the illustrative comment at `serve.go:358` still reads `Complexity = 10*(1+1+1+2*(1+1+1+5*(1+1+1+1))) = 490`. The arithmetic is still valid for the explicit-args query shown, but it no longer reflects the new default behavior — an equivalent query with `first` omitted now scores `2450`, a 5× increase. Unparameterized `transactions`/`operations`/`stateChanges` root queries that passed the default 1000 complexity limit before this PR may now be rejected. Either the example should be updated to match the new default (and the prose adjusted), or the complexity limit should be raised in tandem.

https://github.com/stellar/wallet-backend/blob/f8d81be8c20a424bb2db6b55b173b60ea46284a9/internal/serve/serve.go#L321-L389

https://github.com/stellar/wallet-backend/blob/f8d81be8c20a424bb2db6b55b173b60ea46284a9/internal/serve/graphql/utils.go#L14-L16

aditya1702:
> ### Code review (follow-up)
> Found 1 additional issue:
> 
> 1. The PR changes `paginatedQueryComplexityFunc`'s default multiplier from `10` to `int(graphqlutils.DefaultPageLimit)` = `50`, but the illustrative comment at `serve.go:358` still reads `Complexity = 10*(1+1+1+2*(1+1+1+5*(1+1+1+1))) = 490`. The arithmetic is still valid for the explicit-args query shown, but it no longer reflects the new default behavior — an equivalent query with `first` omitted now scores `2450`, a 5× increase. Unparameterized `transactions`/`operations`/`stateChanges` root queries that passed the default 1000 complexity limit before this PR may now be rejected. Either the example should be updated to match the new default (and the prose adjusted), or the complexity limit should be raised in tandem.
> 
> https://github.com/stellar/wallet-backend/blob/f8d81be8c20a424bb2db6b55b173b60ea46284a9/internal/serve/serve.go#L321-L389
> 
> https://github.com/stellar/wallet-backend/blob/f8d81be8c20a424bb2db6b55b173b60ea46284a9/internal/serve/graphql/utils.go#L14-L16
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

@aristidesstaffieri I fixed this in https://github.com/stellar/wallet-backend/pull/574/commits/5cba0ba4f00063800120761625e6dfbc7e686930

aristidesstaffieri (approved):

PR #586, opened 2026-04-20
Store account address in balance tables as `BYTEA` instead of `TEXT`

### LINKED ISSUE #582: Store account ID as `BYTEA` instead of `TEXT` in all balance tables
This is more efficient storage wise and also reduces index size without any tradeoffs on query performance. It also matches the way we store account IDs on transactions, operations and state changes tables

aditya1702:
Closes #582 

I also changed the Upsert methods to use an `UNNEST` query instead of `Batch.Queue` for better efficiency. The batch method makes N individual calls to postgres vs UNNEST which upserts in 1 single call

### DIFF
--- a/internal/data/mocks.go
+++ b/internal/data/mocks.go
@@ -8,6 +8,8 @@ import (
 	"github.com/google/uuid"
 	"github.com/jackc/pgx/v5"
 	"github.com/stretchr/testify/mock"
+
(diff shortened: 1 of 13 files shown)


aristidesstaffieri:
### Code review

aristidesstaffieri (approved):

PR #589, opened 2026-04-21
fix(internal/services): bound wasm classification cost and isolate panics

aristidesstaffieri:
### What

Harden the WASM classification path in `protocol-setup` so a single attacker-crafted bytecode returned by RPC cannot halt classification.

- Reject WASM blobs larger than 2 MiB in `wasmSpecExtractor.ExtractSpec` before calling wazero.
- Wrap `wazero.Runtime.CompileModule` in a 10s timeout via a child `context.WithTimeout`.
- Cap the `xdr.Unmarshal` loop over `contractspecv0` at 10,000 `ScSpecEntry` entries.
- Extract the per-WASM classify body into `classifyOne` and wrap it in `defer/recover`, so a panic on one blob is logged and the loop continues.
- Tests added: oversize-WASM rejection, cancelled-caller-context propagation, timeout-constant sanity check, and a `panicOnceExtractor` that proves a panicking WASM no longer aborts the classify run.

Files touched: `internal/services/protocol_validator.go`, `internal/services/protocol_setup.go`, and their test files.

### Why

`protocol-setup` fetches bytecode for every unclassified WASM hash via RPC and feeds it through `wazero.CompileModule` + `xdr.Unmarshal`. Before this change:

- There was no upper bound on `len(wasmCode)`, no per-call deadline, and no cap on the XDR decode loop. A hostile or pathological WASM could make compile or decode consume unbounded CPU/memory.
- A panic in wazero or the XDR decoder would escape the per-WASM log-and-skip path and abort the whole run.

When `protocol-setup` fails mid-run, `classification_status` is left at `in_progress`, which blocks `protocol-migrate history` and `protocol-migrate current-state` (both assert `classification_status == success`). Recovery requires an operator to manually reset DB state. Any Soroban deployer can reach this code path, so the failure mode is directly attacker-triggerable.

This PR bounds every dimension of per-WASM cost (size, time, entry count) and ensures a single bad blob is logged and skipped rather than halting the entire classification. 

### Known limitations

The chosen bounds are conservative:

- `maxWasmBytes = 2 MiB` is wider than the current on-chain contract-code limit (~64 KiB) to stay correct across future protocol upgrades.
- `wasmCompileTimeout = 10 s` is ~20–50× the p99 of a legitimate compile.
- `maxSpecEntries = 10_000` is ~30–100× any real contract.

CLI-level signal handling (so an operator can Ctrl-C a stuck `protocol-setup`) is intentionally out of scope here and tracked separately under the "CLI commands use `context.Background()`" finding, which will land on its own branch.

N/A

#### PR Structure

#### Thoroughness

- [ ] All updated queries have been tested (refer to [this](https://stackoverflow.com/a/29163465) check if the data set returned by the updated query is expected to be same as the original one). _N/A — no query changes._

#### Release

- [ ] The new functionality is gated with a feature flag if this is not ready for production. _N/A — this is a hardening fix on an unreleased feature branch._

### DIFF
--- a/internal/data/mocks.go
+++ b/internal/data/mocks.go
@@ -326,4 +326,3 @@ func (m *ProtocolContractsModelMock) BatchGetByProtocolIDs(ctx context.Context, 
 	}
 	return args.Get(0).(map[string][]ProtocolContracts), args.Error(1)
 }
-
--- a/internal/data/mocks.go
+++ b/internal/data/mocks.go
@@ -326,3 +326,4 @@ func (m *ProtocolContractsModelMock) BatchGetByProtocolIDs(ctx context.Context, 
 	}
 	return args.Get(0).(map[string][]ProtocolContracts), args.Error(1)
 }
+
--- a/internal/integrationtests/main_test.go
+++ b/internal/integrationtests/main_test.go
@@ -71,18 +71,6 @@ func TestIntegrationTests(t *testing.T) {
 		})
 	})
 
-	// Phase 1: Validate balances from checkpoint before fixture transactions
-	t.Run("AccountBalancesAfterCheckpointTestSuite", func(t *testing.T) {
-		suite.Run(t, &AccountBalancesAfterCheckpointTestSuite{
-			testEnv: testEnv,
-		})
-	})
-
-	// Only proceed if checkpoint balance validation succeeded
-	if t.Failed() {
-		t.Fatal("AccountBalancesAfterCheckpointTestSuite failed, skipping remaining tests")
-	}
-
 	// Data migration tests — protocol setup plus setup-to-live protocol state production
 	t.Run("DataMigrationTestSuite", func(t *testing.T) {
 		suite.Run(t, &DataMigrationTestSuite{testEnv: testEnv})
--- a/internal/serve/graphql/resolvers/account_balances.go
+++ b/internal/serve/graphql/resolvers/account_balances.go
@@ -507,3 +507,4 @@ func (r *Resolver) getSACBalanceNodes(
 
 	return nodes, nil
 }
+
--- a/internal/serve/graphql/resolvers/account_balances.go
+++ b/internal/serve/graphql/resolvers/account_balances.go
@@ -507,4 +507,3 @@ func (r *Resolver) getSACBalanceNodes(
 
 	return nodes, nil
 }
-
--- a/internal/serve/graphql/resolvers/account_balances_utils.go
+++ b/internal/serve/graphql/resolvers/account_balances_utils.go
@@ -99,4 +99,3 @@ func buildTrustlineBalanceFromDB(trustline data.TrustlineBalance, networkPassphr
 		IsAuthorizedToMaintainLiabilities: isAuthorizedToMaintainLiabilities,
 	}, nil
 }
-
--- a/internal/serve/graphql/resolvers/account_balances_utils.go
+++ b/internal/serve/graphql/resolvers/account_balances_utils.go
@@ -99,3 +99,4 @@ func buildTrustlineBalanceFromDB(trustline data.TrustlineBalance, networkPassphr
 		IsAuthorizedToMaintainLiabilities: isAuthorizedToMaintainLiabilities,
 	}, nil
 }
+
--- a/internal/services/protocol_setup.go
+++ b/internal/services/protocol_setup.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"runtime/debug"
 	"time"
 
 	"github.com/jackc/pgx/v5"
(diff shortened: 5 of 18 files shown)


internal/services/wasm_spec_extractor_test.go
```diff
@@ -64,6 +65,40 @@ func TestWasmSpecExtractor_RealWasm(t *testing.T) {
...
+	extractor := NewWasmSpecExtractor()
+	defer func() { require.NoError(t, extractor.Close(context.Background())) }()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	// Any input under the size cap is fine; the extractor should short-circuit
+	// because the caller's context is already done once we derive a child.
+	specs, err := extractor.ExtractSpec(ctx, []byte("not a valid wasm"))
+	require.Error(t, err)
```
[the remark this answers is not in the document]
aristidesstaffieri: fixed in 0d1ee96e0c0c69b40a6845cbd4d82a06275fad7f

internal/services/protocol_setup.go
```diff
@@ -207,6 +198,38 @@ func (s *protocolSetupService) classify(ctx context.Context, protocolIDs []strin
...
+	wasmHash string,
+	bytecode []byte,
+	protocolIDs []string,
+	validatorsByProtocol map[string]ProtocolValidator,
+	matchedHashes map[string][]types.HashBytea,
+) {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Ctx(ctx).Errorf("panic classifying WASM %s, skipping: %v", wasmHash, r)
+		}
```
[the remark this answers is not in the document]
aristidesstaffieri: added the stack in 6bc8716ef9f96af6089b416863975180321cb8f0

internal/services/protocol_validator.go
```diff
@@ -38,8 +59,18 @@ func NewWasmSpecExtractor() WasmSpecExtractor {
...
+		return nil, fmt.Errorf("wasm module too large: %d bytes (max %d)", len(wasmCode), maxWasmBytes)
+	}
+
+	compileCtx, cancel := context.WithTimeout(ctx, wasmCompileTimeout)
+	defer cancel()
+
+	compiledModule, err := e.runtime.CompileModule(compileCtx, wasmCode)
 	if err != nil {
 		return nil, fmt.Errorf("compiling WASM module: %w", err)
 	}
```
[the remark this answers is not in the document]
aristidesstaffieri: this is hardened in 4fa003eba72ccba95fae7b57177a1eb24b3cc3a5

internal/data/mocks.go
aditya1702, line 328: line removed by mistake?

internal/integrationtests/main_test.go
aditya1702, line 75: Why is this removed?

internal/serve/graphql/resolvers/account_balances.go
aditya1702, line 509: newline removed?

internal/serve/graphql/resolvers/account_balances_utils.go
aditya1702, line 101: same here

internal/services/protocol_validator.go
```diff
@@ -38,13 +65,35 @@ func NewWasmSpecExtractor() WasmSpecExtractor {
...
 	defer func() {
-		if closeErr := compiledModule.Close(ctx); closeErr != nil {
+		// Detach from the caller's ctx: if the caller cancels while we are
+		// decoding spec entries, Close should still run to release wazero's
+		// compiled state rather than returning early with a ctx error.
```
aditya1702, line 93: I dont understand this - why are we creating a closeCtx and not just passing the caller ctx here?

aristidesstaffieri: `compiledModule.Close(ctx)` honors ctx cancellation in wazero. If we passed the caller ctx here, a cancellation during the `xdr.Unmarshal` loop just below would cause Close to return early and leak wazero's compiled native code / compiler bookkeeping — the exact teardown the defer exists for would get skipped.

Deriving `closeCtx` from `context.Background()` decouples teardown from caller cancellation so Close always runs. `wasmCloseTimeout` (5s) still bounds it so a wedged runtime can't hang the defer.

This was the Copilot finding above on the same line; landed in 4fa003eb.

internal/integrationtests/main_test.go
aristidesstaffieri: Not actually removed — this is a de-duplication. The same `t.Run("AccountBalancesAfterCheckpointTestSuite", ...)` block still exists earlier in the file at `main_test.go:38-48`, so the suite is still invoked before fixture transactions.

At parent `af076f3a` the file had two identical copies of this block (lines 39-48 and 75-84) — a duplicate that slipped in during an earlier merge/rebase. This drops the second copy. The cleanup is already on `feature/data-migrations` (commit 4930aa69); it's showing in this PR's diff because the merge-base pre-dates that commit.

internal/data/mocks.go
aristidesstaffieri: hmm yeah I'm not sure if this happened as part of gofmt, added them back in 8bc52e12205fd60eb717505f82de2c930150ba4a

internal/services/protocol_validator.go
aditya1702: Ah got it.

aditya1702 (approved):

PR #592, opened 2026-04-21
fix(internal): tighten SEP-41 validator signature matching

aristidesstaffieri:
### What

Tightens SEP-41 validation to require exact ordered input and output signatures when classifying contracts. The validator now compares ordered slices instead of normalizing inputs to a map and outputs to a set, and the test suite adds regression coverage for reordered transfer params and duplicate balance outputs.

### Why

The previous implementation could misclassify non-SEP-41 contracts as SEP-41 when function parameters were reordered or outputs were duplicated. That creates a security-sensitive false-positive path where attacker-controlled contracts can be pulled into SEP-41 indexing and state production.

### Known limitations
N/A

N/A

#### PR Structure

#### Thoroughness

- [x] All updated queries have been tested (no queries changed).

#### Release

- [x] The new functionality is gated with a feature flag if this is not ready for production. (N/A - validator fix only.)

### DIFF
--- a/internal/services/sep41_validator.go
+++ b/internal/services/sep41_validator.go
@@ -34,8 +34,13 @@ var scSpecTypeNames = map[xdr.ScSpecType]string{
 // sep41FunctionSpec defines the expected signature for a SEP-41 token function.
 type sep41FunctionSpec struct {
 	name            string
-	expectedInputs  map[string]string
+	expectedInputs  []sep41FunctionInputSpec
 	expectedOutputs []string
+}
+
+type sep41FunctionInputSpec struct {
+	name     string
+	typeName string
 }
 
 // sep41RequiredFunctions defines all required functions for SEP-41 token standard compliance.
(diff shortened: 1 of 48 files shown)


internal/services/sep41_validator.go
```diff
@@ -34,89 +34,94 @@ var scSpecTypeNames = map[xdr.ScSpecType]string{
 // sep41FunctionSpec defines the expected signature for a SEP-41 token function.
 type sep41FunctionSpec struct {
 	name            string
-	expectedInputs  map[string]string
+	expectedInputs  []sep41FunctionInputSpec
 	expectedOutputs []string
 }
 
+type sep41FunctionInputSpec struct {
```
aditya1702: nit: since we will have this same struct used for other protocol specs, can we rename it `contractFunctionInputSpec` or something more general?

aditya1702 (approved):

internal/services/sep41_validator.go
aristidesstaffieri: renamed in 632a8fd9013e21aba9b2e5b3f0773ff64947243c

PR #577, opened 2026-04-15
Add Relay-style cursor pagination for account balances queries

aditya1702:
Closes #576 

  ## What

  Converts `Account.balances` from a flat list (`[Balance!]`) to a Relay cursor connection (`BalanceConnection!`) with `first/after/last/before` pagination, and adds keyset pagination support to every backing data source.

  ### Schema

  - `Account.balances` changed from `[Balance!]` to `balances(first, after, last, before): BalanceConnection!`
  - Added `BalanceConnection` and `BalanceEdge` types in `pagination.graphqls`

  ### Resolver layer (`account_balances.go`: +530 lines)

  - Multi-source pagination engine that merges native XLM, classic trustlines, SAC, and SEP-41 balances into one Relay connection
  - Source-aware opaque cursors (`v1:<source>:<id>`) so pagination can resume across heterogeneous balance types
  - Forward and backward source walkers that skip fully-consumed sources and apply the keyset cursor only to the source that owns the page boundary
  - SEP-41 sentinel optimization: the extra "+1" row for `hasNextPage` is carried as a cursor-only node to avoid a wasted RPC call

  ### Data layer

  - `TrustlineBalanceModel.GetByAccount` — now accepts `limit`, `cursor` (asset UUID), and `sortOrder` for keyset pagination
  - `SACBalanceModel.GetByAccount` — same keyset pagination parameters, cursoring by contract UUID
  - **New:** `AccountContractTokensModel.GetSEP41ByAccount` — paginated query for SEP-41 contracts by junction-table UUID
  - Updated mocks and added pagination tests for all three models

  ### Shared pagination infrastructure (`utils.go`)

  - Added `CursorTypeString` cursor type and `StringCursor` field on `PaginationParams`
  - `NewConnectionWithRelayPagination` now considers `StringCursor` when computing `hasCursor`

  ### Wiring

  - `balanceReaderAdapter` updated to pass pagination params through to the data models
  - `BalanceReader` interface signatures widened with `limit/cursor/sortOrder`
  - Complexity function registered for `Account.balances`
  - `pkg/wbclient` updated to query the new `edges { node { ... } }` shape

  ## Why

  ### The parent branch moved balances onto `Account` but without pagination

  `feature/balance-from-account-query` relocated balance fetching from root queries to `Account.balances`, but returned a flat `[Balance!]` list. This meant every query still loaded **all** balances for an account in one shot — unbounded response size and unbounded RPC fan-out for SEP-41 contracts.

  ### Accounts can have many balances

[...]

PR #609, opened 2026-05-07
Add balance pagination parameters to the Go client

### LINKED ISSUE #608: Add balance pagination parameters to the Go client
We added pagination to balances graphql query but forgot to update the `GetAccountBalances` client method with the pagination params. We have to update it.

aditya1702:
## Summary

Closes #608.

The `Account.balances` GraphQL query has supported full Relay-style pagination (`first`/`after`/`last`/`before`) since #577. However, the Go SDK method `GetAccountBalances` was never updated — it hardcoded `first: 100`, looped internally to fetch every page, and returned a flat `[]types.Balance`. That made it the only paginated query in the SDK without pagination args.

This PR closes that gap and ships two follow-ups (a convenience helper and a tightening of the polymorphic decode trust boundary) on top.

### Core: paginated `GetAccountBalances`

- **New types** in `pkg/wbclient/types/types.go`: `BalanceEdge` and `BalanceConnection`. `BalanceEdge` has a custom `UnmarshalJSON` that dispatches the polymorphic `Balance` node through `UnmarshalBalance` — same pattern as the existing `StateChangeEdge`.
- **GraphQL query string** in `pkg/wbclient/queries.go`: declares all four Relay pagination variables (`$first`, `$after`, `$last`, `$before`) and requests the per-edge `cursor` plus full `PageInfo` (`startCursor`, `endCursor`, `hasNextPage`, `hasPreviousPage`).
- **`GetAccountBalances` signature** in `pkg/wbclient/client.go`:
  ```go
  func (c *Client) GetAccountBalances(
      ctx context.Context,
      address string,
      first, last *int32,
      after, before *string,
  ) (*types.BalanceConnection, error)
  ```
  Internal pagination loop is removed. Pagination args go through the existing `buildPaginationVars` helper, which validates them with the same rules as every other paginated method.

### Follow-ups in this PR

- **`GetAllAccountBalances` convenience helper** in `pkg/wbclient/client.go`. Drives forward pagination internally with a fixed page size and returns a flat `[]types.Balance` — the one-call "fetch every balance for this address" path, for callers that don't need page-level control. Returns a **non-nil empty slice** when the account has no balances (so callers that JSON-marshal the result get `[]` on the wire, not `null`). Pagination errors are wrapped with `"getting account balances page: %w"` so failures partway through the loop are distinguishable from a single-page error.
- **Strict polymorphic decode at both layers.** The schema declares `edges: [BalanceEdge!]!` and `node: Balance!` — both non-null. Two custom `UnmarshalJSON` methods enforce that contract:
  - `BalanceEdge.UnmarshalJSON` rejects null/missing `node` fields inside an edge object (was previously absorbed silently as `Node = nil`).
  - `BalanceConnection.UnmarshalJSON` rejects `null` array elements in `edges` — the case that `BalanceEdge.UnmarshalJSON` cannot catch because Go's `encoding/json` never routes a null array element through a custom decoder.

  Together they guarantee that any successfully-decoded `*BalanceConnection` has non-nil edges *and* non-nil `edge.Node` values, so callers can drop defensive nil-checks (the freighter-backend currently has these).
- **`(*BalanceConnection).Balances() []Balance` helper** for callers that want the flat slice without writing the edge-iteration loop themselves. This is a thin low-level transformation and returns `nil` for an empty connection; the public `GetAllAccountBalances` owns the stable non-nil empty contract.
- **Pinned the new contracts** with unit tests in `pkg/wbclient/types/types_test.go` (rejects null/missing node, rejects null edge slot in an array, decodes a valid native balance, decodes a full valid connection, helper covers nil-receiver / empty / populated / nil-edge-skip cases).

`StateChangeEdge` and `StateChangeConnection` are intentionally untouched in this PR — that surface has its own consumers and tightening it is a separate refactor.

### Tests

- The six existing checkpoint and live-ingestion integration tests (the "fetch all balances for an account" cases) now call `GetAllAccountBalances` and iterate `[]Balance` directly.
- Two new pagination integration tests: `TestCheckpoint_Account1_ForwardPagination` and `TestCheckpoint_Account1_BackwardPagination`. These call `GetAccountBalances` explicitly and round-trip cursors via `EndCursor`/`after` and `StartCursor`/`before` respectively. The backward test is the most important regression-defense: the previous SDK query did not even declare `\$last`/`\$before`, so a forward-only test would have missed that whole class of bug.

## Breaking change

This is a deliberate breaking change to the Go SDK. The signature change cannot be hidden behind a backwards-compatible wrapper without keeping two methods around, and the auto-pagination behavior is being removed from `GetAccountBalances` by design — every other paginated SDK method requires the caller to drive pagination explicitly, and `GetAccountBalances` should match. Consumers that want one-call "fetch all" semantics use the new `GetAllAccountBalances`.

The only internal caller (`internal/integrationtests/account_balances_test.go`) is updated in this PR.

## Test plan

- [x] `go build ./...`
- [x] `go test -race ./pkg/wbclient/... -timeout 2m` — 63 tests pass (53 prior + 10 new types tests)
- [ ] Integration tests (Docker required, run locally before merge):
  ```bash
  ENABLE_INTEGRATION_TESTS=true go test -v -race \
    ./internal/integrationtests/ \
    -run 'TestIntegrationTests/AccountBalancesAfterCheckpointTestSuite' \
    -timeout 30m
  ```
  Note: integration suites are subtests of a single `TestIntegrationTests` function (`internal/integrationtests/main_test.go:17`), so `-run` patterns must be slash-qualified.
- [ ] `make check` (lint, vet, etc.)

### DIFF
--- a/pkg/wbclient/types/types.go
+++ b/pkg/wbclient/types/types.go
@@ -338,22 +338,27 @@ type BalanceConnection struct {
 }
 
 // UnmarshalJSON implements custom JSON unmarshaling for BalanceConnection
-// to reject null entries in the edges array. The GraphQL schema declares
-// edges as [BalanceEdge!]!, so a null edge is a server bug; failing fast
-// at decode time surfaces it instead of silently dropping data downstream.
+// and enforces the schema's non-null guarantees. The GraphQL schema declares
+// edges as [BalanceEdge!]! — both the field itself and each element are
+// non-null — so each of the following is a server bug and is rejected here:
+//   - a missing or null edges field on the connection
+//   - a null entry within the edges array
 //
-// Null nodes inside an edge are caught by BalanceEdge.UnmarshalJSON;
-// this method covers the array-element case that the standard library
-// never routes through that decoder.
+// Null nodes inside an edge object are caught separately by
+// BalanceEdge.UnmarshalJSON.
 func (c *BalanceConnection) UnmarshalJSON(data []byte) error {
 	type tempConnection struct {
-		Edges    []*BalanceEdge `json:"edges,omitempty"`
+		Edges    []*BalanceEdge `json:"edges"`
 		PageInfo *PageInfo      `json:"pageInfo"`
 	}
 
 	var temp tempConnection
 	if err := json.Unmarshal(data, &temp); err != nil {
 		return fmt.Errorf("unmarshaling balance connection: %w", err)
+	}
+
+	if temp.Edges == nil {
+		return fmt.Errorf("balance connection missing required edges field: the GraphQL schema declares edges as non-null")
 	}
 
 	for i, edge := range temp.Edges {
--- a/pkg/wbclient/types/types.go
+++ b/pkg/wbclient/types/types.go
@@ -319,8 +319,7 @@ func (e *BalanceEdge) UnmarshalJSON(data []byte) error {
 	e.Cursor = temp.Cursor
 
 	if len(temp.Node) == 0 || string(temp.Node) == "null" {
-		e.Node = nil
-		return nil
+		return fmt.Errorf("balance edge missing required node (cursor=%q): the GraphQL schema declares Balance as non-null", temp.Cursor)
 	}
 
 	node, err := UnmarshalBalance(temp.Node)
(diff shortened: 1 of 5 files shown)


aristidesstaffieri:
### Code review

pkg/wbclient/types/types.go
```diff
@@ -295,3 +295,96 @@ type StateChangeConnection struct {
...
+// to reject null entries in the edges array. The GraphQL schema declares
+// edges as [BalanceEdge!]!, so a null edge is a server bug; failing fast
+// at decode time surfaces it instead of silently dropping data downstream.
+//
+// Null nodes inside an edge are caught by BalanceEdge.UnmarshalJSON;
+// this method covers the array-element case that the standard library
+// never routes through that decoder.
+func (c *BalanceConnection) UnmarshalJSON(data []byte) error {
+	type tempConnection struct {
+		Edges    []*BalanceEdge `json:"edges,omitempty"`
```
aristidesstaffieri: should this accept null edges? The schema declares this not to be nullable.

aditya1702: @aristidesstaffieri This is explicitly handled in unmarshling a balance edge: https://github.com/stellar/wallet-backend/pull/609/changes/BASE..ff476519d641f9a818300713197af92d864b0118#diff-c6552a6959bb5e200135cc74e3513f65fe75828bdafe3f93695591aa079e12ceR322

where I return an error if the balance edge node is null

aristidesstaffieri: this only handles a missing node inside a single edge, not a missing edges field.

aditya1702: @aristidesstaffieri Ah got it now. Good catch, fixed it here: https://github.com/stellar/wallet-backend/pull/609/commits/47085e082135081f0de8f545b4340561d24d30d2

aristidesstaffieri (approved):

PR #612, opened 2026-05-08
Classify account-balance response failures by category in SDK

aditya1702:
## Summary

Follow-up to #609

The previously-merged client treated three semantically different response failures as one undifferentiated error: account-not-found, server schema violations, and pagination bugs. Without separation, downstream consumers (freighter-backend) cannot route them appropriately — every failure looks systemic, including normal "user typed an invalid address" traffic.

This PR introduces a sentinel error and tightens schema validation so callers can classify each category cleanly:

| Category | Trigger | New behavior |
|---|---|---|
| **Address-scoped** | `accountByAddress: null` (account not indexed) | Returns `wbclient.ErrAccountNotFound` wrapped with the address. Classify with `errors.Is(err, wbclient.ErrAccountNotFound)`. |
| **Systemic — schema violation** | missing/null `balances`, `pageInfo`, or `edges` field | Descriptive decode-time or method-level error referencing the missing field. |
| **Systemic — pagination bug** (already in #611) | `HasNextPage=true` with no `EndCursor`, repeated cursor | Already handled in the prior follow-up. |

### `pkg/wbclient/client.go`

- **New exported sentinel** `ErrAccountNotFound = errors.New("account not found")`. Documented for `errors.Is` use by callers.
- **`AccountBalancesData.AccountByAddress` is now `*struct{...}`** (pointer to anonymous struct). Without the pointer, `json.Unmarshal` cannot distinguish `null` (server says account doesn't exist) from `{}` (server returned an account object with all fields nil) — both decode to the same zero-value inline struct. The pointer makes `null` observable as a nil pointer.
- **`GetAccountBalances` now returns**:
  - `ErrAccountNotFound` (wrapped with the address) when `accountByAddress` is null.
  - `"account balances response missing required balances field for address %s"` when the inner `balances` is nil despite a non-null account.
- **`GetAllAccountBalances` loop adds two defensive belt-and-suspenders checks** for nil connection and nil `PageInfo` from a successful primitive call. After the strict `UnmarshalJSON` work in #609 these conditions are unreachable from the JSON path, but enforcing them at the loop layer protects against future decoder-loosening regressions. The combined `PageInfo == nil || !HasNextPage` check is split so "schema violation" (loud error) and "normal end-of-stream" (silent break) have separate code paths.

### `pkg/wbclient/types/types.go`

- **`BalanceConnection.UnmarshalJSON` now rejects missing/null `pageInfo`** — the schema declares `pageInfo: PageInfo!` non-null. The strict-decode coverage now matches all three non-null guarantees the schema makes on a `BalanceConnection` (`edges` field, each edge element, and `pageInfo`).

### `pkg/wbclient/types/types_test.go`

- **Two new pin tests** for the strict `pageInfo`: rejects missing field, rejects `pageInfo: null`. Existing happy-path / null-edges tests still pass; the empty-edges-array case (`edges: []`) remains valid.

## Caller migration note

Consumers like freighter-backend should now route errors:

```go
balances, err := client.GetAllAccountBalances(ctx, address)
if errors.Is(err, wbclient.ErrAccountNotFound) {
    // address-scoped: render "account not found" UI
} else if err != nil {
    // systemic: log/alert, retry policy, etc.
}
```

## Test plan

- [x] `go build ./...`
- [x] `go vet ./pkg/wbclient/... ./internal/integrationtests/...`
- [x] `go test -race ./pkg/wbclient/... -timeout 2m` — 16 types tests pass (14 prior + 2 new pageInfo strict tests)
- [ ] Integration tests (run before merge):
  ```bash
  ENABLE_INTEGRATION_TESTS=true go test -v -race \
    ./internal/integrationtests/ \
    -run 'TestIntegrationTests/AccountBalancesAfterCheckpointTestSuite' \
    -timeout 30m
  ```
- [ ] `make check`

## Note on commit history

This branch was the source for the now-merged #609 and #611, so its commit list shows pre-merge individual commits alongside the merge commits that brought in the squash results. The actual net diff against `main` is only the most recent commit (52 insertions, 5 deletions across 3 files) — see the **Files changed** tab for the substantive review surface.

### DIFF
--- a/internal/integrationtests/account_balances_test.go
+++ b/internal/integrationtests/account_balances_test.go
@@ -176,11 +176,11 @@ func (suite *AccountBalancesAfterCheckpointTestSuite) TestCheckpoint_HolderContr
 	}
 }
 
-// TestCheckpoint_Account1_ForwardPagination exercises the SDK's forward
-// pagination round-trip. Pre-fix the SDK ignored caller-supplied first/after
-// args (issue #608); this test guards against any regression that drops
-// them again. Account1 has 3 balances, so first=1 plus a follow-up call
-// using the returned EndCursor must return a different edge.
+// TestCheckpoint_Account1_ForwardPagination verifies that GetAccountBalances
+// honors the first and after pagination args. Account1 has 3 balances at
+// checkpoint, so calling with first=1 must return a single edge with
+// HasNextPage=true; passing the returned EndCursor as after on the next
+// call must return a different edge.
 func (suite *AccountBalancesAfterCheckpointTestSuite) TestCheckpoint_Account1_ForwardPagination() {
 	pageSize := int32(1)
 	address := suite.testEnv.BalanceTestAccount1KP.Address()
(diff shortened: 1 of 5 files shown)


aristidesstaffieri:
### Code review

aristidesstaffieri (approved):

PR #615, opened 2026-05-12
fix: correct swapped JSON tags on PaginationLinks Prev/Next fields

### LINKED ISSUE #593: PaginationLinks: Prev and Next JSON tags are swapped
## Summary

`internal/entities/pagination.go` declares `PaginationLinks` with its `Prev` and `Next` field JSON tags swapped relative to the Go field names:

```go
type PaginationLinks struct {
    Self string `json:"self"`
    Prev string `json:"next"`
    Next string `json:"prev"`
}
```

## Impact

Any code that populates `PaginationLinks` by Go field name emits cursors under the wrong JSON key. Clients that look for `_links.next` will find the previous-page URL, and vice versa. Symmetrically-wrong tests and clients mask the bug.

## Expected

```go
type PaginationLinks struct {
    Self string `json:"self"`
    Prev string `json:"prev"`
    Next string `json:"next"`
}
```

## Notes

Changing the JSON tags is a wire-format fix and may break clients currently relying on the (wrong) behavior. Coordinating with client callers is advisable.

`internal/entities/pagination.go:3-7`

SAY-5:
Fixes #593.

`internal/entities/pagination.go` had the `Prev` and `Next` field tags
swapped, so any code that populated `PaginationLinks` by Go field name
emitted cursors under the wrong JSON key. The struct is currently unused
in this repository, so swapping the tags to match the field names has no
wire-format impact today.

Added a round-trip JSON test in `internal/entities/pagination_test.go`
that fails on the previous (swapped) tags and passes on the fix.

### DIFF
--- a/internal/entities/pagination.go
+++ b/internal/entities/pagination.go
@@ -2,8 +2,8 @@ package entities
 
 type PaginationLinks struct {
 	Self string `json:"self"`
-	Prev string `json:"next"`
-	Next string `json:"prev"`
+	Prev string `json:"prev"`
+	Next string `json:"next"`
 }
 
 type Pagination struct {
--- a/internal/entities/pagination_test.go
+++ b/internal/entities/pagination_test.go
@@ -1,0 +1,34 @@
+package entities
+
+import (
+	"encoding/json"
+	"testing"
+)
+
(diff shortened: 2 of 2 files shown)


aristidesstaffieri (approved):
@SAY-5 thanks for the contribution!

PR #616, opened 2026-05-13
Mirror Balance nil-handling pattern in account-scoped transaction/operation/state-change queries

aditya1702:
## Summary

Brings `GetAccountTransactions`, `GetAccountOperations`, and `GetAccountStateChanges` in `pkg/wbclient` in line with the existing `GetAccountBalances` behavior so that SDK consumers can distinguish *"account does not exist"* from *"account exists but has no items"*.

Today, a `(nil, nil)` return from any of these three methods is ambiguous. After this change, an absent account produces a wrapped `ErrAccountNotFound` that callers can match with `errors.Is`.

The change is split across two commits:

1. **Type-layer validators** — adds custom `UnmarshalJSON` to `TransactionEdge`, `OperationEdge`, `TransactionConnection`, `OperationConnection`, and `StateChangeConnection` (plus a clarifying comment on the pre-existing `StateChangeEdge.UnmarshalJSON`). These reject schema-violating server responses — null entries inside the edges array, and missing or null `pageInfo` — and accept everything the schema permits.
2. **Client-layer ErrAccountNotFound** — reshapes the three response data structs so `AccountByAddress` is a pointer, and returns `ErrAccountNotFound` (wrapped with the queried address) when the field comes back null.

## Schema asymmetry — why this isn't a copy of `BalanceConnection.UnmarshalJSON`

The GraphQL schema's non-null declarations differ across the four connection types, so the new validators are looser than `BalanceConnection.UnmarshalJSON`:

| Type | `edges` field | edge entries | `node` | parent `Account.X` field |
|------|---------------|--------------|--------|--------------------------|
| Balance | `[BalanceEdge!]!` — non-null | non-null | `Balance!` — non-null | non-null |
| Transaction | `[TransactionEdge!]` — **nullable** | non-null | `Transaction` — **nullable** | **nullable** |
| Operation | `[OperationEdge!]` — **nullable** | non-null | `Operation` — **nullable** | **nullable** |
| StateChange | `[StateChangeEdge!]` — **nullable** | non-null | `BaseStateChange` — **nullable** | **nullable** |

This means:
- For these three types, a null `edges` field, null `node`, or null parent connection field is **schema-valid** and passes through silently.
- Only null entries inside the edges array and null `pageInfo` are rejected (the schema's only non-null declarations).
- Callers iterating `connection.Edges` must nil-check `edge.Node` for these three types — unlike `BalanceConnection` where `edge.Node` is guaranteed non-nil after unmarshal.

## What is intentionally *not* changed

- No new `GetAllAccount{Transactions,Operations,StateChanges}` paginating helpers.
- No new flat-slice helpers like `BalanceConnection.Balances()` for these three types.
- No schema tightening — the `.graphqls` files are untouched.
- `BalanceConnection` / `BalanceEdge` / `GetAccountBalances` / `GetAllAccountBalances` are unchanged.

## Test plan

- [x] Type-layer: 27 new unit tests in `pkg/wbclient/types/types_test.go` covering edge null-node accept paths, connection null-edge reject path, connection null/missing pageInfo reject path, end-to-end decode.
- [x] Client-layer: 10 new tests in `pkg/wbclient/client_test.go` using `httptest.Server` to cover `ErrAccountNotFound`, null-connection passthrough, and empty-edges connection for each of the three methods.
- [x] Full `go test -race ./pkg/wbclient/...` passes (105 tests across 3 packages).
- [x] `go vet ./pkg/wbclient/...` passes.
- [x] `go build ./pkg/wbclient/...` passes.

### DIFF
--- a/pkg/wbclient/client.go
+++ b/pkg/wbclient/client.go
@@ -63,19 +63,19 @@ type StateChangesData struct {
 }
 
 type AccountTransactionsData struct {
-	AccountByAddress struct {
+	AccountByAddress *struct {
 		Transactions *types.TransactionConnection `json:"transactions"`
 	} `json:"accountByAddress"`
 }
 
 type AccountOperationsData struct {
-	AccountByAddress struct {
+	AccountByAddress *struct {
 		Operations *types.OperationConnection `json:"operations"`
 	} `json:"accountByAddress"`
 }
 
 type AccountStateChangesData struct {
-	AccountByAddress struct {
+	AccountByAddress *struct {
 		StateChanges *types.StateChangeConnection `json:"stateChanges"`
 	} `json:"accountByAddress"`
 }
@@ -376,6 +376,10 @@ func (c *Client) GetAccountTransactions(ctx context.Context, address string, sin
 		return nil, err
 	}
 
+	if data.AccountByAddress == nil {
+		return nil, fmt.Errorf("%w: %s", ErrAccountNotFound, address)
+	}
+
 	return data.AccountByAddress.Transactions, nil
 }
 
@@ -404,6 +408,10 @@ func (c *Client) GetAccountOperations(ctx context.Context, address string, since
 	data, err := executeGraphQL[AccountOperationsData](c, ctx, buildAccountOperationsQuery(fields), variables)
 	if err != nil {
 		return nil, err
+	}
+
+	if data.AccountByAddress == nil {
+		return nil, fmt.Errorf("%w: %s", ErrAccountNotFound, address)
 	}
 
 	return data.AccountByAddress.Operations, nil
@@ -449,6 +457,10 @@ func (c *Client) GetAccountStateChanges(ctx context.Context, address string, tra
 	data, err := executeGraphQL[AccountStateChangesData](c, ctx, buildAccountStateChangesQuery(), variables)
 	if err != nil {
 		return nil, err
+	}
+
+	if data.AccountByAddress == nil {
+		return nil, fmt.Errorf("%w: %s", ErrAccountNotFound, address)
 	}
 
 	return data.AccountByAddress.StateChanges, nil
--- a/pkg/wbclient/client_test.go
+++ b/pkg/wbclient/client_test.go
@@ -26,128 +26,137 @@ func graphqlServer(t *testing.T, dataJSON string) *httptest.Server {
 	}))
 }
 
-func TestGetAccountTransactions_AccountNotFound(t *testing.T) {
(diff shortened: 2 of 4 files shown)


aristidesstaffieri:
### Code review

Found 2 minor issues:

1. Test asymmetry: `TestGetAccountTransactions` has a "sends well-formed GraphQL request body" subtest, but the parallel `TestGetAccountOperations` and `TestGetAccountStateChanges` don't — leaving request-body / variable-marshaling coverage uneven across the three new methods.

https://github.com/stellar/wallet-backend/blob/833411fa21cb90b9b1a84ff7c20191677449c864/pkg/wbclient/client_test.go#L64-L90

2. The \`json:"node,omitempty"\` tag on the \`tempEdge.Node json.RawMessage\` field is a no-op during unmarshal — \`omitempty\` only affects marshaling. Functionally harmless because the explicit \`len(temp.Node) == 0 || string(temp.Node) == "null"\` check handles both missing and null nodes, but the tag could mislead readers. Same pattern repeats at lines 321 and 400 (and is pre-existing on \`BalanceEdge\` at 478).

https://github.com/stellar/wallet-backend/blob/833411fa21cb90b9b1a84ff7c20191677449c864/pkg/wbclient/types/types.go#L243-L258

aditya1702:
Thanks @aristidesstaffieri — both points validated and fixed in b5896031.

**1. Test asymmetry — fixed.** Added `"sends well-formed GraphQL request body"` subtests to `TestGetAccountOperations` and `TestGetAccountStateChanges`. The StateChanges variant additionally exercises the `filter` object encoding (transactionHash, category) since that's the unique part of `GetAccountStateChanges` not covered by the Transactions/Operations request-body tests — gives genuine differential coverage rather than three near-identical copies.

**2. `omitempty` is a no-op on unmarshal — confirmed and fixed.** You're right that `omitempty` only affects marshaling (the encoder skips zero-valued fields when *writing* JSON); on a struct used purely as an unmarshal target it's dead weight. Behavior is unchanged here because the explicit `len(temp.Node) == 0 || string(temp.Node) == "null"` check is what enforces nullable-node semantics — but the misleading tag is gone now.

I cleaned up all four occurrences in the same commit (TransactionEdge, OperationEdge, StateChangeEdge, BalanceEdge) so the pattern stays consistent across the file. The pre-existing `BalanceEdge` case was a copy-paste source for the new ones, so leaving it would have re-seeded the same mistake later.

All 115 tests pass.

aristidesstaffieri (approved):

PR #620, opened 2026-06-02
feat: nest account-scoped operations & state changes on Account.transactions

### LINKED ISSUE #619: Add new methods to the Go client according to Freighter's history redesign needs
Freighter will use wallet backend's endpoints to power its new redesigned history view. Since wallet backend exposes a graphql API, it is flexible to query the txns/ops/state changes in different ways. We want the client to expose methods to do the following:

- Get an account's list of transactions
- For each of these transactions, get all state changes for that account
- For each of these transactions, get all operations for that account

aditya1702:
Closes #619 

## TL;DR

- **What:** `accountByAddress.transactions` now embeds each transaction's account-scoped `operations` and `stateChanges` on the edge — a full account history in one GraphQL call instead of N follow-up requests.
- **Account-scoped:** each edge shows only the operations/state changes involving the queried account, not everything in the transaction.
- **No N+1:** the new fields resolve lazily and batch across the whole page via dataloaders; the queries pin the TimescaleDB partition column for chunk exclusion and filter strictly by account.
- **SDK:** new `GetAccountTransactionsWithOpsAndStateChanges` method + response types, with strict (non-null-enforcing) JSON decoding.
- **Page-size cap:** `account.transactions/operations/stateChanges` now cap `first`/`last` at **100** (returning `BAD_USER_INPUT` above that), mirroring `account.balances` — the nested per-transaction fan-out makes unbounded pages expensive.
- **Backward-compatible:** `node`, `cursor`, and `pageInfo` are unchanged; callers that don't request the new fields are unaffected.
- **Correctness fix:** `BatchGetByAccountAddress` now always selects `ledger_created_at` (the partition key); without it, a query omitting `ledgerCreatedAt` on the node would have *silently* returned empty nested lists. Guarded by a regression test.

---

## Summary (layer by layer)

- **Data layer** — `BatchGetAccountOperationsByToIDs` / `BatchGetAccountStateChangesByToIDs`: per-`to_id` `LATERAL` queries that pin `ledger_created_at` (the TimescaleDB partition column) for chunk exclusion, scope by `account_id`, and return the complete per-transaction set (no `LIMIT`), ordered DESC. Operations have no account column, so they're scoped through `operations_accounts` via the TOID band; state changes scope directly via their `account_id` column.
- **Dataloaders** — `accountOperationsByToIDLoader` / `accountStateChangesByToIDLoader` wrap those methods, batching all of a page's transactions into one query keyed by transaction `to_id`.
- **Schema + resolvers** — `Account.transactions` now returns `AccountTransactionConnection`; its `AccountTransactionEdge` carries `operations` / `stateChanges` (both `@goField(forceResolver: true)` so they resolve lazily through the loaders). The Go backing struct threads the account address from the parent `Account` resolver down to the edge resolvers — it scopes the lookups but is not part of the public schema. Page size on `transactions`/`operations`/`stateChanges` is capped at 100 via `parseAccountPaginationParams`.
- **SDK** — `GetAccountTransactionsWithOpsAndStateChanges` issues the nested query; new connection/edge types whose `UnmarshalJSON` enforces the schema's non-null contract (`edges`, `node`, `pageInfo`) and dispatches polymorphic state-change nodes by `__typename`.
- **Partition-pin fix** — `BatchGetByAccountAddress` always selects `ledger_created_at` so the nested edge resolvers receive a real ledger time to pin (see TL;DR).

## Test Plan

- [x] `make check` clean — fmt, vet, lint, shadow, exhaustive, deadcode, goimports, govulncheck (+ `tidy` applied)
- [x] `go test ./internal/data/... ./internal/serve/graphql/... ./pkg/wbclient/...` → 499 passed
- [x] Cross-account exclusion verified at the data layer **and** the resolver layer, for both operations and state changes
- [x] Partition-pin regression test: nested operations resolve correctly when `ledgerCreatedAt` is omitted from the node selection (load-bearing — fails if the fix is reverted)
- [x] Page-size cap regression test: `parseAccountPaginationParams` rejects `first`/`last` > 100 with `BAD_USER_INPUT`
- [ ] Integration tests deferred per the design spec (follow-up issue)

> Cross-repo ordering (per spec): this must merge and publish a tagged SDK before the freighter-backend dependency bump.


pkg/wbclient/types/types.go
```diff
@@ -545,6 +545,90 @@ func (c *BalanceConnection) UnmarshalJSON(data []byte) error {
...
+	}
+	var temp tempEdge
+	if err := json.Unmarshal(dataBytes, &temp); err != nil {
+		return fmt.Errorf("unmarshaling detailed transaction edge: %w", err)
+	}
+	e.Cursor = temp.Cursor
+	e.Operations = temp.Operations
+	if len(temp.Node) == 0 || string(temp.Node) == "null" {
+		return fmt.Errorf("detailed transaction edge missing required node: the GraphQL schema declares Transaction as non-null")
+	}
```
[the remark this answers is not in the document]
aditya1702: Fixed in c5e03588 — `DetailedTransactionEdge.UnmarshalJSON` now rejects a null/missing `operations` or `stateChanges` list (and null elements), matching the schema (`[Operation!]!` / `[BaseStateChange!]!`) and the existing node/edges/pageInfo enforcement. Added SDK tests for both null-list cases.

internal/serve/graphql/dataloaders/operation_loaders.go
```diff
@@ -71,6 +73,29 @@ func operationsByToIDLoader(models *data.Models) *dataloadgen.Loader[OperationCo
...
+			accountID := keys[0].AccountID
+			columns := keys[0].Columns
+			toIDs := make([]int64, len(keys))
+			ledgerCreatedAts := make([]time.Time, len(keys))
+			for i, key := range keys {
+				toIDs[i] = key.ToID
+				ledgerCreatedAts[i] = key.LedgerCreatedAt
+			}
+			return models.Operations.BatchGetAccountOperationsByToIDs(ctx, accountID, toIDs, ledgerCreatedAts, columns)
+		},
```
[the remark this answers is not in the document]
aditya1702: Fixed in c5e03588 — account-scoped loaders now group the batch by account (`newAccountScopedLoader`) and map results back per (account, toID), so an aliased multi-account request cannot cross-contaminate edges that share a transaction. Added a `LoadAll` regression test that fails if the batch is scoped to a single account.

internal/serve/graphql/dataloaders/statechange_loaders.go
```diff
@@ -103,3 +105,25 @@ func stateChangesByOperationIDLoader(models *data.Models) *dataloadgen.Loader[St
...
+			accountID := keys[0].AccountID
+			columns := keys[0].Columns
+			toIDs := make([]int64, len(keys))
+			ledgerCreatedAts := make([]time.Time, len(keys))
+			for i, key := range keys {
+				toIDs[i] = key.ToID
+				ledgerCreatedAts[i] = key.LedgerCreatedAt
+			}
+			return models.StateChanges.BatchGetAccountStateChangesByToIDs(ctx, accountID, toIDs, ledgerCreatedAts, columns)
+		},
```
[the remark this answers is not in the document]
aditya1702: Fixed in c5e03588 — same per-account grouping as the operations loader (`newAccountScopedLoader`); a shared-transaction multi-account batch no longer leaks. Covered by the same `LoadAll` no-leak regression test.

internal/data/operations.go
```diff
@@ -210,6 +210,45 @@ func (m *OperationModel) BatchGetByToID(ctx context.Context, toID int64, columns
...
 }
 
+// BatchGetAccountOperationsByToIDs returns, for each (toID, ledgerCreatedAt) pair, all operations
+// in that transaction involving accountAddress. ledgerCreatedAt is the parent transaction's ledger
+// time (== the operation's ledger time); pinning the partition column triggers primary chunk
+// exclusion. No LIMIT: operations per tx are ≤100 by Stellar protocol. operations has no account
+// column, so scope through operations_accounts (segmentby=account_id) via the TOID band, then a PK
+// LATERAL into operations.
+func (m *OperationModel) BatchGetAccountOperationsByToIDs(ctx context.Context, accountAddress string, toIDs []int64, ledgerCreatedAts []time.Time, columns string) ([]*types.Operation, error) {
+	columns = prepareColumnsWithID(columns, types.Operation{}, "o", "id")
```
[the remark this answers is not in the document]
aditya1702: Fixed in c5e03588 — added a length-equality guard on `toIDs`/`ledgerCreatedAts` that returns a clear error instead of silently mis-pairing through UNNEST.

internal/data/statechanges.go
```diff
@@ -485,6 +485,38 @@ func (m *StateChangeModel) BatchGetByOperationID(ctx context.Context, operationI
 	return stateChanges, nil
 }
 
+// BatchGetAccountStateChangesByToIDs returns, for each (toID, ledgerCreatedAt) pair, all state
+// changes in that transaction that belong to accountAddress. ledgerCreatedAt is the parent
+// transaction's ledger time (== the state change's ledger time, same ledger); pinning the
+// partition column triggers primary chunk exclusion. No LIMIT: an account's per-tx state
+// changes are bounded in practice. Account scoping is via the account_id column on state_changes.
+func (m *StateChangeModel) BatchGetAccountStateChangesByToIDs(ctx context.Context, accountAddress string, toIDs []int64, ledgerCreatedAts []time.Time, columns string) ([]*types.StateChange, error) {
+	columns = prepareColumnsWithID(columns, types.StateChange{}, "sc", "to_id")
```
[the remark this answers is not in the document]
aditya1702: Fixed in c5e03588 — added the same parallel-array length guard as the operations method.

internal/serve/graphql/dataloaders/loaders.go
```diff
@@ -124,6 +134,65 @@ func newOneToManyLoader[K comparable, PK comparable, V any, T any](
...
+				toIDs := make([]int64, len(indices))
+				ledgerCreatedAts := make([]time.Time, len(indices))
+				for j, i := range indices {
+					toIDs[j] = toID(keys[i])
+					ledgerCreatedAts[j] = ledgerCreatedAt(keys[i])
+				}
+
+				// All keys for one account in one connection request the same columns, so reading
+				// columns from the group's first key is safe.
+				items, err := fetch(ctx, account, toIDs, ledgerCreatedAts, columns(keys[indices[0]]))
```
aristidesstaffieri: **Account-scoped loader collapses differing column selections within one account → silent missing fields**

`newAccountScopedLoader` groups a batch by account and issues one `fetch` per account using only the **first** key's `Columns` (`columns(keys[indices[0]])`). The comment above asserts "All keys for one account in one connection request the same columns" — but a single request can alias the same account across *different* connections with different nested sub-selections:

```graphql
{
  a: accountByAddress(address:"G…"){ transactions{ edges{ operations{ id } } } }
  b: accountByAddress(address:"G…"){ transactions{ edges{ operations{ id operationType } } } }
}
```

`GetDBColumnsForFields` returns `{id}` for alias `a` and `{id, operationType}` for alias `b`, so these become two distinct keys (same `AccountID`/`ToID`, different `Columns`) in one dataloader batch. The single fetch runs with whichever key is first, and rows map back to **both** keys by `to_id` — so the alias requesting extra columns silently receives zero-valued fields (`operationType` empty) with no error. Affects both the operations and state-changes loaders.

internal/serve/serve.go
```diff
@@ -115,6 +119,24 @@ func Serve(cfg Configs) error {
...
+		registerAdminHandlers(adminMux)
+		adminServer := &http.Server{
+			Addr:              fmt.Sprintf(":%d", cfg.AdminPort),
+			Handler:           adminMux,
+			ReadHeaderTimeout: 5 * time.Second,
+		}
+		go func() {
+			log.Infof("Starting admin server with pprof endpoints on port %d", cfg.AdminPort)
+			if err := adminServer.ListenAndServe(); err != http.ErrServerClosed {
+				log.Fatalf("starting admin server on %s: %v", adminServer.Addr, err)
```
aristidesstaffieri: **`log.Fatalf` in the admin-server goroutine kills the whole serve process on bind failure**

If `AdminPort` is already bound (port reuse during a pod restart, lingering `TIME_WAIT`, or a misconfig), `ListenAndServe` returns `bind: address already in use`, this `log.Fatalf` calls `os.Exit`, and the entire serve process goes down — even though the pprof admin server is purely optional profiling infrastructure and the main API server would otherwise be healthy.

Consider logging the error and continuing (degrade to "no pprof") rather than fataling, so an optional debug feature can't take down request serving.

internal/data/statechanges.go
```diff
@@ -495,6 +495,60 @@ func (m *StateChangeModel) BatchGetByOperationID(ctx context.Context, operationI
...
+// account_id, orderby ledger_created_at); uncompressed chunks can use index range scans;
+// chunks outside the range are excluded entirely.
+func (m *StateChangeModel) BatchGetAccountStateChangesByToIDs(ctx context.Context, accountAddress string, toIDs []int64, ledgerCreatedAts []time.Time, columns string) ([]*types.StateChange, error) {
+	if len(toIDs) != len(ledgerCreatedAts) {
+		return nil, fmt.Errorf("toIDs and ledgerCreatedAts must be parallel arrays of equal length, got %d and %d", len(toIDs), len(ledgerCreatedAts))
+	}
+	if len(toIDs) == 0 {
+		return []*types.StateChange{}, nil
+	}
+	lo, hi := ledgerCreatedAts[0], ledgerCreatedAts[0]
```
aristidesstaffieri, line 519: **Prune window collapses if any node carries a zero `ledger_created_at`**

`lo`/`hi` are derived directly from the caller-supplied `ledgerCreatedAts` with no guard against a zero `time.Time`. A single zero entry makes `lo = 0001-01-01`, turning the `ledger_created_at >= $3` range filter into a full-account-history scan — defeating the chunk-exclusion this function's doc comment promises.

The `transactions.go` change in this PR keeps `Node.LedgerCreatedAt` populated today, so this isn't a live bug — but the function's contract doesn't enforce it, and a future caller could silently regress into the whole-history-decompress cliff. (The operations band's exact `oa.ledger_created_at = i.ledger_created_at` equality has the same implicit dependency.) Worth a cheap guard or at least a documented precondition.

pkg/wbclient/types/types.go
```diff
@@ -545,6 +545,109 @@ func (c *BalanceConnection) UnmarshalJSON(data []byte) error {
...
+
+// UnmarshalJSON enforces the schema's non-null guarantees, matching BalanceConnection. The
+// schema declares edges as [AccountTransactionEdge!]! and pageInfo as PageInfo!, so each of
+// the following is a server bug and is rejected here:
+//   - a missing or null edges field on the connection
+//   - a null entry within the edges array
+//   - a missing or null pageInfo field on the connection
+//
+// Null nodes inside an edge object are caught separately by AccountTransactionEdge.UnmarshalJSON.
+func (c *AccountTransactionConnection) UnmarshalJSON(dataBytes []byte) error {
```
aristidesstaffieri, line 587: **Cleanup: five near-identical strict-null `UnmarshalJSON` connection decoders**

`AccountTransactionConnection.UnmarshalJSON` (and `AccountTransactionEdge.UnmarshalJSON` just above) duplicate, nearly verbatim, the temp-struct + edges-nil + per-edge-nil + pageInfo-nil pattern already implemented in `BalanceConnection`, `TransactionConnection`, `OperationConnection`, and `StateChangeConnection`, plus the node-null-guard and `string(raw) == "null"` element check from the edge types.

The only real axis of variation is whether nil `edges` is rejected. A generic helper parameterized over edge type and an "edges required" flag would collapse the five copies. As-is, any change to the null-handling contract (e.g. an error-message format a test/client greps, or handling a new state-change `__typename`) must be hand-applied to all five — and missing one yields a connection type that silently accepts payloads the others reject.

Non-blocking cleanup, not a correctness issue.

aristidesstaffieri:
## ⚠️ Security: this PR is exposed to a variant of the existing GraphQL N+1 / complexity-undercount DoS

A DoS was reported against the existing API (unpaginated list fields score complexity 1 → undercount; `Account.balances` has no dataloader → N+1). **This PR is vulnerable to a close variant, and in one respect makes the attack easier.** Confirmed against the branch code, not reasoned in the abstract.

### Both original flaws still exist here
- **N+1 unchanged:** `Account.balances` → `getAccountBalances` directly, no dataloader — still 2 sequential DB queries per G-address account. This PR doesn't touch it.
- **Complexity undercount extended:** in `addComplexityCalculation`, complexity functions are registered only for paginated connection fields. Every unpaginated `[…!]!` list field gets the default multiplier of **1** — including the two this PR adds.

### What this PR adds
`AccountTransactionEdge` introduces two unpaginated list fields:

```graphql
operations:   [Operation!]!        @goField(forceResolver: true)
stateChanges: [BaseStateChange!]!  @goField(forceResolver: true)
```

Neither is registered in `addComplexityCalculation`, so **both score multiplier 1** — the same shape as `Transaction.accounts` in the original report. The fields themselves are correctly batched by the new account-scoped dataloaders, but the **objects they return** open a new, cheap traversal path into the pre-existing `Account.balances` N+1:

```graphql
{
  accountByAddress(address: "G_busy_account") {
    transactions(first: 100) {        # capped at 100 by this PR
      edges {
        operations {                  # NEW field: multiplier 1, UNCAPPED (SQL LIMIT 4096 per tx)
          accounts {                  # Operation.accounts: multiplier 1, dataloader-batched fetch...
            balances(first: 1) {      # ...but Account.balances has NO dataloader -> 2 seq queries each
              edges { node { __typename } }
            }
          }
        }
      }
    }
  }
}
```

**Complexity ≈ 601** (`100 × (1 + operations[1] + accounts[1] + balances[3])`) — far under the 5000 limit — while at runtime `Account.balances` fires for every Account node under every operation across all 100 transactions: `100 × ops/tx × accounts/op × 2 sequential queries`, i.e. thousands of pool-blocking queries from one ~600-complexity request.

### Why the new field is *worse* than the pre-existing path
The equivalent old path `node { operations(first: N) { … } }` goes through `Transaction.operations`, a **paginated connection that IS registered** (`config.Complexity.Transaction.Operations`) — it multiplies by `N` and is pagination-bounded. The new `AccountTransactionEdge.operations` is a **raw list with no pagination args and no complexity function**: costs 1 in the model but returns the account's entire operation set for the transaction (`LIMIT 4096`), uncapped. Cheaper, larger, uncapped.

There is also a **standalone amplification**: `operations` alone can return up to `100 × 4095 ≈ 400k` objects for complexity ~1 — response/CPU amplification independent of the pool issue.

### What this PR's page cap does and doesn't do
`parseAccountPaginationParams` (cap 100) is a genuine improvement and bounds the *outer* page. It does **not**: cap the nested `operations`/`stateChanges` lists (no args to cap), register complexity for the new fields, or fix the `Account.balances` N+1. (The cap is also enforced in the resolver, while complexity is checked earlier — so it can't be the complexity defense.)

### Recommendations
1. **Register complexity** for the new fields and the still-unregistered `Transaction.accounts` / `Operation.accounts`, with a fixed multiplier reflecting realistic max fan-out instead of 1.
2. **Cap the nested lists** so a single transaction can't return up to 4095 operations uncounted.
3. **Fix the root cause:** batch `Account.balances` (and the unbatched `Account.transactions/operations/stateChanges`) behind a dataloader, and/or add max-depth / total-node limits — complexity tuning alone is fragile.
4. Default-open auth (`CLIENT_AUTH_PUBLIC_KEYS` optional) means this is reachable unauthenticated — worth addressing alongside.

internal/data/statechanges.go
aditya1702: @aristidesstaffieri The `ledger_created_at` values come from the transactions stored in our db - this will always be a non-zero value. I think we can safely ignore this. Wdyt?

aristidesstaffieri: ah yes you're right I'll resolve this one

internal/serve/graphql/dataloaders/loaders.go
aditya1702: Fixed in f09bfba — the batch now groups by `(account, columns)` instead of account alone, so an aliased same-account request with differing sub-selections fetches each column set on its own. Added a `LoadAll` regression test (lean-columns key first) that fails pre-fix.

https://github.com/stellar/wallet-backend/pull/620/commits/f09bfba5eb2e53c1d676ce3f2d8d49341a5d8d10

internal/serve/serve.go
aditya1702: Fixed in f09bfba — the admin/pprof goroutine now logs the bind error via `log.Errorf` and degrades to "no pprof" instead of `log.Fatalf` taking down the main API server.

https://github.com/stellar/wallet-backend/pull/620/commits/f09bfba5eb2e53c1d676ce3f2d8d49341a5d8d10

pkg/wbclient/types/types.go
aditya1702: Done in f09bfba — collapsed the five connection decoders into one generic `unmarshalConnection[E]` helper with a `requireEdges` flag, preserving the exact error strings. Edge decoders left as-is (distinct/polymorphic nodes).

https://github.com/stellar/wallet-backend/pull/620/commits/f09bfba5eb2e53c1d676ce3f2d8d49341a5d8d10

aditya1702:
Addressed rec #1 in f09bfba, scoped to the `accounts` N+1 vector:

- Registered ×50 (`DefaultPageLimit`) complexity on `Transaction.accounts` / `Operation.accounts`. The deep `operations { accounts { balances } }` path now costs ~15,200 and is rejected; the cost propagates up through the (unmultiplied) `operations` field, so every route into the N+1 inherits it.
- We are not capping the nested operations and state changes since the main point of this PR is to have that functionality so Freighter can get everything in 1 query. This should not be an issue in practice since operations are capped at 4097 and state changes should be minimal for an account per txn

Recs 3 (dataloader the balances N+1), and 4 (auth) are out of scope here. Tests in `complexity_test.go`.

https://github.com/stellar/wallet-backend/pull/620/commits/f09bfba5eb2e53c1d676ce3f2d8d49341a5d8d10

aristidesstaffieri (approved):

Issue #626, opened 2026-06-16
Fix state changes generation based on Freighter needs


aditya1702:
There seems to be some small bugs and incorrectness in the way we generate some specific state changes. We need to validate and fix all these issues for the upcoming Freighter history migration to wallet backend

piyalbasu:
Break this down into smaller tickets. More details to come from @aditya1702

PR #630, opened 2026-06-22
feat(graphql): expose deployer address on contract-deploy AccountChange state changes

### LINKED ISSUE #632: Expose `DeployerAddress` in the AccountChange state change graphql schema
Currently we set the deployer address when new contracts are created but dont expose it via the graphql API. Lets update the schema files and the generated files.

JiahuiWho:
Closes #632 

### What

Add a deployerAddress field to AccountChange:
- schema: deployerAddress on AccountChange
- resolver: resolves from DeployerAccountID
- getDBColumns: map deployerAddress -> deployer_account_id so the column is selected
- regenerate gqlgen
- wbclient: add deployerAddress to the AccountChange fragment + struct

### Why

For contract deployments, the deployer is stored but invisible through the API

### Known limitations

N/A

https://github.com/stellar/wallet-backend/issues/626

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/integrationtests/data_validation_test.go
+++ b/internal/integrationtests/data_validation_test.go
@@ -151,6 +151,17 @@ func validateAccountChange(suite *DataValidationTestSuite, ac *types.AccountChan
 	suite.Require().Equal(expectedFunderAddress, *ac.FunderAddress, "funder address mismatch")
 }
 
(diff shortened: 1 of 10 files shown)


aditya1702 (commented):
I am thinking we should also add a case for deployer ID in our integration tests. I dont think we are creating any state change for that case right now

aditya1702 (approved):

PR #641, opened 2026-06-25
fix(indexer): fold fee-phase changes into native balances

### LINKED ISSUE #637: Native XLM balance over-reported: ingestion misses transaction fee-phase balance changes
## Summary

`wallet-backend` over-reports the **native XLM balance** for accounts whose most recent balance movement was **fee-only** — e.g. a fee-bump fee source, or any transaction whose operations don't modify the source account's own ledger entry. The native-balance ingestion consumes only **per-operation** ledger changes and never the transaction **fee phase**, so fee-only balance updates are dropped and the stored balance goes stale (too high by the uncaptured fees).

Found via a v1↔v2 account-balance validation sweep ahead of the Freighter v1→v2 migration.

## Evidence (measured, pubnet)

Account `GDXPMBB4EV2IEFWZKR2QF3CEFHDUDQESRTEBM6OJFLN7LRFTVLW7TWJR`:

| Source | native balance |
|---|---|
| Horizon (`horizon.stellar.org`, on-chain) | `10.4979100` |
| wallet-backend balances | `10.5000000` |

Over-reported by `0.0020900` XLM. The account has `subentry_count=0`, no liabilities, and `minimumBalance` is computed correctly (`1.0`) — so this is **not** the base-reserve calc; it is the `balance` field itself. The account was almost certainly funded at a round `10.5` and has since paid `~0.00209` XLM in fees that were never applied to its stored balance.

## Root cause

`internal/indexer/processors/accounts.go` — `AccountsProcessor.ProcessOperation`:

```go
changes, err := opWrapper.Transaction.GetOperationChanges(opWrapper.Index)   // operation meta ONLY
...
accChange.Balance = int64(account.Balance)                                   // absolute balance
```

- `GetOperationChanges` (go-stellar-sdk `ingest/ledger_transaction.go`) returns only the **operation meta** changes. The fee debit is emitted in the separate **fee phase** (`GetFeeChanges()`), which this processor never reads.
- The only use of `FeeChanges` in the codebase is `internal/indexer/processors/participants.go` (participant extraction); it is **not** applied to `native_balances`.

So when a transaction changes an account's balance **only** in the fee phase (fee-bump fee account, or a tx whose operations are sourced by / act on other accounts), `AccountsProcessor` sees no account change for it and never updates `native_balances` — leaving the prior, pre-fee absolute balance.

The base-reserve / `minimumBalance` calculation in the same file is correct and is **not** the issue.

### Horizon ingests this correctly (for contrast)

Horizon's `AccountsProcessor` is a **change processor** fed by `ingest.NewLedgerChangeReader`, which starts in `feeChangesState` and emits `GetFeeChanges()` **before** the operation-meta changes. So Horizon processes the fee-phase account change (which carries the post-fee balance) and stores the correct absolute balance. wallet-backend's per-operation iteration omits that phase — which is why Horizon and wallet-backend read the *same* absolute `account.Balance` field yet disagree.

## Impact

Native XLM balances are over-reported (too high by accumulated uncaptured fees) for any account with fee-bump activity, or transactions whose operations don't touch the source's own entry. Classic trustline / SAC / SEP-41 balances are unaffected (they are not fee-debited).

## Suggested fix

Apply fee-phase balance changes to native balances — e.g. ingest account-balance changes from the **full** ledger change stream (fee changes + operation meta), as Horizon does via `LedgerChangeReader`, rather than only `GetOperationChanges`. (Alternatively, additionally fold `Transaction.GetFeeChanges()` into `AccountsProcessor`.)

## Repro

```bash
ACC=GDXPMBB4EV2IEFWZKR2QF3CEFHDUDQESRTEBM6OJFLN7LRFTVLW7TWJR
# on-chain truth
curl -s "https://horizon.stellar.org/accounts/$ACC" | jq '.balances[] | select(.asset_type=="native") | .balance'
# wallet-backend: query the same account's native balance and compare
```

Verified against wallet-backend module `v0.0.0-20260615202139` and go-stellar-sdk `ingest`. Cross-checked the asymmetry in wallet-backend, the SDK, and Horizon.

aditya1702:
Closes #637

Accounts whose XLM balance moves **only** in the fee phase (fee-bump fee sources, or txs whose operations don't touch the source's own entry) had stale, too-high balances — the per-operation meta loop never saw them. `AccountsProcessor.ProcessTransactionFees` now folds the fee debit (`GetFeeChanges`) and Soroban fee refund (`GetPostApplyFeeChanges`) into native balances.

Changes are ranked by Stellar's canonical close order **fee < operation < refund** using an explicit `(phase, tx, op)` sort key (`accountSortKey`) instead of a TOID. This makes the buffer's per-account "keep highest key" dedup a tie-free total order — correct regardless of push/merge order, with no dependence on tx-ordered merge or TOID field widths. The key is in-memory only and never persisted, so `AccountChange.OperationID` is renamed `SortKey`.

**Tests:** sort-key ordering plus multi-tx fee/fee and refund/refund tie cases (unit); a fee-bump fixture whose fee source is touched by no operation, asserting the indexed native balance reflects the fee debit (integration).

## Review follow-ups

- **Buffer no-op resurrection hardened.** The buffer's create/add→remove no-op previously `delete`d the key, which erased the highest-key guard and let a later lower-key change re-insert a stale phantom. Now tracked with a tombstone (records the cancellation's order value): a later change is dropped when its order is ≤ the tombstone and re-creates the key only when strictly higher. Applied to account, trustline, and SAC-balance dedup (push + merge) via shared generic `pushWithTombstone`/`mergeWithTombstone` helpers; merge propagates tombstones across buffers.
- **Refund ordering documented.** The fee-phase refund is read from `GetPostApplyFeeChanges`, the protocol-23+ post-tx-set-apply refund that Core applies after every operation in the ledger, so `phaseRefund` dominating operations is correct (not an inversion). Stated explicitly in the sort-key docs.
- **Backfill doc corrected.** The backfill buffer dedups account changes in memory but never persists them (only the live per-ledger path writes native balances), so the sort key omits the ledger; corrected the stale "only the live buffer dedups balances" comment.
- **Cleanups.** DRY'd the per-operation account-change loop (`buildAccountChanges` shared with the fee path); decoupled a buffer test helper from the production sort-key encoding; derived the fee-bump integration balance from `DefaultFundingAmount`/`FeeBumpFixtureFeeStroops` instead of a hardcoded string.


internal/indexer/processors/accounts.go
```diff
@@ -19,6 +19,31 @@ const (
...
+// in the high bits — above the tx (24 bits) and op (13 bits) fields — so fee < operation
+// < refund holds for any tx/op count, and the key is a strict total order with no ties
+// (≤1 fee and ≤1 refund per account per tx; operations are unique per (tx, op)). That
+// makes the dedup independent of push/merge order.
+//
+// The ledger is intentionally omitted: the only buffer that deduplicates balances is the
+// per-ledger live buffer, so all keys compared share one ledger. If balances are ever
+// deduplicated across ledgers in a single buffer, prepend the ledger as the top term.
+func accountSortKey(phase uint8, txIndex, opIndex uint32) int64 {
+	return int64(phase)<<37 | int64(txIndex)<<13 | int64(opIndex)
```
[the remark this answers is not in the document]
aditya1702: We dont process balances during backfill

internal/indexer/processors/accounts.go
```diff
@@ -19,6 +19,31 @@ const (
...
+// selects the chronologically-last change and writes its absolute balance. Phase sits
+// in the high bits — above the tx (24 bits) and op (13 bits) fields — so fee < operation
+// < refund holds for any tx/op count, and the key is a strict total order with no ties
+// (≤1 fee and ≤1 refund per account per tx; operations are unique per (tx, op)). That
+// makes the dedup independent of push/merge order.
+//
+// The ledger is intentionally omitted: the only buffer that deduplicates balances is the
+// per-ledger live buffer, so all keys compared share one ledger. If balances are ever
+// deduplicated across ledgers in a single buffer, prepend the ledger as the top term.
+func accountSortKey(phase uint8, txIndex, opIndex uint32) int64 {
```
aristidesstaffieri, line 48: **Cross-ledger dedup: the ledger-less sort key breaks the backfill `batchBuffer`.**

This doc-comment states "the only buffer that deduplicates balances is the per-ledger live buffer, so all keys compared share one ledger." That precondition doesn't hold: the backfill path (`internal/services/ingest_backfill.go:319`) creates a single `batchBuffer` and reuses it across up to `backfillDBInsertBatchSize` ledgers (default 100) before flushing.

So keys from different ledgers *are* compared. If account X changes via an operation in ledger N (large SortKey) and via a fee debit in ledger N+1 (`phaseFee`/tx0 → SortKey ≈ 0), the `existing.SortKey > change.SortKey` guard keeps ledger N and silently discards the newer ledger-N+1 balance — leaving a stale native balance after backfill. This is exactly the across-ledger case the comment's last line warns about. Suggest prepending the ledger sequence as the top term of the key.

internal/indexer/indexer.go
```diff
@@ -273,6 +282,18 @@ func (i *Indexer) processTransaction(ctx context.Context, tx ingest.LedgerTransa
 		}
 	}
 
+	// Fold transaction fee-phase changes (fee debits + Soroban fee refunds) into
+	// native balances. These are charged/refunded outside any operation's meta, so
+	// the per-operation loop above never sees an account whose balance moves only in
+	// those phases — e.g. a fee-bump fee source (issue #637).
+	feeAccountChanges, feeErr := i.accountsProcessor.ProcessTransactionFees(ctx, tx)
```
aristidesstaffieri, line 289: **Phase-dominant ordering inverts close order within a ledger for shared accounts.**

Fee/refund changes are pushed into the same per-ledger buffer as operation changes, and `accountSortKey` makes `phase` the dominant term with no global within-ledger order. For an account touched in two different transactions of one ledger, a Soroban fee **refund** (`phaseRefund=2`) in an earlier tx outranks an **operation** change (`phaseOperation=1`) in a later tx. The buffer keeps the earlier refund balance and discards the chronologically-later operation result, so the persisted balance reflects an earlier point in the ledger.

(Fee-vs-operation within a tx is fine — fees are all charged before operations apply. It's the cross-transaction refund-vs-operation case that inverts.) Same root cause as the sort-key comment in `processors/accounts.go`; encoding the ledger + a within-ledger order resolves it.

internal/indexer/indexer_buffer.go
```diff
@@ -241,8 +241,8 @@ func (b *IndexerBuffer) PushAccountChange(accountChange types.AccountChange) {
 	accountID := accountChange.AccountID
 	existing, exists := b.accountChangesByAccountID[accountID]
 
-	// Keep the change with highest OperationID
-	if exists && existing.OperationID > accountChange.OperationID {
+	// Keep the change with highest SortKey
+	if exists && existing.SortKey > accountChange.SortKey {
```
aristidesstaffieri: **CREATE→REMOVE no-op `delete` (line 251 below) can let a later lower-key change resurrect the account.**

This guard only returns early when `exists` is true. After the CREATE→REMOVE no-op at line 251 deletes the account from the buffer, a later `PushAccountChange`/`Merge` for the same account ID finds `exists==false`, falls through this guard, and unconditionally re-inserts. A subsequent fee-phase UPDATE (lower sort key than the create+remove operation) would resurface a merged-away account as a phantom `native_balances` UPSERT. The mirrored guard in `Merge` (~line 460) has the same shape.

This may be unreachable today since fee collection precedes op-apply (a same-ledger-created account can't also be a fee source), but relying on that ordering invariant is fragile. Worth hardening — e.g. tombstone the removed account ID rather than deleting the key outright.

internal/indexer/processors/accounts.go
```diff
@@ -59,7 +84,66 @@ func (p *AccountsProcessor) ProcessOperation(ctx context.Context, opWrapper *Tra
...
+	if err != nil {
+		return nil, err
+	}
+
+	return append(feeChanges, refundChanges...), nil
+}
+
+// feeAccountChanges converts the account-typed entries in a fee-phase change set into
+// AccountChanges stamped with the given sort key.
+func (p *AccountsProcessor) feeAccountChanges(changes []ingest.Change, ledgerSeq uint32, sortKey int64) ([]types.AccountChange, error) {
```
aristidesstaffieri: **Near-duplicate of `ProcessOperation`'s change loop — consider sharing one.**

This loop (skip non-account changes → `buildAccountChange` → skip → append) is structurally identical to the loop in `ProcessOperation` (lines 82–96); the only difference is the sort key, which is constant across all changes in a single call. Since this PR is already adding `feeAccountChanges` and touched `ProcessOperation`'s loop (the `processAccountChange` → `buildAccountChange` rename), `ProcessOperation` could compute its sort key once and delegate here, e.g. `feeAccountChanges(changes, ledgerSeq, accountSortKey(phaseOperation, txIndex, opIndex))`. Keeps the skip/append/error handling in one place so the two paths can't silently diverge.

internal/indexer/indexer_buffer_test.go
```diff
@@ -945,3 +945,215 @@ func TestIndexerBuffer_ProtocolContracts(t *testing.T) {
...
+const (
+	rankFee    int64 = 0
+	rankOp     int64 = 1
+	rankRefund int64 = 2
+)
+
+// accountRank builds an order-preserving dedup key (phase, then tx, then op) mirroring how
+// AccountsProcessor ranks native-balance changes. The buffer only compares the int64, so
+// only the ordering matters here — not the exact encoding.
+func accountRank(phase, tx, op int64) int64 {
```
aristidesstaffieri, line 992: **Test re-encodes the production bit layout — it won't catch an encoding change.**

`accountRank` (and the `rankFee/rankOp/rankRefund` constants above) re-implement `accountSortKey`'s exact layout (`phase<<37 | tx<<13 | op`) and phase ordering. If the production encoding ever changes (different shift widths, reordered phases), this helper keeps the old layout and these tests still pass green — so they'd no longer exercise the real ordering they claim to mirror. Suggest importing/calling the production `accountSortKey` and phase constants directly instead of duplicating them here.

internal/integrationtests/account_balances_test.go
```diff
@@ -373,3 +373,25 @@ func (suite *AccountBalancesAfterLiveIngestionTestSuite) TestLiveIngestion_Holde
...
+		context.Background(),
+		suite.testEnv.FeeBumpSourceKP.Address(),
+	)
+	suite.Require().NoError(err)
+	suite.Require().Len(balances, 1, "fee-bump fee source should have exactly one (native) balance")
+
+	nb, ok := balances[0].(*types.NativeBalance)
+	suite.Require().True(ok, "expected a native balance, got %T", balances[0])
+	suite.Require().Equal(types.TokenTypeNative, nb.GetTokenType())
+	suite.Require().Equal("9999.9999800", nb.GetBalance(),
```
aristidesstaffieri: **Derive this expected balance from the constants instead of hardcoding.**

The fee charge is now encoded in three independent spots: `FeeBumpFixtureFeeStroops` (`= MinBaseFee * 2`, submit.go), this literal `"9999.9999800"`, and the "200 stroops" prose comment above. `FeeBumpFixtureFeeStroops` was added precisely to make the charge deterministic, yet this assertion never references it — if `MinBaseFee` or the fixture's op count changes, the constant updates but the string doesn't, and the test fails with a magic-number mismatch. Suggest `amount.StringFromInt64(amountToStroops(DefaultFundingAmount) - FeeBumpFixtureFeeStroops)` so they stay in lockstep.

internal/indexer/processors/accounts.go
```diff
@@ -73,9 +74,70 @@ func (p *AccountsProcessor) ProcessOperation(ctx context.Context, opWrapper *Tra
...
+	defer func() {
+		if p.metricsService != nil {
+			duration := time.Since(startTime).Seconds()
+			p.metricsService.StateChangeProcessingDuration.WithLabelValues("AccountsProcessor").Observe(duration)
+		}
+	}()
+
+	ledgerSeq := tx.Ledger.LedgerSequence()
+
+	feeChanges, err := p.feeAccountChanges(tx.GetFeeChanges(), ledgerSeq, toid.New(int32(ledgerSeq), 0, 0).ToInt64())
```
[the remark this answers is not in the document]
aditya1702: Already fixed in c141ea35 — the key now includes `tx.Index` (`accountSortKey(phase, txIndex, opIndex)`). This predates that refactor.

internal/indexer/processors/accounts.go
```diff
@@ -73,9 +74,70 @@ func (p *AccountsProcessor) ProcessOperation(ctx context.Context, opWrapper *Tra
...
+	if err != nil {
+		return nil, err
+	}
+
+	refundChanges, err := p.feeAccountChanges(tx.GetPostApplyFeeChanges(), ledgerSeq, toid.New(int32(ledgerSeq)+1, 0, 0).ToInt64()-1)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(feeChanges, refundChanges...), nil
```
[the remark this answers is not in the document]
aditya1702: Already fixed in c141ea35 — the key now includes `tx.Index` (`accountSortKey(phase, txIndex, opIndex)`). This predates that refactor.

internal/indexer/processors/accounts.go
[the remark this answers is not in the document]
aditya1702: Not a stale-balance bug: backfill dedups account changes in memory but never persists them — `insertIntoDB` writes only txs/ops/state_changes; only the live per-ledger path calls `ProcessTokenChanges`. Fixed the misleading doc comment in 83691aa4.

internal/indexer/processors/accounts.go
aditya1702: Not a stale-balance bug: backfill dedups account changes in memory but never persists them — `insertIntoDB` writes only txs/ops/state_changes; only the live per-ledger path calls `ProcessTokenChanges`. Fixed the misleading doc comment in 83691aa4.

internal/indexer/indexer.go
aditya1702: `GetPostApplyFeeChanges` returns only the protocol-23+ refund, which Core applies **after all operations** (`processPostTxSetApply`, global-final) — so it's the account's chronologically-last change and refund-dominant ordering is correct. Pre-23 refunds go to `TxChangesAfter`, which this path doesn't read, so there's nothing to mis-order. The per-tx inversion is the pre-23 model this method never surfaces. Doc clarified in 83691aa4.

internal/indexer/indexer_buffer.go
aditya1702: Fixed in 6d89b222 — replaced the `delete` with a tombstone recording the cancellation's order: later changes with order ≤ it are dropped, a strictly-higher one re-creates the key. Applied to account/trustline/SAC (push + merge) via a shared helper; merge propagates tombstones.

internal/indexer/processors/accounts.go
aditya1702: Done in 83691aa4 — `ProcessOperation` now delegates to the shared `buildAccountChanges` (renamed from `feeAccountChanges`).

internal/indexer/indexer_buffer_test.go
aditya1702: Done in 6d89b222 — `accountRank` no longer copies the production bit layout; the real encoding is covered by `TestAccountSortKey` in the processors package.

internal/integrationtests/account_balances_test.go
aditya1702: Done in 525b4a2b — now derived: `amount.StringFromInt64(amount.ParseInt64(DefaultFundingAmount) − FeeBumpFixtureFeeStroops)`.

aristidesstaffieri (approved):

PR #642, opened 2026-06-25
feat(services): windowed coalescing + core protocol-migration

### LINKED ISSUE #645: Windowed coalescing + foldable processor for fast protocol data-migration
The protocol data-migration path (`protocol-migrate`) needs to be fast, batch-correct, and hand off cleanly to live ingestion.

**Scope:**
- **Windowed coalescing** — commit K ledgers per DB transaction (`--window-size`, default 100) at `synchronous_commit=off`; shrink to K=1 at the live tip so handoff stays ledger-precise (the CAS cursor is the only gate).
- **Foldable `ProtocolProcessor`** — `ProcessLedger` accumulates into staged sets; the engine owns `Reset()`. Batch-equivalence is the contract.
- **SEP-41** — the validator is the sole owner of `contract_tokens`; wbclient gains SEP-41 (non-SAC) balance support.
- **Live resolution by event id** — resolve only the contracts that emitted events, not a protocol's whole set.
- **Migration metrics** on `--metrics-port`; staged progress logs; RPC keep-alive fix; `LOG_LEVEL` env-bind.
- **Production-faithful E2E** — runs the real binaries through the datastore backend + a live ledger exporter.

aditya1702:
Closes #645

⚠️ **Stacked on #638** (datastore backend). Its base is `pr/datastore`, so this diff shows only the core changes. Review/merge #638 first; GitHub retargets this to `main` automatically once #638 lands.

The migration speedup and supporting changes:
- **Windowed coalescing** — commit K ledgers per DB transaction (`--window-size`, default 100) at `synchronous_commit=off`. Tip-shrink to K=1 at the live tip keeps handoff ledger-precise (the CAS cursor is the only gate).
- **Foldable `ProtocolProcessor`** — `ProcessLedger` accumulates into staged sets; the engine owns `Reset()`. Batch-equivalence is the contract.
- **SEP-41** — validator is the sole owner of `contract_tokens`; wbclient gains SEP-41 (non-SAC) balance support.
- **Live resolution by event id** — resolve only the contracts that emitted events, not a protocol's whole set.
- **Migration metrics** on `--metrics-port`; staged progress logs; RPC keep-alive fix; `LOG_LEVEL` env-bind.
- **Production-faithful E2E** — runs the real binaries through the datastore backend + a live ledger exporter.


aristidesstaffieri:
The folding contract here is "keep accumulating until someone calls `Reset()`", and right now nothing stops you from calling `ProcessLedger` twice without a Reset in between. Both callers happen to get it right (live ingestion Resets inside the retry block before staging, migration Resets after each flush succeeds), so it's fine today. But it's the kind of thing that breaks quietly later: if someone adds a retry around the migration persist, or writes a new caller that folds → persists → folds on the same processor, the deltas just double up and get committed as if they were correct. No error, no log, wrong balances.

Could we add a guard so that misuse blows up instead of silently corrupting? Something like a `sealed` flag that `Persist*` sets and `Reset()` clears, checked at the top of `ProcessLedger`:

```go
func (p *processor) ProcessLedger(_ context.Context, input services.ProtocolProcessorInput) error {
    if p.sealed {
        return fmt.Errorf("sep41: ProcessLedger called after Persist without Reset(); deltas would double-count")
    }
    // ...
}
```

That keeps both current callers green (live Resets every ledger, migration Resets right after flush, and the handoff path Resets without persisting so the flag stays clear). Worth a line in the `ProcessLedger` doc comment spelling out that Persist seals the staged sets, plus a quick test that fold → persist → fold-without-reset errors.

Not a blocker since it works today, just want to avoid leaving a rake in the grass.

aristidesstaffieri:
`id.Value()` returns `(nil, nil)` for an empty `HashBytea` (the `h == ""` branch up in `types.go`), so when the id is empty `val` is a nil interface and `val.([]byte)` panics instead of giving us an error. Since this runs on the ingestion goroutine, that's a crash rather than a handled failure.

Nothing hits it today — `distinctEventContractIDs` only ever produces 64-char hex IDs — but `BatchGetByContractIDs` is exported and takes arbitrary `[]types.HashBytea`, so the next caller that happens to pass an empty id takes the whole goroutine down. Cheap to make safe with the comma-ok form:

```go
b, ok := val.([]byte)
if !ok {
    return nil, fmt.Errorf("contract id %q produced a non-bytes value", id)
}
idBytes[i] = b
```

Or reject empty ids up front if an empty contract id is never legitimate here. Either way it turns a panic into an error the caller can deal with.

aristidesstaffieri:
Progress logging and the `getHealth` tip refresh are both gated on `seq % windowSize == 0` here, so the commit-window size also ends up controlling observability cadence. With the default window of 1 that logs + hits `getHealth` every ledger; with `--window-size=1000` you only get a progress line / tip update every 1000 ledgers. Might be worth giving progress its own interval (a fixed `seq%100` like before, or a separate flag) so the two concerns aren't sharing one knob.

aristidesstaffieri:
Small hot-path nit: `distinctEventContractIDs` does `hex.EncodeToString(ev.ContractId[:])` to build the `HashBytea`s, and `BatchGetByContractIDs` then hex-decodes each one right back to bytes for the `bytea[]` query. The raw 32 bytes are already in hand at `ev.ContractId[:]`, so passing `[][]byte` straight through (dedup on `[32]byte`) would skip both the encode and the decode on every ledger — and as a bonus avoid the `val.([]byte)` assertion entirely.

aditya1702:
@aristidesstaffieri addressed in 4083e3cc:

**1. Folding guard.** Added a `needsReset` flag on the SEP-41 processor — `Persist*` sets it, `Reset()` clears it, and `ProcessLedger` returns an error if it's asked to fold while set. Both current callers stay green (live ingestion Resets before every ledger; migration Resets after each flush, including the CAS-fail handoff). Doc comment on `ProcessLedger` spells out that Persist seals the staged sets, plus a `TestProcessor_NeedsResetGuard` that asserts fold → persist → fold-without-reset errors.

**2. `val.([]byte)` panic.** Resolved by #4 below rather than a comma-ok: `BatchGetByContractIDs` no longer routes ids through `HashBytea.Value()`, so there's no assertion left to panic on.

**3. Observability tied to window size.** Pulled out a fixed `progressLogInterval = 100`. Progress logging and the `getHealth` target-tip refresh now gate on `seq % progressLogInterval` instead of `seq % windowSize`, so `--window-size` only governs commit batching. `flushWindow` was already separate (per-tracker `pending >= windowSize`), so commit behavior is unchanged.

**4. Hot-path encode/decode.** `distinctEventContractIDs` now dedups on the raw `xdr.ContractId` (`[32]byte`) and returns `[][]byte`; `BatchGetByContractIDs` takes `[][]byte` straight to the `bytea[]` query. The per-ledger hex encode→decode round-trip and the `[]byte` assertion are both gone.

aristidesstaffieri (approved):

PR #648, opened 2026-07-02
perf(indexer): single-owner lock-free IndexerBuffer + per-tx result folding

### LINKED ISSUE #647: Optimize IndexerBuffer: single-owner, lock-free, per-tx result folding
What

Make `internal/indexer.IndexerBuffer` single-owner / lock-free, and stop allocating a full `IndexerBuffer` per transaction. Workers now build a lightweight per-tx result that is folded into one ledger buffer serially.

Why

In production each `IndexerBuffer` is only ever touched by one goroutine — the indexer creates one per tx worker and merges them serially after the pool finishes. So the `RWMutex`, the thread-safe participant sets, and the defensive `maps.Clone()` in the getters were all guarding a race that cannot happen. On top of that, allocating N full buffers per ledger (~15 maps each) and then merging them added allocations and a second, redundant dedup pass on the hot ingestion path.

Outcome

- Lock-free single-owner buffer; per-participant `Push*` drop to 0 allocs/op.
- N per-ledger buffer allocations and the `Merge`/`mergeWithTombstone` code path removed.
- Dedup / tombstone (highest-order-wins) semantics unchanged; full `-race` unit suite green.

### LINKED ISSUE #653: indexer: pushWithTombstone create+remove dedup is order-sensitive within a single transaction
**Summary**

`pushWithTombstone` (`internal/indexer/indexer_buffer.go`) nets a create+remove of the same key within one ledger to nothing via a tombstone. That is correct only if the CREATE is folded before the REMOVE. Within a single transaction, changes are appended while ranging the `opsParticipants` map in `processTransaction` (`internal/indexer/indexer.go`), so the relative order of a CREATE-op change and a REMOVE-op change is nondeterministic — which flips the result.

**Behavior** (for a tx that both creates and removes the same key: `CreateAccount`+`AccountMerge`; `ChangeTrust` add + `ChangeTrust` limit-0)

- CREATE folds first → REMOVE hits `isNoopRemove`, key deleted + tombstone recorded → nets to nothing ✓
- REMOVE folds first → no existing entry, so `isNoopRemove` (guarded by `exists`) never fires; REMOVE stored as a phantom with no tombstone, and the lower-order CREATE is then dropped at the `order(existing) > order(change)` check → a spurious REMOVE persists ✗

Same ledger + input, different output between runs. Cross-transaction (create in tx1, remove in tx2 of the same ledger) is safe — results fold in tx-index order.

**Impact**

Nondeterministic ingestion output for identical input (reproducibility / backfill consistency). The downstream effect of the spurious REMOVE depends on how REMOVE changes are applied (may be an idempotent no-op).

**Scope note**

Pre-existing in the tombstone code introduced by #641; not introduced or worsened by #648 (which preserves the within-tx map-iteration order — its "dedup/tombstone semantics unchanged" claim holds for this hazard). Filed separately so it's tracked against the owning code.

**Fix direction**

Fold each change slice in ascending order value before dedup (sort by `SortKey`/`OperationID`), or emit changes ordered from `processTransaction`, so CREATE always precedes REMOVE. Add a regression test with a create+remove-in-one-tx ledger — the map iteration means a naive test passes most runs.

aditya1702:
## What

Rewrites `internal/indexer` buffering around the fact that each buffer is only ever owned by one goroutine, and trims allocations along the whole ingestion hot path:

- **Lock-free `IndexerBuffer`** — the `RWMutex`, thread-safe participant sets, and defensive `maps.Clone()` getters are removed.
- **Per-tx result folding** — parallel workers build a lightweight `TransactionResult` instead of a full buffer each; a serial fold (`IngestTransactionResult`) replays them into the single ledger buffer with the same dedup/tombstone semantics. `Merge` is deleted.
- **Allocation cuts** — participant sets are plain `map[string]struct{}` end to end (buffer → services → data COPY loops, dropping `deckarep/golang-set` from that path), the trustline asset parse + SHA-1 ID derivation is memoized per asset string, the rarely-used per-tx result maps allocate lazily, and known-size slices are pre-sized.
- **One buffer per ingestion loop** — the live loop reuses a single buffer and `Clear()`s it before each ledger rather than allocating one per ledger, matching what backfill already did per batch. The asset-parse memo and the maps' backing arrays now survive across ledgers on both paths.
- **Uniform set variant in the soroban participant path** — the helpers in `contract_operations.go` build thread-unsafe sets like the rest of that path, removing the last mutex-backed sets from per-operation participant extraction. This is allocation- and latency-neutral in the benchmarks; the reason to do it is that mixing `golang-set` variants panics at runtime rather than failing to compile, since `Union` type-asserts its argument to the receiver's concrete type.

`main` is merged in: liquidity-pool changes (#650) and the `state_change_id` sub-base machinery (#667–#670) flow through the `TransactionResult` fold.

Also fixes #653: the worker now walks a transaction's operations in ascending opID order, so `pushWithTombstone`'s same-ledger create+remove netting sees CREATE before REMOVE deterministically (previously map-order iteration could leave a spurious REMOVE). A 12-op regression test asserts ordered emission and the netting; the sort costs nothing measurable (one small slice per tx).

## Why

In production each `IndexerBuffer` is single-owner: workers never share one, and the fold runs serially after the worker pool finishes. The locks, thread-safe sets, and clones guarded a race that cannot happen, while every ledger allocated N full buffers (~15 maps each) plus a redundant second dedup pass in `Merge`. Closes #647.

## Benchmarks

`go test -bench=RealLedger -benchmem ./internal/indexer` — real pubnet ledger fixtures, Apple M2 Pro. *Fold* is the serial phase this PR introduces; *full process* is the whole per-ledger indexing pass. Fold numbers are steady state (asset-parse memo warm), which is now what both ingestion paths do.

| Ledger | Fold | Full process |
|--------|------|--------------|
| `51000294` (largest) | 0.51 ms · 370 KB · 2.9K allocs | 6.9 ms · 25.3 MB · 263K allocs |
| `63092988` | 0.33 ms · 239 KB · 1.8K allocs | 4.8 ms · 15.2 MB · 160K allocs |
| `57500008` | 0.25 ms · 182 KB · 1.4K allocs | 2.8 ms · 12.8 MB · 135K allocs |

## Safety

- A race-detector tripwire test (`TestIndexer_ProcessLedgerTransactions_RealLedgerParallel`) drives every real fixture through the parallel path into one shared buffer; any future worker write to the buffer fails `make unit-test` (`-race`).
- Dedup/tombstone (highest-order-wins) semantics are unchanged — the fold reuses the same `Push*` dedup.
- Reusing the live buffer makes `Clear()` a correctness contract, not just an allocation one: `accountSortKey` deliberately omits the ledger from its key, so account changes from two ledgers must never coexist in the balance-change maps. A regression test covers it — create+merge in ledger N, `Clear()`, then a lower-sort-key fee change in N+1 must still land, since a surviving tombstone would silently drop a real balance instead of failing loudly. Verified by mutation (deleting `clear(b.accountTombstones)` fails the test).
- `make check` and the full `-race` unit suite pass.

Closes #647. Fixes #653.


internal/indexer/indexer.go
```diff
@@ -207,79 +202,71 @@ func (i *Indexer) processTransaction(ctx context.Context, tx ingest.LedgerTransa
...
 
-	// Insert transaction participants
-	for participant := range txParticipants.Iter() {
-		buffer.PushTransaction(participant, *dataTx)
+	result := &TransactionResult{
+		Transaction:           dataTx,
+		TxParticipants:        txParticipants.ToSlice(),
+		Operations:            make(map[int64]*types.Operation, len(opsParticipants)),
+		OpParticipants:        make(map[int64][]string, len(opsParticipants)),
+		ProtocolWasmBytecodes: make(map[string][]byte),
```
aristidesstaffieri: `ProtocolWasmBytecodes` and `ContractEvents` are allocated for every `TransactionResult`, but almost no transaction produces either — 2N throwaway maps per ledger on the hot path this PR is optimizing. Leaving them nil and `make()`-ing lazily on first insert avoids the allocations; the range loops over them in `IngestTransactionResult` are already no-ops over nil maps.

internal/indexer/indexer_buffer.go
```diff
@@ -68,7 +62,6 @@ type ContractEventKey struct {
 }
 
 type IndexerBuffer struct {
```
aristidesstaffieri, line 69: Now that the `RWMutex` is gone and correctness rests on the single-owner "serial fold after `group.Wait()`" convention, consider adding a test that runs `ProcessLedgerTransactions` on a multi-tx ledger under the race detector (`-race`). The convention is correct today, but it's enforced only by a comment — a race test would catch any future change that accidentally folds or persists a buffer from a background goroutine at review time, instead of it surfacing as a `fatal error: concurrent map writes` in production.

internal/indexer/indexer.go
```diff
@@ -160,44 +149,50 @@ func (i *Indexer) ProcessLedgerTransactions(ctx context.Context, transactions []
...
 	if err != nil {
-		return 0, fmt.Errorf("creating data transaction: %w", err)
+		return nil, fmt.Errorf("creating data transaction: %w", err)
 	}
 
-	// Count all unique participants for metrics
+	// Count all unique participants for metrics. This transient set unions the participants
+	// processor's (thread-safe) sets, so it must be thread-safe too — deckarep's Union
+	// type-asserts its argument to the receiver's variant.
 	allParticipants := set.NewSet[string]()
```
aristidesstaffieri: `allParticipants` uses the thread-safe `set.NewSet` to compute the `ParticipantCount` metric, taking a lock on every `Add`/`Union` for a local that never escapes this goroutine. (This allocation predates the PR, but since it now routes the count through `result.ParticipantCount` and rewrites this block, flagging while it's in view.) It's only forced thread-safe because deckarep's `Union` type-asserts its argument to the receiver's variant — a plain `map[string]struct{}`, or summing the source sets' cardinalities, drops the per-tx locking.

internal/indexer/indexer_buffer.go
```diff
@@ -345,201 +273,141 @@ func (b *IndexerBuffer) PushSACBalanceChange(sacBalanceChange types.SACBalanceCh
...
+	for _, participant := range r.TxParticipants {
+		b.PushTransaction(participant, r.Transaction)
+	}
+
+	for opID, participants := range r.OpParticipants {
+		operation := r.Operations[opID]
+		for _, participant := range participants {
+			b.PushOperation(participant, operation, r.Transaction)
 		}
 	}
```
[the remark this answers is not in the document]
aditya1702: Fixed in 95b73622 — the fold now skips (and logs) any `OpParticipants` key with no matching operation instead of dereferencing nil.

internal/indexer/indexer.go
aditya1702: Fixed in 95b73622 — both maps are gone from the `TransactionResult` literal and allocate lazily at their single insert sites; the fold's range over nil is a no-op as you noted.

internal/indexer/indexer_buffer.go
aditya1702: Added in bb9b4ca8 — `TestIndexer_ProcessLedgerTransactions_RealLedgerParallel` drives every real pubnet fixture through `ProcessLedgerTransactions` into one shared buffer, so `make unit-test` (`-race`) fails on any future worker write to the buffer.

internal/indexer/indexer.go
aditya1702: Fixed in 95b73622 — the count is a plain `map[string]struct{}`, and the participants processor now builds thread-unsafe sets (its Soroban merge adds elements directly instead of `Union`, which type-asserts its argument to the receiver's variant). 537fea3f then goes further and removes deckarep sets from the buffer's participant storage and getters end to end (buffer → services → data COPY).

internal/indexer/indexer_buffer.go
```diff
@@ -397,232 +337,187 @@ func (b *IndexerBuffer) PushLiquidityPoolShareChange(change types.LiquidityPoolS
...
-		if existing, exists := b.participantsByToID[toID]; exists {
-			// Merge into existing set - iterate and add (Union creates new set)
-			for participant := range otherParticipants.Iter() {
-				existing.Add(participant)
-			}
-		} else {
-			// Clone the set instead of creating empty + iterating
-			b.participantsByToID[toID] = otherParticipants.Clone()
+// Each change-family slice (TrustlineChanges, AccountChanges, SACBalanceChanges, LPShareChanges,
+// LPChanges) must be ordered ascending by its order value per key (CREATE before REMOVE):
```
JiahuiWho: Can we reword the comment to describe the weaker guarantee that actually holds? 
`AccountChanges` doesn't guarantee this. Fee changes are added after operation changes in `processTransaction`. That's still correct because netting only requires `CREATE` before `REMOVE`, and fee changes are never CREATE/REMOVE.

internal/indexer/indexer_buffer.go
```diff
@@ -397,232 +337,187 @@ func (b *IndexerBuffer) PushLiquidityPoolShareChange(change types.LiquidityPoolS
...
+func (b *IndexerBuffer) IngestTransactionResult(r *TransactionResult) {
+	for _, participant := range r.TxParticipants {
+		b.PushTransaction(participant, r.Transaction)
+	}
+
+	for opID, participants := range r.OpParticipants {
+		// Invariant: every OpParticipants key must resolve to an operation in r.Operations.
+		operation := r.Operations[opID]
+		if operation == nil {
+			log.Errorf("operation %d missing from TransactionResult.Operations (tx %s); dropping its participants", opID, r.Transaction.Hash)
```
JiahuiWho: nit: the equivalent worker-side check in `processTransaction` uses `log.Ctx(ctx).Errorf`. If this ever fires you'd want the ledger context. Consider threading ctx through.

internal/indexer/indexer_buffer.go
```diff
@@ -94,6 +86,11 @@ type IndexerBuffer struct {
 	lpShareTombstones     map[LiquidityPoolShareChangeKey]int64
 	lpTombstones          map[string]int64
 	uniqueTrustlineAssets map[uuid.UUID]data.TrustlineAsset
+	// parsedAssetsByString memoizes the parse + deterministic-ID derivation per unique asset
+	// string (nil value = string is known-invalid). It is content-derived — the same string always
+	// yields the same result — so it is never cleared in Clear() and is bounded by the number of
+	// unique asset strings ever seen.
+	parsedAssetsByString  map[string]*data.TrustlineAsset
```
JiahuiWho, line 94: Since live ingestion creates a fresh buffer per ledger, this memo only warms up within a single ledger there. The steady-state benchmark numbers apply to backfill. Worth a follow-up to reuse one buffer + `Clear()` in the live loop like backfill does?

internal/indexer/processors/participants.go
```diff
@@ -189,5 +189,10 @@ func (p *ParticipantsProcessor) GetOperationParticipants(op *TransactionOperatio
 		return nil, fmt.Errorf("getting soroban participants: %w", err)
 	}
 
-	return participants.Union(sorobanParticipants), nil
+	// sorobanParticipants comes from contract_operations.go's thread-safe sets, so merge by
+	// element rather than Union, which type-asserts its argument to the receiver's variant.
+	for _, participant := range sorobanParticipants.ToSlice() {
+		participants.Add(participant)
+	}
+	return participants, nil
```
JiahuiWho: Since the soroban sets in `contract_operations.go` are also built and consumed within a single worker, converting them to `NewThreadUnsafeSet` too would remove the last mutex-backed sets on the hot path, and eliminate the mixed-variant panic hazard this workaround guards against. Follow-up is fine

JiahuiWho (approved):

internal/indexer/indexer_buffer.go
aditya1702: Reworded in 3713f9e5.

Confirmed the gap: `processTransaction` appends `feeAccountChanges` after the sorted-opID walk, and `phaseFee` sorts *below* every operation, so those entries carry the lowest keys in the slice. `AccountChanges` is the only family that isn't globally ascending — the other four are built solely inside the sorted loop.

The godoc now states the weaker invariant that actually holds (a key's create/add must precede the remove that cancels it, not that the slice is sorted), and spells out why the fee-phase entries can't break it: `buildAccountChange` derives CREATE/REMOVE from `Pre`/`Post` nil-ness, and a fee debit or Soroban refund always has both, so they're always `AccountOpUpdate` and can never satisfy `accountIsNoopRemove`.

internal/indexer/indexer_buffer.go
aditya1702: Logging the ledger in 3713f9e5, but not via ctx — `log.Set` isn't called anywhere in this repo, so `log.Ctx(ctx)` resolves to `DefaultLogger` and would carry no ledger fields. The sibling check in `processTransaction` gets nothing extra from it either. Threading ctx would mean changing `IndexerBufferInterface` plus 7 test/bench call sites for no behavioral gain today.

The ledger is already on the value being folded, so the line now reads `(ledger %d, tx %s)` off `r.Transaction.LedgerNumber`.

If you'd still rather have the call-site consistency so a future `log.Set` picks this up automatically, say so and I'll thread ctx through as well.

internal/indexer/indexer_buffer.go
aditya1702: Done in a7d1b2bb rather than deferred. Verified the buffer never escapes an iteration — the persist retries run synchronously on the ingestion goroutine — so it's hoisted above the loop with `Clear()` at the top of each iteration (top rather than after the persist, so the reset is unconditional regardless of how the previous iteration ended).

Worth flagging what this makes load-bearing. `accountSortKey` deliberately omits the ledger from its key, and its godoc pinned that to "the live path uses a fresh buffer per ledger". With reuse, the safety now rests on `Clear()` wiping the balance-change maps **and** their tombstones — and nothing tested that. So there's a new regression test: create+merge in ledger N, `Clear()`, then a lower-sort-key fee change in N+1 must still land. A surviving tombstone would silently *drop* a real balance rather than fail loudly. Mutation-checked by deleting `clear(b.accountTombstones)` and confirming the test fails.

One honest caveat on the payoff: in live mode the memo win is negligible — microseconds of asset parsing against a ~180 ms persist at ~1 ledger/5s. The real gain is the map capacity retention and consistency with backfill. Note the memo's bound also shifts from per-ledger to per-process (unique asset strings, ~200 B each), which is the property backfill already had across a run.

internal/indexer/processors/participants.go
aditya1702: Done in 7a236582 rather than deferred. All 8 constructors in `contract_operations.go` are `NewThreadUnsafeSet` now, with the test expectations converted in lockstep — `assert.Equal` compares the sets' concrete types, not just their contents.

The hazard is real: both variants' `Union` type-asserts its argument to its own concrete type (`threadsafe.go:134`, `threadunsafe.go:308`), so mixing panics at runtime rather than failing to compile. I kept the merge element-wise instead of switching to `Union`, but for a different reason now — `Union` allocates a third set and `Each` doesn't — and rewrote the comment to say that. The variant requirement is recorded on `participantsForSorobanOp` so a thread-safe set doesn't get reintroduced there.

On "remove the last mutex-backed sets on the hot path": structurally true, but not measurable. benchstat over the five real-ledger process benchmarks (6 runs each) puts every ledger within noise (p > 0.05), with a ~0.2% allocation drop on the most soroban-heavy one from dropping `ToSlice`. Flagging that so the PR doesn't claim a speedup it doesn't have — the value here is the removed panic hazard.

Heads-up: this and the buffer reuse pull 4 files in past what you approved (`ingest_live.go`, `accounts.go`, `contract_operations.go` + its test), so they're worth a re-look.

PR #657, opened 2026-07-08
feat(protocols): Add code to parse and process ContractData entires for Blend processors

### LINKED ISSUE #680: Migration engine can fold ledgers ahead of live's classification frontier, permanently losing protocol state
## Summary

The protocol migration engine can fold ledgers that live ingestion has not yet reached. Because live ingestion is the **only** producer of protocol contract membership, any contract deployed in that lead is invisible to the engine for the entire span between its deploy ledger and the point live catches up — and cursor-passed ledgers are never replayed, so that protocol state is permanently lost.

The classification-gap repair added in #657 heals exactly one ledger of this span (the deploy ledger). The remainder is unbounded, deterministic, and silent.

All line references are at `d9aac773` (head of #657).

## Why it happens

Two independent position markers:

- **Live's position** — `latest_ledger_cursor`, resumed from the DB on startup, so it can sit arbitrarily far behind tip after a restart or backlog.
- **The protocol's position** — `protocol_<id>_history_cursor` / `_current_state_cursor`, advanced by whichever producer wins the CAS.

And one asymmetry that does the damage: **live ingestion is the sole classifier.** `internal/services/ingest_live.go:238` is the only writer of `protocol_contracts` anywhere in the tree. The engine only ever reads committed rows (`refreshTrackerContracts` → `GetByProtocolID`, `internal/services/protocol_migrate.go:548`).

So a contract does not exist, as far as the engine is concerned, until **live personally reaches its deploy ledger**. Nothing currently prevents the engine's fold loop (`protocol_migrate.go:339`) from running ahead of that point.

## Timeline

Live is lagging tip by *k*. The engine has caught up and is folding at tip.

1. Live processes ledger **L**. A protocol contract **C** is deployed there; live's classifier matches the wasm and buffers a `protocol_contracts` row.
2. Live CASes the protocol cursor expecting `L-1`. It finds `L+k`. Lost.
3. The repair fires (`cursor >= L`, `ingest_live.go:448`/`:455`) and heals **ledger L only** — its candidate set is "contracts classified *this* ledger".
4. Live commits. C's row becomes visible.
5. The engine's next window-start membership refresh picks C up and folds it from roughly `tip+1` onward.

**Ledgers L+1 .. L+k are folded by nobody for C.**

## Why it is unbounded and deterministic

The missed range equals live's lag at the moment it processed L. No timing luck is involved — if live is behind when it reaches a deploy ledger, that many ledgers are lost.

It also does not self-limit. The engine hands off only when **its own** CAS fails (`protocol_migrate.go:598`), which requires live to win a ledger, which cannot happen while live is behind. So the engine keeps folding at tip for the entire duration of the lag.

## Impact

- **History rows** — never exist for L+1..L+k. Cursor-passed ledgers are never replayed, so there is no in-band path that ever creates them.
- **Snapshot / current-state columns** — self-heal on C's next on-chain write. Cost is staleness bounded by write frequency.
- **Additive fold columns** (cost basis, lifetime claimed emissions, cumulative counters) — permanently wrong by the sum of the missed deltas, for the lifetime of the row. A later fold adds onto a stale base; no snapshot ever corrects it.

The commit message for the window-start refresh fix in #657 makes exactly this point about a *single* missed ledger ("additive fold columns never heal from a missed ledger, so those deltas were permanently lost"). Here it is *k* of them.

This is not a regression introduced by #657 — before the repair the loss included the deploy ledger too. #657 strictly improves it. The issue is that the remaining span is unbounded and currently undocumented.

## Proposed fix: gate the engine on live's classification frontier

**Invariant:** the engine must not fold a ledger whose classifications are not yet committed.

**Available proof:** `latest_ledger_cursor` is advanced by `UpdateGuarded` (`ingest_live.go:266`) inside the **same transaction** as the `protocol_contracts` insert (`ingest_live.go:238`). So `latest_ledger_cursor >= seq` is atomic proof that every classification discoverable at `seq` is committed and visible.

Gate the fold on it — wait before `GetLedger(seq)` (`protocol_migrate.go:361`) until live's cursor reaches `seq`, read lazily and cached the way `cachedTip` already is.

Properties:

- The engine can never fold a ledger whose membership it cannot see, so the *k*-ledger gap cannot form. Correct by construction rather than detected after the fact.
- Free during historical backfill: live's cursor is far ahead and the gate never binds. It binds only when the engine is about to fold ahead of live — precisely the unsound case. (With no live history at all, membership is empty and the run produces nothing anyway.)
- It replaces the timing argument in #657's window-start refresh ("`GetLedger` blocks until the ledger closes, so the read lands after live's commit") with a happens-before that also holds when live lags tip and `GetLedger` returns straight from history.
- Live then always wins the frontier CAS, so the engine hands off cleanly instead of racing past it.

### Scope

This is an **engine-level** change and cannot be scoped down:

- **Not to one strategy.** `protocolMigrateEngine` is shared and parameterized (`protocol_migrate.go:71`); history and current state hold independent cursors and each races live separately. Fixing one leaves the other exposed.
- **Not to ContractData processors.** `contractsByProtocol[t.protocolID]` is passed to every tracker's input (`protocol_migrate.go:433`) regardless of `RequiresContractData()`, and event-only trackers never refresh membership at all (`protocol_migrate.go:549` early-returns) — so they are the worse-off case, not the exempt one.

Two things to weigh before implementing:

- **The strategies feel it differently.** History starts at `oldest_ledger_cursor` (`protocol_migrate_history.go:91`), so live is far ahead for essentially the whole run and the gate rarely binds. Current state starts at a configured `StartLedger` (`protocol_migrate_current_state.go:90`); if that is set near tip, the gate is load-bearing from the start and will change run pacing.
- **It introduces a new coupling.** Neither strategy reads `latest_ledger_cursor` today. The dependency already exists implicitly — the engine consumes membership only live produces — but making it explicit is a design decision, not a mechanical fix.

### Residual after the fix

A window whose membership snapshot predates classifications landing inside it: bounded by `WindowSize`, and confined to the band where the engine is within one window of live. That is the case the #657 repair and the window-start refresh already cover. Gate plus repair leaves no unbounded hole.

## Not a solution

Detect-and-log. An out-of-band warning that requires an operator to notice and run a targeted re-migration is not a correctness mechanism — the guarantee has to hold in-band.

## References

- #657 — adds `repairClassificationGap` (`ingest_live.go:388`) and the window-start membership refresh
- `internal/services/ingest_live.go:238`, `:266`, `:388`, `:448`, `:455`
- `internal/services/protocol_migrate.go:339`, `:361`, `:433`, `:548`, `:598`

aditya1702:
### Framework: `ContractDataChanges` + capability gate

First of 5 stacked PRs adding Blend Capital v2 lending support.

**TL;DR:** Two things. (1) Protocol processors can now receive ContractData ledger-entry changes, extracted with a footprint gate instead of a reader. (2) The migrate engine now folds a ledger **only after live ingestion has committed it** — one ordering rule that makes every "the writer didn't see this ledger's classifications" bug impossible by construction. Fixes #680. The in-transaction repair path this PR previously carried is gone: with the gate, there is nothing left to repair.

#### ContractData plumbing

- `RequiresContractData() bool` on `ProtocolProcessor` and a `ContractDataChanges map[string][]ingest.Change` field on `ProtocolProcessorInput`. SEP-41 returns `false` — event-only protocols are unaffected.
- **Extractor** walks a ledger's successful transactions and groups every ContractData ledger-entry change by owning contract C-address, preserving tx application order (last-write-wins folding per entry key stays deterministic). Per-tx entry removals surface as `Post == nil`; ledger-level archival evictions are not surfaced. Contract-id encode errors fail fast instead of silently dropping changes. Synthesized transactions carry the reader-identical `Hash` (`TxProcessing[i].Result.TransactionHash`), since `GetChanges` attaches the transaction to every change it returns.
- **Footprint gate, no reader:** Soroban guarantees writes ⊆ the declared read-write footprint, so a skim of the already-decoded envelopes' footprints skips ledgers that touch no tracked contract outright (protocol-migrate previously paid ~7ms/ledger building a `LedgerTransactionReader` — SHA-256 of every envelope — on every ledger of the range); a hit extracts directly from transaction meta. A fixture corpus over real pubnet ledgers (169 grouped changes across 5 ledgers, one pre-CAP-46-11 classic-account-owned entry documented as skipped) pins meta-based output ≡ the reader-based reference on the change fields, and that every changed contract, tracked alone, triggers the gate.
- **Live ingestion** (§2.6) extracts lazily and memoized from the transactions the staging pass already materialized — at most once per ledger, only when a CAS-winning processor requires it; a protocol still backfilling costs nothing. `RequiresContractData()` processors receive the FULL committed protocol membership (`GetByProtocolID`), not just this ledger's event emitters: entries can change without events, and event decoding disambiguates against the full tracked set.

#### The frontier gate (replaces the repair)

The problem: live ingestion is the sole classifier — `protocol_contracts` rows commit inside live's per-ledger transaction, together with live's cursor. The engine snapshots membership from committed rows only. So any ledger the engine folded **before live committed it** could be missing contracts classified at that very ledger, and cursor-passed ledgers are never replayed. That's a permanent, silent gap (#680), and the CAS-race variant of it is what the previous revision's `repairClassificationGap` healed one ledger at a time.

The fix: the engine's fold loop now waits until `latest_ingest_ledger >= seq` before fetching ledger `seq`. Live's cursor advances in the same transaction as the ledger's classifications, so `cursor >= seq` is atomic proof that every classification discoverable at `seq` is committed and visible.

What that buys, all by construction:

- **Complete membership everywhere.** The gate closes every open window whenever `seq` crosses the last known frontier value — before it even reads the new one — so no window ever contains a ledger past the cursor value it opened under, which is exactly the bound the window-start membership refresh covers (`flushWindow` no-ops when nothing is pending, so this is free in bulk and the intended shrink-to-one cadence at the frontier). Membership refreshes per window for **every** tracker, event-only ones included: their processors filter events by membership too, so a run-start-only snapshot had the same silent-drop failure through a different door.
- **The repair is deleted, not fixed.** A live CAS loss now only ever means "the engine is behind and will fold this ledger later, seeing everything this transaction commits" or "restart replay of an already-committed ledger". The winner-folded-without-membership case cannot occur, so `repairClassificationGap`, its M1–M4 test scenarios, and `IngestStore.GetInTx` are gone (~490 lines). This also closes the previously documented mid-window residual gap — it required the engine running ahead of live.
- **Clean handoff.** The gate's flush commits each cursor to the last folded ledger before waiting, so live's `CAS(cursor → cursor+1)` matches the moment it reaches the frontier. The gate subsumes the old tip-flush (`flushWindowsAtTip`): the live cursor never exceeds the chain tip, so the frontier flush always fires first, and the engine's only `GetLatestLedgerSequence` call disappears.
- **Cost:** one cached comparison per ledger during bulk; a logged 1s poll only at the frontier. If live is down, the engine idles at live's cursor instead of folding ledgers whose classifications aren't committed — waiting is the correct behavior, folding ahead is exactly the #680 bug.

Also in this PR: the protocol-state metric's Help text no longer lists a `load_current_state` phase that nothing emits.

**Deviations register (reviewer sign-off):**
| # | Item |
|---|---|
| 10 | PR adds a method to a public interface — in-repo implementers enumerated and compile/vet-verified: sep41 `processor`, `ProtocolProcessorMock`, `testRecordingProcessor`, plus `testProtocolProcessor` (`ingest_test.go`, discovered via `go vet`) |

### DIFF
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -528,20 +528,30 @@ func ExtractContractEventsForLedger(ledgerMeta xdr.LedgerCloseMeta) (map[Contrac
 
 // ExtractContractDataChangesForLedger walks a ledger's successful transactions
 // and returns every ContractData ledger-entry change grouped by the owning
-// contract's C-address strkey. Unlike ExtractContractEventsForLedger it uses
+// contract's C-address strkey. It materializes the ledger's transactions via
 // the LedgerTransactionReader (GetChanges needs envelope↔meta pairing), so it
-// takes a ctx and the network passphrase and is strictly heavier — callers
-// gate it behind ProtocolProcessor.RequiresContractData.
-//
-// Within a contract, changes preserve transaction application order, so
-// last-write-wins folding per entry key is deterministic. Ledger-level
-// archival evictions are NOT surfaced (GetChanges only walks fee/tx/op meta);
-// per-tx entry removals appear with Post == nil.
+// is strictly heavier than ExtractContractEventsForLedger — callers gate it
+// behind ProtocolProcessor.RequiresContractData. Callers that already hold
+// the ledger's transactions (live ingestion's main pipeline) use
+// ExtractContractDataChangesFromTransactions directly instead of paying for
+// a second reader build.
 func ExtractContractDataChangesForLedger(ctx context.Context, networkPassphrase string, ledgerMeta xdr.LedgerCloseMeta) (map[string][]ingest.Change, error) {
 	transactions, err := GetLedgerTransactions(ctx, networkPassphrase, ledgerMeta)
 	if err != nil {
 		return nil, fmt.Errorf("getting transactions for ledger %d: %w", ledgerMeta.LedgerSequence(), err)
 	}
+	return ExtractContractDataChangesFromTransactions(transactions, ledgerMeta.LedgerSequence())
+}
+
+// ExtractContractDataChangesFromTransactions is
+// ExtractContractDataChangesForLedger over already-materialized transactions;
+// ledgerSeq is used only for error context.
+//
+// Within a contract, changes preserve transaction application order, so
+// last-write-wins folding per entry key is deterministic. Ledger-level
+// archival evictions are NOT surfaced (GetChanges only walks fee/tx/op meta);
+// per-tx entry removals appear with Post == nil.
+func ExtractContractDataChangesFromTransactions(transactions []ingest.LedgerTransaction, ledgerSeq uint32) (map[string][]ingest.Change, error) {
 	out := make(map[string][]ingest.Change)
 	for i := range transactions {
 		tx := transactions[i]
@@ -550,7 +560,7 @@ func ExtractContractDataChangesForLedger(ctx context.Context, networkPassphrase 
 		}
 		changes, chErr := tx.GetChanges()
 		if chErr != nil {
-			return nil, fmt.Errorf("getting changes for ledger %d tx %d: %w", ledgerMeta.LedgerSequence(), tx.Index, chErr)
+			return nil, fmt.Errorf("getting changes for ledger %d tx %d: %w", ledgerSeq, tx.Index, chErr)
 		}
 		for _, change := range changes {
 			if change.Type != xdr.LedgerEntryTypeContractData {
@@ -575,7 +585,7 @@ func ExtractContractDataChangesForLedger(ctx context.Context, networkPassphrase 
 			if encErr != nil {
 				// Callers rely on this function returning every ContractData
 				// change; silently dropping one would corrupt downstream state.
-				return nil, fmt.Errorf("encoding contract id for ledger %d tx %d: %w", ledgerMeta.LedgerSequence(), tx.Index, encErr)
+				return nil, fmt.Errorf("encoding contract id for ledger %d tx %d: %w", ledgerSeq, tx.Index, encErr)
 			}
 			out[addr] = append(out[addr], change)
 		}
@@ -584,18 +594,22 @@ func ExtractContractDataChangesForLedger(ctx context.Context, networkPassphrase 
 }
 
 // ProcessLedger extracts transactions from a ledger and indexes them.
-// Returns the participant count for optional metrics recording.
-func ProcessLedger(ctx context.Context, networkPassphrase string, ledgerMeta xdr.LedgerCloseMeta, ledgerIndexer *Indexer, buffer *IndexerBuffer) (int, error) {
+// Returns the participant count for optional metrics recording, plus the
+// materialized transactions so callers with further per-transaction work
+// (live ingestion's ContractData extraction) can reuse them instead of
+// paying for a second LedgerTransactionReader build — its constructor
+// re-hashes every transaction envelope.
+func ProcessLedger(ctx context.Context, networkPassphrase string, ledgerMeta xdr.LedgerCloseMeta, ledgerIndexer *Indexer, buffer *IndexerBuffer) (int, []ingest.LedgerTransaction, error) {
 	ledgerSeq := ledgerMeta.LedgerSequence()
 	transactions, err := GetLedgerTransactions(ctx, networkPassphrase, ledgerMeta)
 	if err != nil {
-		return 0, fmt.Errorf("getting transactions for ledger %d: %w", ledgerSeq, err)
+		return 0, nil, fmt.Errorf("getting transactions for ledger %d: %w", ledgerSeq, err)
 	}
 
 	participantCount, err := ledgerIndexer.ProcessLedgerTransactions(ctx, transactions, buffer)
 	if err != nil {
-		return 0, fmt.Errorf("processing transactions for ledger %d: %w", ledgerSeq, err)
+		return 0, nil, fmt.Errorf("processing transactions for ledger %d: %w", ledgerSeq, err)
 	}
 
-	return participantCount, nil
+	return participantCount, transactions, nil
 }
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -556,17 +556,21 @@ func ExtractContractDataChangesForLedger(ledgerMeta xdr.LedgerCloseMeta, tracked
 		if !resultPair.Result.Successful() {
 			continue
 		}
-		// A minimal LedgerTransaction: GetChanges touches only these fields
-		// (verified by the fixture equivalence test against the reader-based
-		// reference) — the envelope is deliberately absent. Result pairs and
-		// TxApplyProcessing share application order, the same index alignment
-		// ExtractContractEventsForLedger relies on.
+		// A minimal LedgerTransaction: GetChanges touches only these fields —
+		// the envelope is deliberately absent, while Hash mirrors the reader's
+		// value because GetChanges attaches this transaction to every Change
+		// it returns. The fixture equivalence test compares the change fields
+		// (type, reason, operation index, pre/post entries) against the
+		// reader-based reference; the attached transaction is not compared.
+		// Result pairs and TxApplyProcessing share application order, the same
+		// index alignment ExtractContractEventsForLedger relies on.
 		tx := ingest.LedgerTransaction{
 			Index:         uint32(i + 1),
 			Result:        resultPair,
 			UnsafeMeta:    ledgerMeta.TxApplyProcessing(i),
 			LedgerVersion: ledgerMeta.ProtocolVersion(),
 			Ledger:        ledgerMeta,
+			Hash:          resultPair.TransactionHash,
 		}
 		if err := collectContractDataChanges(&tx, ledgerSeq, out); err != nil {
 			return nil, err
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -526,31 +526,103 @@ func ExtractContractEventsForLedger(ledgerMeta xdr.LedgerCloseMeta) (map[Contrac
 	return out, nil
 }
 
-// ExtractContractDataChangesForLedger walks a ledger's successful transactions
-// and returns every ContractData ledger-entry change grouped by the owning
-// contract's C-address strkey. It materializes the ledger's transactions via
-// the LedgerTransactionReader (GetChanges needs envelope↔meta pairing), so it
-// is strictly heavier than ExtractContractEventsForLedger — callers gate it
-// behind ProtocolProcessor.RequiresContractData. Callers that already hold
-// the ledger's transactions (live ingestion's main pipeline) use
-// ExtractContractDataChangesFromTransactions directly instead of paying for
-// a second reader build.
-func ExtractContractDataChangesForLedger(ctx context.Context, networkPassphrase string, ledgerMeta xdr.LedgerCloseMeta) (map[string][]ingest.Change, error) {
-	transactions, err := GetLedgerTransactions(ctx, networkPassphrase, ledgerMeta)
-	if err != nil {
-		return nil, fmt.Errorf("getting transactions for ledger %d: %w", ledgerMeta.LedgerSequence(), err)
+// ExtractContractDataChangesForLedger returns every ContractData ledger-entry
+// change from a ledger's successful transactions, grouped by the owning
+// contract's C-address strkey — but only after a cheap footprint gate: if no
+// transaction's read-write footprint contains a ContractData key owned by a
+// tracked contract, the ledger is skipped outright (empty map). Soroban
+// guarantees writes ⊆ the declared read-write footprint (host storage is
+// footprint-seeded and a write outside it traps; RestoreFootprint restores
+// exactly the read-write keys; protocol-23 auto-restores are indices into it),
+// so the gate can never skip a ledger that actually changes a tracked
+// contract's entries.
+//
+// Both the gate and the extraction walk the already-decoded LedgerCloseMeta
+// directly — GetChanges reads only the transaction meta, result, and ledger
+// version, so no LedgerTransactionReader (which re-hashes every envelope just
+// to pair envelopes with metas) is ever built. When the gate passes, the FULL
+// ledger's ContractData changes are returned, not just the tracked subset:
+// the map is shared across protocols and each processor filters by its own
+// membership.
+func ExtractContractDataChangesForLedger(ledgerMeta xdr.LedgerCloseMeta, trackedContracts map[xdr.ContractId]struct{}) (map[string][]ingest.Change, error) {
+	if !ledgerTouchesTrackedContractData(ledgerMeta, trackedContracts) {
+		return map[string][]ingest.Change{}, nil
 	}
-	return ExtractContractDataChangesFromTransactions(transactions, ledgerMeta.LedgerSequence())
+
+	ledgerSeq := ledgerMeta.LedgerSequence()
+	out := make(map[string][]ingest.Change)
+	for i := 0; i < ledgerMeta.CountTransactions(); i++ {
+		resultPair := ledgerMeta.TransactionResultPair(i)
+		if !resultPair.Result.Successful() {
+			continue
+		}
+		// A minimal LedgerTransaction: GetChanges touches only these fields
+		// (verified by the fixture equivalence test against the reader-based
+		// reference) — the envelope is deliberately absent. Result pairs and
+		// TxApplyProcessing share application order, the same index alignment
+		// ExtractContractEventsForLedger relies on.
+		tx := ingest.LedgerTransaction{
+			Index:         uint32(i + 1),
+			Result:        resultPair,
+			UnsafeMeta:    ledgerMeta.TxApplyProcessing(i),
+			LedgerVersion: ledgerMeta.ProtocolVersion(),
+			Ledger:        ledgerMeta,
+		}
+		if err := collectContractDataChanges(&tx, ledgerSeq, out); err != nil {
+			return nil, err
+		}
+	}
+	return out, nil
 }
 
-// ExtractContractDataChangesFromTransactions is
-// ExtractContractDataChangesForLedger over already-materialized transactions;
-// ledgerSeq is used only for error context.
-//
-// Within a contract, changes preserve transaction application order, so
-// last-write-wins folding per entry key is deterministic. Ledger-level
-// archival evictions are NOT surfaced (GetChanges only walks fee/tx/op meta);
-// per-tx entry removals appear with Post == nil.
+// ledgerTouchesTrackedContractData reports whether any transaction in the
+// ledger declares a read-write footprint ContractData key owned by a tracked
+// contract. It walks envelopes in whatever order the ledger stores them —
+// fine for a ledger-level boolean, which is exactly why the gate is
+// per-ledger rather than per-transaction: envelopes are in tx-set order while
+// metas are in application order, and matching the two is the expensive
+// pairing this function exists to avoid.
+func ledgerTouchesTrackedContractData(ledgerMeta xdr.LedgerCloseMeta, trackedContracts map[xdr.ContractId]struct{}) bool {
+	if len(trackedContracts) == 0 {
+		return false
+	}
+	for _, env := range ledgerMeta.TransactionEnvelopes() {
+		var ext xdr.TransactionExt
+		switch env.Type {
+		case xdr.EnvelopeTypeEnvelopeTypeTx:
+			ext = env.V1.Tx.Ext
+		case xdr.EnvelopeTypeEnvelopeTypeTxFeeBump:
+			ext = env.FeeBump.Tx.InnerTx.V1.Tx.Ext
+		default:
+			// V0 envelopes predate Soroban: no footprint, no ContractData.
+			continue
+		}
+		sorobanData, ok := ext.GetSorobanData()
+		if !ok {
+			continue
+		}
+		for _, key := range sorobanData.Resources.Footprint.ReadWrite {
+			if key.Type != xdr.LedgerEntryTypeContractData {
+				continue
+			}
+			contractID, ok := key.ContractData.Contract.GetContractId()
+			if !ok {
+				continue
+			}
+			if _, tracked := trackedContracts[contractID]; tracked {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// ExtractContractDataChangesFromTransactions extracts the same
+// contract-grouped ContractData changes as ExtractContractDataChangesForLedger
+// over already-materialized transactions, ungated — live ingestion's main
+// pipeline has the transactions in hand and processes one ledger per close,
+// so a footprint gate buys nothing there. ledgerSeq is used only for error
+// context.
 func ExtractContractDataChangesFromTransactions(transactions []ingest.LedgerTransaction, ledgerSeq uint32) (map[string][]ingest.Change, error) {
 	out := make(map[string][]ingest.Change)
 	for i := range transactions {
@@ -558,39 +630,53 @@ func ExtractContractDataChangesFromTransactions(transactions []ingest.LedgerTran
 		if !tx.Result.Successful() {
 			continue
 		}
-		changes, chErr := tx.GetChanges()
-		if chErr != nil {
-			return nil, fmt.Errorf("getting changes for ledger %d tx %d: %w", ledgerSeq, tx.Index, chErr)
-		}
-		for _, change := range changes {
-			if change.Type != xdr.LedgerEntryTypeContractData {
-				continue
-			}
-			entry := change.Post
-			if entry == nil {
-				entry = change.Pre
-			}
-			if entry == nil {
-				continue
-			}
-			contractData, ok := entry.Data.GetContractData()
-			if !ok {
-				continue
-			}
-			contractIDBytes, ok := contractData.Contract.GetContractId()
-			if !ok {
-				continue
-			}
-			addr, encErr := strkey.Encode(strkey.VersionByteContract, contractIDBytes[:])
-			if encErr != nil {
-				// Callers rely on this function returning every ContractData
-				// change; silently dropping one would corrupt downstream state.
-				return nil, fmt.Errorf("encoding contract id for ledger %d tx %d: %w", ledgerSeq, tx.Index, encErr)
-			}
-			out[addr] = append(out[addr], change)
+		if err := collectContractDataChanges(&tx, ledgerSeq, out); err != nil {
+			return nil, err
 		}
 	}
 	return out, nil
+}
+
+// collectContractDataChanges appends tx's ContractData changes into out,
+// grouped by the owning contract's C-address strkey.
+//
+// Within a contract, changes preserve transaction application order, so
+// last-write-wins folding per entry key is deterministic. Ledger-level
+// archival evictions are NOT surfaced (GetChanges only walks fee/tx/op meta);
+// per-tx entry removals appear with Post == nil.
+func collectContractDataChanges(tx *ingest.LedgerTransaction, ledgerSeq uint32, out map[string][]ingest.Change) error {
+	changes, chErr := tx.GetChanges()
+	if chErr != nil {
+		return fmt.Errorf("getting changes for ledger %d tx %d: %w", ledgerSeq, tx.Index, chErr)
+	}
+	for _, change := range changes {
+		if change.Type != xdr.LedgerEntryTypeContractData {
+			continue
+		}
+		entry := change.Post
+		if entry == nil {
+			entry = change.Pre
+		}
+		if entry == nil {
+			continue
+		}
+		contractData, ok := entry.Data.GetContractData()
+		if !ok {
+			continue
+		}
+		contractIDBytes, ok := contractData.Contract.GetContractId()
+		if !ok {
+			continue
+		}
+		addr, encErr := strkey.Encode(strkey.VersionByteContract, contractIDBytes[:])
+		if encErr != nil {
+			// Callers rely on this function returning every ContractData
+			// change; silently dropping one would corrupt downstream state.
+			return fmt.Errorf("encoding contract id for ledger %d tx %d: %w", ledgerSeq, tx.Index, encErr)
+		}
+		out[addr] = append(out[addr], change)
+	}
+	return nil
 }
 
 // ProcessLedger extracts transactions from a ledger and indexes them.
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -573,7 +573,9 @@ func ExtractContractDataChangesForLedger(ctx context.Context, networkPassphrase 
 			}
 			addr, encErr := strkey.Encode(strkey.VersionByteContract, contractIDBytes[:])
 			if encErr != nil {
-				continue
+				// Callers rely on this function returning every ContractData
+				// change; silently dropping one would corrupt downstream state.
+				return nil, fmt.Errorf("encoding contract id for ledger %d tx %d: %w", ledgerMeta.LedgerSequence(), tx.Index, encErr)
 			}
 			out[addr] = append(out[addr], change)
 		}
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -11,6 +11,7 @@ import (
 	"github.com/alitto/pond/v2"
 	set "github.com/deckarep/golang-set/v2"
 	"github.com/stellar/go-stellar-sdk/ingest"
+	"github.com/stellar/go-stellar-sdk/strkey"
 	"github.com/stellar/go-stellar-sdk/support/log"
 	"github.com/stellar/go-stellar-sdk/xdr"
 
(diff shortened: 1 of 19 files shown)


internal/services/ingest_live.go
```diff
@@ -128,13 +135,23 @@ func (m *ingestService) persistLedgerData(
...
-					LedgerCloseTime:   ledgerCloseTime,
-					ContractEvents:    contractEvents,
-					ProtocolContracts: contracts,
-					StagingMode:       StagingModeBoth,
+					LedgerSequence:      ledgerSeq,
+					LedgerCloseTime:     ledgerCloseTime,
+					ContractEvents:      contractEvents,
+					ProtocolContracts:   contracts,
+					StagingMode:         StagingModeBoth,
+					ContractDataChanges: contractDataChanges,
```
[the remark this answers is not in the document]
aditya1702: Fixed in aeaef32d: for `RequiresContractData()` processors the live path now resolves the protocol's full committed membership via `GetByProtocolID` (same as the migration engine) instead of the event-derived subset, with the buffered same-ledger overlay unchanged. Covered by a new CAS-gating subtest: a classified contract emitting no events reaches the requiring processor's `ProtocolContracts`; event-only processors keep the cheaper event-derived membership.

internal/services/protocol_migrate.go
```diff
@@ -369,16 +370,32 @@ func (s *protocolMigrateEngine) processAllProtocols(ctx context.Context, protoco
...
+		// will fold this ledger requires them; all trackers share the map.
+		var ledgerContractDataChanges map[string][]ingest.Change
+		if trackersRequireContractData(trackers, seq) {
+			cdStart := time.Now()
+			var cdErr error
+			ledgerContractDataChanges, cdErr = indexer.ExtractContractDataChangesForLedger(ctx, s.networkPassphrase, ledgerMeta)
+			cdDur := time.Since(cdStart)
+			timers.extract += cdDur
+			s.metrics.PhaseDuration.WithLabelValues("extract").Observe(cdDur.Seconds())
+			if cdErr != nil {
```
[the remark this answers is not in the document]
aditya1702: Fixed in 1849ed96: the ContractData extraction now observes under a distinct `extract_contract_data` phase label, keeping "extract" at one observation per ledger. The wall-clock timers log still aggregates both (it reports total extraction time).

internal/indexer/indexer.go
```diff
@@ -462,6 +463,61 @@ func ExtractContractEventsForLedger(ledgerMeta xdr.LedgerCloseMeta) (map[Contrac
...
+				continue
+			}
+			contractIDBytes, ok := contractData.Contract.GetContractId()
+			if !ok {
+				continue
+			}
+			addr, encErr := strkey.Encode(strkey.VersionByteContract, contractIDBytes[:])
+			if encErr != nil {
+				continue
+			}
```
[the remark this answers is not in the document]
aditya1702: Fixed in 1849ed96: encode failure now returns a wrapped error instead of skipping the change. (Unreachable in practice for a 32-byte contract id, but fail-fast is the right contract for a completeness-critical extractor.)

internal/services/ingest_live.go
```diff
@@ -128,13 +161,38 @@ func (m *ingestService) persistLedgerData(
 					continue
 				}
 
-				contracts := getEffectiveProtocolContracts(protocolID, committedByProtocol[protocolID], bufferedContracts, classification)
+				committed := committedByProtocol[protocolID]
+				if processor.RequiresContractData() {
+					if !contractDataExtracted {
+						var cdErr error
+						contractDataChanges, cdErr = indexer.ExtractContractDataChangesFromTransactions(transactions, ledgerSeq)
```
JiahuiWho: Should we hoist this to `ingestProcessedDataWithRetry` (or `ingestLiveLedgers`) and cache the result across retry attempts, instead of recomputing it inside each one?

The extraction only depends on `transactions`, which doesn't change between attempts, so retries redo the same work inside an open transaction, and only when the DB is already struggling.

internal/services/ingest_live.go
```diff
@@ -128,13 +161,38 @@ func (m *ingestService) persistLedgerData(
...
+					}
+
+					// ContractData-driven processors need the protocol's FULL committed
+					// membership, not just this ledger's event emitters: entries can
+					// change on a contract that emitted no event this ledger, and event
+					// decoding may disambiguate against tracked contracts that appear
+					// only in another contract's topics. Protocols requiring contract
+					// data have bounded membership, so the per-ledger query stays cheap.
+					var fullErr error
+					committed, fullErr = m.models.ProtocolContracts.GetByProtocolID(ctx, protocolID)
```
JiahuiWho: Should `GetByProtocolID` run on `dbTx` instead of the pool? Seems like a cheap swap since the connection is already in hand.

Right now it's called inside the open persist transaction but executes on `m.DB`, so it grabs a second pool connection while already holding one. Fine today because live ingestion is single-threaded, but it's the classic pool-saturation deadlock pattern if this path ever runs concurrently.

aristidesstaffieri:
**Race at the migration/live frontier: a contract classified in the ledger being folded can permanently lose its deploy-ledger ContractData**

This is latent today (the only in-tree processor is event-only — `sep41/processor.go` returns `RequiresContractData() == false`), but the seam this PR builds arms it the moment a ContractData-requiring processor (e.g. the Blend pool processor) registers. Flagging it now since the fix probably belongs in this layer.

**The asymmetry**

Both producers of protocol state need contract membership, because ContractData extraction is footprint-gated: `ExtractContractDataChangesForLedger(ledgerMeta, tracked)` skips any transaction whose footprint touches no tracked contract. If a contract isn't in `tracked` when its ledger is folded, its entries are never extracted at all.

- Live ingestion builds an *effective* membership: committed rows **plus a same-ledger overlay** (`getEffectiveProtocolContracts` in `ingest_live.go`), so a contract deployed and classified in the ledger being processed is included even though its `protocol_contracts` row isn't committed yet.
- The migration engine has no overlay. It uses a snapshot of committed rows, refreshed only **after** a window commits (`refreshTrackerContracts`). While folding a ledger, the snapshot is whatever the last refresh saw — and it can never see live's in-flight classification, because that insert sits uncommitted inside live's open transaction.

**The failure timeline**

Say a Blend-style processor tracks pools P1, P2, and migration has caught up to the frontier (cursor = 1999, window size 1 via `flushWindowsAtTip`):

1. Ledger 2000 closes containing a factory deploy of pool **P3**. The constructor writes P3's instance entry and config storage — some of these keys are written exactly once, here.
2. Live ingestion opens its tx for 2000: classifier stages the `protocol_contracts` insert for P3; overlay membership = {P1, P2, P3}; it will attempt `CAS(1999 → 2000)`.
3. Migration concurrently folds 2000 with snapshot {P1, P2}. The deploy tx's footprint touches only P3 → the gate in `trackedContractIDSet` / `ExtractContractDataChangesForLedger` skips it → **P3's constructor writes are never staged.** Migration's `flushWindow` wins the CAS and persists ledger 2000 without them.
4. Live loses its CAS → skips staging (`continue` in `ingest_live.go`) — but the `protocol_contracts` BatchInsert is outside the CAS gate, so P3's classification still commits.
5. Migration refreshes → membership now includes P3, effective from ledger **2001**. Ledger 2000 is cursor-passed and settled; no mechanism replays it.

End state: P3 is a tracked contract, its post-2001 changes flow normally, but its deploy-ledger entries exist on chain and nowhere in the DB. Never-rewritten keys (immutable pool config) never heal; keys a later ledger rewrites heal in current-state but stay missing from history. No error is raised anywhere.

The race is confined to the frontier — below the tip, a contract deployed at tip T has no footprint in older windows, and membership refreshes pick it up long before migration reaches T. But "migration catches up to tip while a new pool deploys via the permissionless factory" is exactly the window this handoff design exists for.

**Possible fixes**

Refreshing harder can't close this — committed-row reads can't see an uncommitted classification. The options change who owns the deploy ledger:

1. **Live-side repair (cheapest, targeted):** when live loses the CAS for ledger L but classified new contracts in L for a ContractData-requiring protocol, it knows the winner folded L with a membership that couldn't include them — and it still holds the deploy transactions. Extract and persist just those contracts' ledger-L entries as a supplemental write in that losing-race case.
2. **Migration-side replay:** after a frontier window commit, diff membership across the refresh; for contracts that appeared, re-fetch the just-committed ledgers and process ContractData for just those contracts. Correct but re-fetches dropped ledgers and bends the "cursor-passed = settled" invariant.
3. **Gate classification behind the CAS** so the ledger's owner classifies — but migration doesn't run the classifier, so this balloons.

Related: `contractsByProtocol` being a shared mutable map that every flush call site must remember to refresh is what makes this area fragile — moving membership onto `protocolTracker` (so refresh is tracker-local) would pair well with whichever fix lands.

internal/services/ingest_live.go
aditya1702: Fixed in 692809a7. Extraction now lives in a memo created once per ledger in `ingestProcessedDataWithRetry`, shared across attempts — a retry reuses the first attempt's result instead of re-extracting inside the open tx. Still lazy: it only runs when a CAS-winning processor needs ContractData.

internal/services/ingest_live.go
aditya1702: Fixed in dd2cee53. `GetByProtocolID` now takes a `db.Querier` (same pattern as `GetSACContractsMissingMetadata`); this call passes `dbTx`, the migrate-path callers pass the pool.

internal/indexer/indexer.go
```diff
@@ -525,19 +526,176 @@ func ExtractContractEventsForLedger(ledgerMeta xdr.LedgerCloseMeta) (map[Contrac
...
+		// reference) — the envelope is deliberately absent. Result pairs and
+		// TxApplyProcessing share application order, the same index alignment
+		// ExtractContractEventsForLedger relies on.
+		tx := ingest.LedgerTransaction{
+			Index:         uint32(i + 1),
+			Result:        resultPair,
+			UnsafeMeta:    ledgerMeta.TxApplyProcessing(i),
+			LedgerVersion: ledgerMeta.ProtocolVersion(),
+			Ledger:        ledgerMeta,
+		}
```
aristidesstaffieri, line 574: **`Hash` is available here for free**

`GetChanges()` attaches this `LedgerTransaction` to every `Change` it returns (`changes[i].Transaction = t`), so processors reach it via `ContractDataChanges`. `Hash` is currently left zero, but it's already in hand — `resultPair.TransactionHash` is the exact field the reader reads for its own `Hash` (`lcm.TransactionHash(i)` → `TxProcessing[i].Result.TransactionHash`).

Costs nothing to set, and saves a future processor from a surprise:

```suggestion
		tx := ingest.LedgerTransaction{
			Index:         uint32(i + 1),
			Result:        resultPair,
			UnsafeMeta:    ledgerMeta.TxApplyProcessing(i),
			LedgerVersion: ledgerMeta.ProtocolVersion(),
			Ledger:        ledgerMeta,
			Hash:          resultPair.TransactionHash,
		}
```

internal/services/processor_registry_test.go
```diff
@@ -26,6 +26,13 @@ func registerStub(id string, base int64) {
 	})
 }
 
+func TestProtocolProcessor_RequiresContractData(t *testing.T) {
+	m := NewProtocolProcessorMock(t)
+	m.On("RequiresContractData").Return(false).Once()
+	var p ProtocolProcessor = m
+	assert.False(t, p.RequiresContractData())
+}
```
aristidesstaffieri: **This only exercises the mock**

`m.On("RequiresContractData").Return(false)` then asserting `false` tests testify's plumbing — no production code is on that path, so it would stay green if `RequiresContractData` were removed from every real processor or if both engines stopped consulting it.

The gate is already covered behaviorally by the tests added later in this PR: `protocol_migrate_test.go:1136`/`:1289` (populated vs. left nil) and `ingest_test.go` cases H/I/J. Suggest dropping this one — it's also the only non-registry test in `processor_registry_test.go`.

If the intent was to pin that the mock satisfies the interface, that's a compile-time assertion rather than a test:

```go
// mocks.go
var _ ProtocolProcessor = (*ProtocolProcessorMock)(nil)
```

internal/indexer/indexer.go
aditya1702: Set in 110b80f7, mirroring the reader's value. Also scoped the comment above the literal — the fixture equivalence test projects `Transaction` away (`projectContractDataChanges` compares only the change fields), so its guarantee never covered this field.

internal/services/processor_registry_test.go
aditya1702: Dropped in cccc76c4. The compile-time assertion already existed (`mocks.go`'s `var _ ProtocolProcessor = (*ProtocolProcessorMock)(nil)`), so nothing replaced it.

aditya1702:
Re: the migration/live frontier race (https://github.com/stellar/wallet-backend/pull/657#issuecomment-4996896051) — implemented option 1 (live-side repair) in ba7ebfb4.

On a lost protocol-cursor swap where this ledger classified new contracts for the protocol, live now checks the lost cursor's committed value in-transaction (new `IngestStore.GetInTx`): at or past this ledger is proof the winner folded it without those contracts, so live re-stages scoped to exactly the gap contracts (already-committed contracts excluded — the winner had those in membership) and persists the lost halves in the same transaction, without moving any cursor. Partial losses repair only the lost half at the matching staging mode. Verification detail that makes the plain read sufficient: a CAS that lost to a winner already past the ledger blocked on the winner's row lock before re-evaluating, so the winner's commit is always visible by the time the repair decision reads the cursor.

Two deltas from the write-up:
- The trigger is `cursor >= L`, not `== L`: when live lags a migration that already folded past the deploy ledger, the same repair heals the deploy ledger itself (the never-rewritten constructor keys — the permanent part of the loss). History rows for later ledgers inside the winner's already-folded range remain the engine's, per the residual below.
- The footprint gate is per-ledger, not per-transaction, which narrows the window but a first-deploy-for-a-protocol ledger is exactly the all-untracked case.

Accepted residual (documented on `repairClassificationGap`): an engine mid-window across the deploy ledger at the moment live's classification commits still folds it with the older membership — that needs live to lag the engine by a full window with the deploy landing mid-window; current-state heals on the contract's next write, that window's history rows stay missing. Closing it needs engine-side replay (your option 2) — deferred with the tracker-local membership refactor.

Covered by CAS-gating cases M1–M4: frontier repair, behind-tip non-repair, partial-loss scoped repair, already-committed exclusion; mutation-verified.

aditya1702:
Follow-up on the frontier race — one more timing crack closed, in 0f0692a3.

The engine refreshed membership immediately after each window commit. At the contested ledger L that refresh fires milliseconds after the engine's commit — but live's losing transaction (which carries the new contract's classification) commits *after* that, since its CAS was lock-blocked behind the engine and it still has the repair and the rest of its persist to run. So the engine could stage L+1 with a snapshot still missing the contract, and if it won L+1 too, live's repair wouldn't fire there (the contract is no longer newly-classified at L+1). One ledger of the contract's events could slip through — and the additive fold columns (cost basis, lifetime claimed) never heal from a missed ledger, unlike the snapshot columns.

The refresh now runs at window start, after `GetLedger` returns for the window's first ledger: the fetch blocks until that ledger has closed, a full ledger interval after any concurrent live transaction for the previous one committed. Same once-per-window cadence, read taken at the latest useful moment. The refresh test now pins the ordering (first folded ledger must carry post-snapshot membership; removing the window-start refresh fails it).

The mid-window residual from the previous comment is unchanged — a window staged over many ledgers still uses one snapshot; that one needs engine-side replay.

aristidesstaffieri:
Re: the classification-gap repair (https://github.com/stellar/wallet-backend/pull/657#issuecomment-5218309830) — I filed #680 for the residual in the `cursor > L` branch.

The `>=` trigger covers two different cases.

`cursor == L` is a race between live and the engine. The repair fixes it completely.

`cursor > L` is not a race. Live ingestion is the only classifier: `ingest_live.go:238` is the only writer of `protocol_contracts`, and the engine reads committed rows only. The engine cannot see a contract until live reaches its deploy ledger.

If live lags the tip by k ledgers:

1. Live processes ledger L and classifies contract C.
2. Live tries the CAS with expected value L-1. The stored value is L+k. The CAS fails.
3. The repair runs because `cursor >= L`. It heals ledger L only.
4. Live commits. The engine sees C at its next refresh and folds C from about tip+1.

Ledgers L+1 to L+k are folded by nobody. The gap equals live's lag. This is deterministic, not timing dependent.

The gap does not close by itself. The engine hands off only when its own CAS fails (`protocol_migrate.go:598`), and live cannot win a CAS while it lags.

Impact:
- History rows for L+1 to L+k never exist.
- Snapshot columns heal at the next write of C.
- Additive fold columns (cost basis, lifetime claimed) stay wrong permanently. The missed deltas are not recoverable.

This is not a regression. The repair improves the previous behavior. #680 has the proposed fix and the scope analysis.

internal/services/ingest_live.go
```diff
@@ -246,6 +279,237 @@ func (m *ingestService) persistLedgerData(
...
+	}
+	start := time.Now()
+	if processErr := processor.ProcessLedger(ctx, input); processErr != nil {
+		return fmt.Errorf("processing classification-gap repair at ledger %d for protocol %s: %w", ledgerSeq, protocolID, processErr)
+	}
+	m.appMetrics.Ingestion.ProtocolStateProcessingDuration.WithLabelValues(protocolID, "process_ledger").Observe(time.Since(start).Seconds())
+
+	if repairHistory {
+		persistStart := time.Now()
+		if persistErr := processor.PersistHistory(ctx, dbTx); persistErr != nil {
```
aristidesstaffieri: `repairClassificationGap` can abort live ingestion with a primary key violation.

The chain:

1. `state_changes` has `PRIMARY KEY (ledger_created_at, to_id, operation_id, state_change_id)` (`2025-06-10.4-statechanges.sql:49`).
2. The repair writes history for a ledger whose cursor it lost. The winner already wrote history for that same ledger.
3. Both producers derive `to_id` and `operation_id` from `toid.New(ledgerSeq, txIdx, opIdx+1)` (`sep41/processor.go:127`), and both use the same `ledger_created_at`.
4. `AssignStateChangeOrdinals` restarts its counter at 1 for each `(ToID, OperationID)` pair and adds the protocol base (`types.go:822`). The counter is local to one staged slice.
5. The first row each producer stages for a shared operation gets `base+1`.
6. `BatchCopy` uses `pgxTx.CopyFrom` with no `ON CONFLICT` (`statechanges.go:158`).
7. SQLSTATE 23505. `isPermanentPersistError` treats class 23 as permanent (`ingest_live.go:1006`), so there is no retry. Live ingestion stops on that ledger.

Precondition: one operation emits events from both a gap contract and a contract in the winner's membership. A Soroban transaction normally carries one `InvokeHostFunction` operation, so all its contract events share one `opID`. A factory that deploys a token with a minting constructor and moves an existing token in the same transaction meets this.

`PersistCurrentState` is safe. Balance and allowance keys include the contract ID, and `processEvent` filters on the emitting contract (`sep41/processor.go:153`), so repair rows and winner rows never share a key.

Root cause: the CAS guaranteed one producer per (ledger, operation). Ordinal assignment depends on that guarantee. The repair writes outside it. The doc comment's "additive across disjoint membership sets" covers row content, not the ID namespace.

Fix — the pattern already exists in `types.go`. The indexer splits its 2^40 namespace at bit 28 into 4096 sub-streams. Protocol emitters are documented as single-stream, with a note that a future multi-stream emitter subdivides its own namespace the same way. The repair is that second stream. Give it a distinct sub-base inside the protocol's namespace. Repair rows then cannot collide with winner rows. This is deterministic, needs no extra query, and needs no coordination between producers.

Smaller alternative: drop the history half of the repair and keep current state only. That removes the collision. It also removes the deploy-ledger history rows, which is the main reason the repair exists.

This finding is independent of #680. The gate proposed there removes the `cursor > L` case. The `cursor == L` case still runs the repair, so the collision survives that fix.

internal/services/ingest_live.go
```diff
@@ -246,6 +279,237 @@ func (m *ingestService) persistLedgerData(
...
+	// contract), so the winner staged their state for this ledger itself.
+	rawIDs := make([][]byte, 0, len(candidates))
+	for _, c := range candidates {
+		raw, decErr := hex.DecodeString(string(c.ContractID))
+		if decErr != nil {
+			return fmt.Errorf("decoding contract id %q for protocol %s classification-gap check at ledger %d: %w", string(c.ContractID), protocolID, ledgerSeq, decErr)
+		}
+		rawIDs = append(rawIDs, raw)
+	}
+	committedRows, err := m.models.ProtocolContracts.BatchGetByContractIDs(ctx, rawIDs)
```
aristidesstaffieri: `GetByProtocolID` was moved to `db.Querier` in dd2cee53, so this call looks like the one that was missed. Making the same change here would silently disable the repair.

The reason is the write ordering documented at `:123-126`: a processor's `Persist*` inserts name-enriched `protocol_contracts` rows before the generic insert at `:238`. In the partial-loss case:

1. `stageAndPersistProtocolLedger` persists the won half for protocol P.
2. The gap contract C is in that processor's membership, through the `getEffectiveProtocolContracts` overlay.
3. An enriching processor writes C's row into `dbTx`.
4. On `dbTx`, this query returns C. `alreadyTracked` then drops C, `gap` is empty, and the repair returns at `:439`.

The pool connection does not see the uncommitted row, so C stays in `gap` and the repair runs. No in-tree processor enriches yet, so both paths agree today. Blend is the expected first one.

Two suggestions:

- State in the comment that the read must not see this transaction's own writes, and that `dbTx` would suppress the repair.
- Add an M5 case: partial loss where the won half's processor inserts a `protocol_contracts` row for the gap contract, asserting the repair still persists. That turns the comment into a guard.

JiahuiWho's pool-saturation point still stands as written, and is unchanged by this: single-threaded today, and this read is behind the `len(candidates) == 0` return at `:409`.

internal/services/ingest_live.go
```diff
@@ -246,6 +279,237 @@ func (m *ingestService) persistLedgerData(
...
+		ContractEvents:      contractEvents,
+		ProtocolContracts:   gap,
+		StagingMode:         mode,
+		ContractDataChanges: contractDataChanges,
+	}
+	start := time.Now()
+	if processErr := processor.ProcessLedger(ctx, input); processErr != nil {
+		return fmt.Errorf("processing classification-gap repair at ledger %d for protocol %s: %w", ledgerSeq, protocolID, processErr)
+	}
+	m.appMetrics.Ingestion.ProtocolStateProcessingDuration.WithLabelValues(protocolID, "process_ledger").Observe(time.Since(start).Seconds())
```
aristidesstaffieri: The repair reuses the `process_ledger`, `persist_history`, and `persist_current_state` phase labels.

In the partial-loss case (M3), `stageAndPersistProtocolLedger` observes `process_ledger` at `:337` and the repair observes it again at `:492`. One protocol emits two samples for one ledger.

This is the same problem Copilot raised about `extract`. The fix in 1849ed96 added a distinct `extract_contract_data` label, described as "keeping `extract` at one observation per ledger". The repair reintroduces the pattern on three labels.

There is a second effect. When both halves are lost (M1) only the repair runs, and its samples are indistinguishable from normal staging. Repair activity cannot be counted, graphed, or alerted on. Given #680, repair frequency is a useful signal.

Suggested fix, matching the 1849ed96 precedent:

- Observe the repair under distinct phases: `repair_process_ledger`, `repair_persist_history`, `repair_persist_current_state`.
- Update the Help text at `metrics/ingestion.go:138`. It enumerates the valid phase values.

Separately, the `Observe` calls sit after the error returns here (`:492`, `:498`, `:506`). In `stageAndPersistProtocolLedger` they sit before (`:337`, `:345`, `:353`), so a failed call is still recorded. A failed repair records no duration. Please match the existing path.

internal/services/ingest_live.go
aditya1702: Validated first — the collision was real: both streams numbered the same (to_id, operation_id) group from the same base, COPY has no ON CONFLICT, and class-23 errors are treated as permanent, so live ingestion would have died on the 23505.

Fixed by deletion instead of the sub-base. f7d09123 gates the engine's fold loop on live's committed cursor (fold ledger L only once `latest_ingest_ledger >= L`). Live commits classifications in the same transaction as that cursor, so the CAS winner now always folds with complete membership — the case the repair existed to heal (winner already past a ledger whose classification committed with live's losing transaction) can no longer occur. `repairClassificationGap`, its M1–M4 scenarios, and `GetInTx` are gone (~490 lines), and with them the second ordinal stream: back to exactly one writer per (protocol, half, ledger), single-stream per emitter as types.go documents. Also fixes #680 — same root cause, and the gate closes the previously documented mid-window residual gap too.

internal/services/ingest_live.go
aditya1702: One correction while validating: the repair's read was `BatchGetByContractIDs` (pool-only, no Querier parameter) — there was no `GetByProtocolID` call in the repair, so the dd2cee53 convention didn't apply to it. Your substance held though: a dbTx read there would have seen this transaction's own enriched rows and silently suppressed the repair.

Moot now — the call site was deleted with the repair (see the previous thread): under the frontier gate a CAS loss can only mean engine-behind (it folds the ledger later, seeing everything this transaction commits) or restart replay, so there is no already-tracked check left to guard and no M5 scenario to test.

internal/services/ingest_live.go
aditya1702: The repair phases went away with the repair itself (previous threads), which resolves both the double process_ledger sample in the partial-loss case and the observe-after-error placement — that asymmetry was real.

Two details from validating: `extract_contract_data` (1849ed96) lives on `wallet_ingestion_phase_duration_seconds`, a different metric than this one — precedent still applies, wrong histogram. And this metric's Help listed a `load_current_state` phase that nothing emits; dropped in 0aadb510. Your observability point survives in the new shape: repair frequency is gone as a signal, replaced by the gate's `frontier_gate` phase samples plus gated-wait logs — both fire exactly when the engine reaches live's frontier.

internal/services/protocol_migrate.go
```diff
@@ -446,41 +515,96 @@ func (s *protocolMigrateEngine) refreshTargetTip(ctx context.Context) {
...
+	if seq <= cachedLiveCursor {
+		return cachedLiveCursor, nil
+	}
+	liveCursor, err := s.ingestStore.Get(ctx, data.LatestLedgerCursorName)
+	if err != nil {
+		return cachedLiveCursor, fmt.Errorf("reading live ingestion cursor before folding ledger %d: %w", seq, err)
 	}
-	if seq <= cachedTip {
-		return cachedTip, nil
+	if seq <= liveCursor {
```
aristidesstaffieri: looks like since the membership is only refresh at the start of the window we can exceed the cursor value inside of the window.

Example

windowSize = 100. The engine is converging on live.

1. Engine opens a window at ledger 500. Live's committed cursor is 504. The window-start refresh snapshots membership: pool X is absent.
2. Engine folds 500-504 off the cached cursor.
3. Live commits ledger 505. That ledger deploys and classifies X. The classification and latest_ledger_cursor = 505 commit together.
4. Live CASes the protocol cursor expecting 504. The engine's committed value is 499. Live loses and takes the continue at ingest_live.go:222. It stages nothing.
5. Engine reaches seq = 505. 505 <= 504 is false, so it takes the fresh read: liveCursor = 505. 505 <= 505 is true, so :552 returns. No flush, no refresh.
6. Engine folds 505-599 with the ledger-500 snapshot. X is not in trackedContractIDSet, so the footprint gate skips its transactions and its ContractData is never extracted. X's events are filtered out of
ProcessLedger.
7. The window CAS(499 -> 599) succeeds. Ledgers 505-599 are cursor-passed.

We can flush before the read so that no window can extend past the cursor value. Adds a flush per cursor read but I think its needed for this case.

aditya1702: Confirmed and fixed in 13faa81c — your trace was exact, including live's lost CAS at step 4 leaving the engine as the only writer for the range.

**What changed:**
1. `gateOnLiveFrontier` now flushes every open window whenever `seq` crosses the cached frontier, **before** the fresh read — your suggested placement. No window can contain a ledger past the cursor value it opened under, which is exactly the bound the window-start refresh covers. `flushWindow` already no-ops on `pending == 0`, so a bulk-chase crossing (once per cache refill) costs one committed window and frontier crossings are the intended shrink-to-one cadence.
2. While closing this, the same hole existed through a second door: `refreshTrackerContracts` early-outed for event-only processors (run-start snapshot forever), so a SEP-41-style contract classified mid-run stayed invisible until handoff even with your fix. Every tracker now refreshes on the window-start cadence.

**Regression test** (`a window never outlives the frontier value that opened it`) replays your scenario with an event-only processor: live commits the classifying ledger mid-window → asserts the crossing flushed the open window as its own CAS'd commit, ledgers before the crossing folded without X, and every ledger after folded WITH X. Mutation-verified: restoring the fresh-read early return fails it at exactly your step 5.

aristidesstaffieri (approved):

PR #659, opened 2026-07-08
feat(blend): BLEND validator, entry/event decoders, and processor

aditya1702:
### BLEND validator, entry/event decoders, and processor

Third of 5 stacked PRs adding Blend Capital v2 lending support (stacked on #658). Adds the ingestion layer between the ledger stream and #658's writer models: a **validator** (classifies pools/backstop by WASM interface), **entry decoders** (ContractData changes → current-state snapshots), **event decoders** (contract events → Blend state-change history rows + additive folds), and the **processor** that stages both and persists in dependency order. Every shape was verified against the deployed v2 WASMs (mainnet and testnet are byte-identical) and `blend-capital/blend-contracts-v2 @ ba22b487`.

Also: the SEP-41 validator's WASM-spec matcher is extracted verbatim into shared `internal/services/wasmspec` (mechanical; SEP-41 unchanged), and everything registers via `init()` + blank imports in `protocol-setup` / `protocol-migrate` / `ingest`. `mocks.go` (deferred from PR2's deadcode gate) lands here with its consumer, the processor tests.

### Validator

Classification is purely interface-based (all required functions must match), exercised by real-WASM fixture tests:

| Pool (6 fns) | Backstop (7 fns) |
|---|---|
| `get_config`, `get_reserve`, `get_positions`, `submit`, `flash_loan`, `bad_debt` | `deposit`, `queue_withdrawal`, `dequeue_withdrawal`, `withdraw`, `claim`, `draw`, `donate` |

Pool matches are enriched best-effort via RPC `get_config` (a failed call never fails validation). Pool names have no on-chain getter — the processor fills them from instance storage.

**Canonical-backstop pin.** Backstop-derived tables key rows by pool/user with no backstop contract id, so an impostor deployed from the real backstop WASM could overwrite genuine rows. Backstop-shaped entries (`UserBalance`/`PoolBalance`/`BEmisData`/`UEmisData`/`RZ`) and events (`deposit`/`queue_withdrawal`/`dequeue_withdrawal`/`withdraw`/backstop `claim`) are therefore folded only from the canonical backstop per network (addresses from docs.blend.capital + blend-utils, WASM hashes verified on-chain; skips get one debug log per contract per window). Junk *pools* stay tolerated — they key rows under their own contract ids.

| Network | Canonical backstop | BLND SAC |
|---|---|---|
| Pubnet | `CAQQR5SWBXKIGZKPBZDH3KM5GQ5GUTPKB7JAFCINLZBC5WXPJKRG3IM7` | `CD25MNVTZDL4Y3XBCPCJXGXATV5WUHHOWMYFF4YBEGU5FCPGMYTVG5JY` |
| Testnet | `CBDVWXT433PRVTUNM56C3JREF3HIZHRBA64NB2C3B2UNCKIS65ZYCLZA` | `CB22KRA3YZVCNCQI64JQ5WE7UY2VAV7WFLK6A2JN3HEX56T2EDAFO7QF` |

### Entry decoders

Twelve ContractData kinds, routed by disjoint key shape; anything else decodes to `KindIgnored` (never an error). Removal and `Option::None` (`ScvVoid`) are explicit nil-payload cases.

| Kind | Key | Durability | Feeds |
|---|---|---|---|
| PoolInstance | instance storage (must contain `Config`) | instance | `blend_pools` — config + `Admin`/`Name` (best-effort) |
| Positions | `Positions(user)` | persistent | `blend_positions` — supplied / collateral / borrowed |
| ResConfig / ResData | `ResConfig(asset)` / `ResData(asset)` | persistent | `blend_reserves` — curve config / rates, supplies, backstop credit |
| EmisData / UserEmis | `EmisData(u32)` / `UserEmis(user, reserve_id)` | persistent | `blend_reserve_emissions` / `blend_emissions` (`token_id >= 0`) |
| BackstopUserBalance / BackstopPoolBalance | `UserBalance(pool, user)` / `PoolBalance(pool)` | persistent | `blend_backstop_positions` / `blend_backstop_pools` |
| BackstopBEmisData / BackstopUEmisData | `BEmisData(pool)` / `UEmisData(pool, user)`, `Option<…>` | persistent | `blend_backstop_pools` emission state / `blend_emissions` (`token_id = -1`) |
| Auction | `Auction(auct_type, user)` | **temporary** | `blend_auctions` |
| RewardZone | bare symbol `"RZ"` | persistent | `blend_pools.in_reward_zone` |

- `Auction` encodings verified against a live mainnet auction; the `AuctionKey` ScMap sorts `auct_type` before `user` (symbol-byte order, reverse of Rust field order). An empty `RZ` vec is a valid live value. An instance entry without `Config` is the backstop's own → ignored.
- Consciously ignored bookkeeping keys (no per-entity, user-facing state): pool `ResList`/`PoolEmis`/`PropAdmin`/`ResInit`; backstop `LastDist`/`DropList`/`BackfillEmis`/`Backfill`/`RzEmis`/`PoolUSDC`.

### Event decoders

Each row carries a (category, reason) pair under the schema-wide convention — category names the on-chain object (`BLEND_SUPPLY`, `BLEND_DEBT`, …), reason the action — plus `key_value.poolId` (where the event names a pool) and `token_id` = reserve asset, BLND, or NULL. Unrecognized symbols decode to nothing; a malformed recognized event errors → logged at Warn, counted in `wallet_ingestion_protocol_decode_failures_total{protocol_id, kind}` (probable contract drift), and skipped, never aborting the ledger. Decode validation rejects `auct_type` above int32 range and `fill_percent` outside 1..100 (the contract enforces ≤ 100 on-chain).

| Event | Row (category, reason) | Fold |
|---|---|---|
| `supply` / `supply_collateral` | (BLEND_SUPPLY, CREDIT) / (BLEND_COLLATERAL, CREDIT) | `net_supplied +=` |
| `withdraw_collateral` | (BLEND_COLLATERAL, DEBIT) | `net_supplied -=` |
| `borrow` / `repay` | (BLEND_DEBT, BORROW) / (BLEND_DEBT, REPAY) | `net_borrowed +=` / `-=` |
| `flash_loan` | (BLEND_DEBT, FLASH_LOAN) | `net_borrowed +=` (a same-tx repay nets it out) |
| `bad_debt` | (BLEND_DEBT, BAD_DEBT) | user's `net_borrowed` reset to 0 |
| `defaulted_debt` | (BLEND_DEBT, BURN) | — (attributed to the pool address) |
| `fill_auction` | (BLEND_AUCTION, FILL) — user + filler rows | per auction type, below |
| `withdraw` (ambiguous) | (BLEND_SUPPLY, DEBIT) *or* (BLEND_BACKSTOP, DEBIT) | `net_supplied -=` (pool side only) |
| `claim` (ambiguous) | (BLEND_EMISSIONS, CLAIM) *or* (BLEND_BACKSTOP_EMISSIONS, CLAIM) | lifetime claimed totals |
| `deposit` / `queue_withdrawal` / `dequeue_withdrawal` | (BLEND_BACKSTOP, CREDIT) / (BLEND_BACKSTOP_QUEUE, ADD) / (BLEND_BACKSTOP_QUEUE, REMOVE) | — |

- **`withdraw`**: pool and backstop shapes are identical; disambiguated by topic[1] — a backstop `withdraw`'s is a tracked Blend pool, a pool `withdraw`'s is a reserve asset (never Blend-tracked).
- **`claim`**: pool = tuple data, backstop = bare `i128`; the category alone now encodes the source. A backstop claim pays Comet LP (BLND auto-swapped + re-deposited), so its row is `token_id` NULL + `units:"backstop_lp"` with no pool address.
- **`fill_auction`** folds mirror on-chain Positions moves: type 0 (UserLiquidation) lot+bid both sides; type 1 (BadDebt) bid both sides; type 2 (Interest) none.

### Processor

`RequiresContractData()` = true. `StateChangeOrdinalBase()` returns the reserved Blend `state_change_id` namespace base, and `PersistHistory` assigns deterministic ordinals in emission order via `types.AssignStateChangeOrdinals` — Blend rows can never collide with the main indexer's or SEP-41's IDs for the same operation. Event groups are folded in on-chain `(TxIdx, OpIdx)` order (not map order): the cost-basis folds accumulate across groups into shared staged state and `bad_debt`'s reset makes them order-sensitive, so a borrow and a bad_debt for one key in different groups of the same ledger must fold in chain order — pinned by a probabilistic regression test. Events stage history rows + additive folds; entries stage last-write-wins snapshots; a `Persist` without a following `Reset()` refuses further folds (no double-count). Persist order: **pools → reserves** (fold SQL resolves asset → `reserve_index` via `blend_reserves`; a window that decoded a reserve's config without its data goes through a config-only partial update — `BatchUpdateConfig` — so a `set_reserve` invariant violation degrades to a stale data half, never zeroed live rates) **→ positions** (delete removed → zero absent → upsert snapshots → net-deltas → auction adjustments) **→ backstop/emission/claimed/auction sets** (independently keyed) **→ reward zone last** (flips `in_reward_zone` on rows the pools upsert just created). Pool claims fold per emitting pool; backstop claims fold account-wide. Auctions stage LWW per `(pool, user, type)` — created-then-filled in one window nets to a delete. Batch equivalence (one- vs two-window folds, removal+recreate, same-ledger cross-group ordering) is covered by processor tests.

### Notes for reviewers

- **Auction FILL rows**: `token_id`/`amount` NULL (multi-asset); per-asset detail in `key_value.lot`/`bid` (exposed as first-class lists in #661); non-reserve assets no-op via the reserves join.
- **`defaulted_debt`**: backstop-side basis not folded (not user-facing).
- **Scalars pinned**: `b_rate`/`d_rate` 1e12, `ir_mod` 1e7, backstop emissions 1e14; the pool reserve-emission scalar is **per-reserve** (`10^decimals`) — PR5 math must join `ReserveConfig.decimals`.
- **Backstop claim auto-restake**: a claim also emits one genuine `deposit` event per pool, so one claim op yields 1 backstop-emissions CLAIM + N backstop CREDIT rows (Σ deposits == claim amount); no double-count — covered by a dedicated processor test.
- **Unknown networks** degrade to NULL `token_id` on pool-claim rows rather than a misattributed token.
- **Known limitation — windowed-migration auction basis**: a multi-ledger migration window values its accumulated auction lot/bid deltas at end-of-window rates; the divergence vs per-ledger ingestion is ≤ borrow APR × window/year (~1e-5 at 50% APR for the default 100-ledger window), affects only `interestEarned`/`interestPaid`, and is one-time (live ingestion runs window=1, exact).
- **Known limitation — evicted auctions**: change reads are tx-meta-only, so a temporary `Auction` entry evicted by TTL leaves a stale `blend_auctions` row. Accepted: auction economics make an unfilled expiry near-impossible.
- **Operational caveat — backstop swap**: if the emitter ever swaps the canonical backstop, the pin must be updated and backstop-derived state migrated.


internal/services/blend/events.go
```diff
@@ -0,0 +1,682 @@
...
+// pool address for a backstop withdraw. BLEND only ever tracks pool and
+// backstop contracts, so a reserve asset is never isTracked — checking
+// isTracked(topic[1]) reliably tells the two apart.
+func decodeWithdrawAmbiguous(topics []xdr.ScVal, data xdr.ScVal, poolID string, isTracked func(addr string) bool) (*DecodedEvent, error) {
+	addr1, from, err := decodeTwoAddrTopics(topics, "withdraw")
+	if err != nil {
+		return nil, err
+	}
+
+	if isTracked(addr1) {
```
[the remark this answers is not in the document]
aditya1702: Fixed at the root in PR #657 (aeaef32d): the live path now passes the protocol's FULL committed membership to `RequiresContractData()` processors (via `GetByProtocolID`, matching the migration engine) rather than this ledger's event emitters. With the complete tracked set, `isTracked(topic[1])` is a sound discriminator: the pool in a backstop withdraw's topics is always classified/tracked, while a pool withdraw's topic[1] is a reserve asset that is never a BLEND contract. Same-ledger-deployed pools are covered by the buffered-contracts overlay.

internal/services/blend/processor.go
```diff
@@ -0,0 +1,1035 @@
...
+			WasmHash:   wasmHash,
+			Name:       sp.Name,
+		})
+	}
+
+	if err := p.pools.BatchUpsert(ctx, dbTx, rows); err != nil {
+		return fmt.Errorf("upserting %d blend pools for ledger %d: %w", len(rows), p.ledgerNumber, err)
+	}
+	if len(nameRows) > 0 && p.protocolContracts != nil {
+		if err := p.protocolContracts.BatchInsert(ctx, dbTx, nameRows); err != nil {
```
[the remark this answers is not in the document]
aditya1702: Fixed in a628a91a (PR #657): the live path now persists this ledger's protocol_wasms rows BEFORE the processor block, so a processor's name-enriched protocol_contracts insert passes the FK filter even when the contract deploys in the same ledger as its wasm upload. The generic contracts insert stays after the processors — its COALESCE on name preserves the enriched value.

internal/services/blend/processor.go
```diff
@@ -0,0 +1,1035 @@
...
+	sb.ledger = p.ledgerNumber
+}
+
+// stageBackstopBEmisData merges (LWW) a backstop pool's emission half into
+// its staged row, independently of any balance half already staged. A Void
+// (Option::None) payload is nil per DecodedEntry's contract and is treated
+// as nothing-to-stage, same as a removal.
+func (p *processor) stageBackstopBEmisData(decoded DecodedEntry) {
+	if decoded.EmisData == nil {
+		return
```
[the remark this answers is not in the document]
aditya1702: Not fixing — the transition can't occur on-chain. In blend-contracts-v2 @ ba22b487, `set_backstop_emis_data` takes `&BackstopEmissionData` (not `Option`), so `None` can never be STORED — the `Option` in the getter only models a storage miss. And `backstop/src/storage.rs` contains zero `remove()` calls on persistent storage, so Some→None and entry-removal transitions don't exist for this key. The nil-payload early return is defensive against a state the contract cannot produce; staging a 'cleared' marker would add dead complexity.

internal/services/blend/processor.go
```diff
@@ -0,0 +1,1035 @@
...
+	sb.ledger = p.ledgerNumber
+}
+
+// stageBackstopUEmis replaces (LWW) the staged user backstop-emission
+// accrual, keyed with Source = the backstopped pool (from decoded.Pool) and
+// TokenID = blenddata.BackstopEmissionTokenID. A Void payload is nil and
+// treated as nothing-to-stage, same as stageBackstopBEmisData.
+func (p *processor) stageBackstopUEmis(decoded DecodedEntry) {
+	if decoded.UserEmis == nil {
+		return
```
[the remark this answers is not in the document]
aditya1702: Not fixing — same reasoning as the BEmisData thread: `set_user_emis_data` takes `&UserEmissionData` (never `Option`), and the backstop contract has no `remove()` calls on persistent storage, so a stored `UEmisData` can never transition to None nor be removed. Stale-on-None is unreachable; the nil-payload return is purely defensive.

internal/services/blend/validator.go
```diff
@@ -0,0 +1,450 @@
...
+		if !ok {
+			log.Ctx(ctx).Debugf("blend: skipping pool contract with undecodable id %q", ct.ContractID)
+			continue
+		}
+		if _, dup := seen[addr]; dup {
+			continue
+		}
+		seen[addr] = struct{}{}
+
+		row, err := v.fetchPoolConfig(ctx, addr)
```
JiahuiWho: This makes RPC calls (get_config, one per pool, inside this loop) while dbTx is open, the pattern #669 just removed from the SEP-41 validator by splitting `Validate` into `Match`/`Prefetch`/`Apply`. 

When this stack rebases onto that interface, the fetch loop belongs in `Prefetch` and the `BatchUpsert` in `Apply`.

internal/services/blend/processor.go
```diff
@@ -0,0 +1,1331 @@
...
+				row.IRMod = sr.data.IRMod
+				row.BackstopCredit = sr.data.BackstopCredit
+				row.LastTime = int64(sr.data.LastTime)
+			} else {
+				// ResData wasn't observed alongside this window's new/changed
+				// ResConfig. In practice set_reserve always writes both entries
+				// together, so this is defensive: zero the rate/supply fields
+				// rather than sending Go's zero-value "" (an invalid numeric
+				// string) on a full-row overwrite.
+				row.BRate, row.DRate, row.BSupply, row.DSupply, row.IRMod, row.BackstopCredit = "0", "0", "0", "0", "0", "0"
```
JiahuiWho: If this branch ever fires for an existing reserve (config-only window), the full-row upsert overwrites its real rates/supplies with these zeros, and it breaks batch equivalence: data-update in window 1 + config-only in window 2 clobbers what window 1 wrote, while the same ledgers in one window don't. Since `set_reserve` writing both entries together is the invariant, would a config-only partial-update path (mirror of `BatchUpdateData`) be safer? Then a future violation degrades to a stale column instead of zeroed live data.

internal/services/blend/processor.go
```diff
@@ -0,0 +1,1331 @@
...
+		// tx walk would have produced (mirrors sep41's processor).
+		txID := toid.New(int32(input.LedgerSequence), int32(key.TxIdx), 0).ToInt64()
+		opID := toid.New(int32(input.LedgerSequence), int32(key.TxIdx), int32(key.OpIdx+1)).ToInt64()
+		opBuilder := processors.NewStateChangeBuilder(input.LedgerSequence, input.LedgerCloseTime, txID, nil).
+			WithOperationID(opID)
+
+		for _, event := range events {
+			if err := p.processEvent(event, opBuilder, input.StagingMode, blndToken); err != nil {
+				// Log and continue — a malformed event must not abort the whole ledger.
+				log.Debugf("blend: skipping malformed event at tx=%d op=%d: %v", key.TxIdx, key.OpIdx, err)
```
JiahuiWho: Since a recognized-symbol decode failure realistically means contract-upgrade shape drift, should this be Warn + a metric rather than Debug?

internal/services/blend/entries.go
```diff
@@ -0,0 +1,869 @@
...
+}
+
+func decodeAuction(cd xdr.ContractDataEntry, elems []xdr.ScVal, removed bool) (DecodedEntry, error) {
+	const kind = "Auction"
+	auctType, user, err := decodeAuctionKey(elems, kind)
+	if err != nil {
+		return DecodedEntry{}, err
+	}
+
+	out := DecodedEntry{Kind: KindAuction, Removed: removed, User: user, AuctionType: int32(auctType)}
```
JiahuiWho, line 654: auct_type is u32 on-chain; values ≥ 2³¹ silently go negative here. Worth rejecting (or keeping uint32) since interface-matched junk pools can store anything?

internal/services/blend/events.go
```diff
@@ -0,0 +1,765 @@
...
+	}
+	filler, ok := addrString(vec[0])
+	if !ok {
+		return nil, fmt.Errorf("blend: fill_auction: data[0] filler is not an address (type %v)", vec[0].Type)
+	}
+	fillPercentStr, ok := i128String(vec[1])
+	if !ok {
+		return nil, fmt.Errorf("blend: fill_auction: data[1] fill_percent is not i128 (type %v)", vec[1].Type)
+	}
+	fillPercent, convErr := strconv.Atoi(fillPercentStr)
```
JiahuiWho: Every other amount rides through as a decimal string. Atoi here means an oversized fill_percent from a junk pool discards the whole fill_auction (rows + folds), and it's the one key_value field typed as a number. Pass it through as a string like the rest?

internal/services/blend/validator.go
```diff
@@ -0,0 +1,450 @@
...
+		return blenddata.Pool{}, fmt.Errorf("fetching get_config: %w", err)
+	}
+	m, ok := val.GetMap()
+	if !ok || m == nil {
+		return blenddata.Pool{}, fmt.Errorf("get_config: expected map, got %v", val.Type)
+	}
+
+	// LastModifiedLedger is deliberately left at its zero value: Validate has
+	// no ledger sequence available (ValidationInput carries none). blend_pools
+	// overwrites this column unconditionally on every upsert, so it must not
```
JiahuiWho: Nit: the model does GREATEST on this column , "overwrites unconditionally" states the opposite

JiahuiWho (approved):

internal/services/blend/processor.go
```diff
@@ -0,0 +1,1331 @@
...
+	}
+	p.ledgerNumber = input.LedgerSequence
+	p.indexContracts(input.ProtocolContracts)
+
+	if len(p.blendContracts) == 0 {
+		return nil
+	}
+
+	blndToken := blndTokenAddress(p.networkPassphrase)
+	for key, events := range input.ContractEvents {
```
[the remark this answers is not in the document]
aditya1702: Good catch — real bug. History-row IDs were already order-immune, but the cost-basis folds accumulate across groups into shared state, and bad_debt's `ZeroBorrowed` is destructive: borrow + bad_debt for the same key in different groups of one ledger could persist either result. Fixed in d6962483: `ProcessLedger` sorts group keys by (TxIdx, OpIdx). Added `TestProcessLedger_FoldsEventGroupsInOpOrder` (19 interleaved groups × 8 runs; mutation-verified it fails without the sort).

internal/services/blend/processor.go
```diff
@@ -0,0 +1,1331 @@
...
+}
+
+// foldAuctionAdj accumulates fold's signed protocol-token lot/bid deltas into
+// the staged auction-adjustment accumulator for (pool, fold.User, fold.Asset).
+func (p *processor) foldAuctionAdj(pool string, fold AuctionFold) {
+	key := auctionKey{Pool: pool, User: fold.User, Asset: fold.Asset}
+	sa, ok := p.stagedAuctionAdjs[key]
+	if !ok {
+		sa = &stagedAuctionAdj{lot: new(big.Int), bid: new(big.Int)}
+		p.stagedAuctionAdjs[key] = sa
```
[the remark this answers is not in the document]
aditya1702: Accurate — documented as a bounded migration caveat in d6962483 rather than converting per-event. The error is the interest accrued between a fill and window close: ≤ borrow APR × window/year (~1e-5 at 50% APR for the default 100-ledger window), affects only the interestEarned/interestPaid display fields, and only during one-time migration — live ingestion runs window=1 and is exact. Godocs at the fold and the conversion SQL state the bound.

internal/services/blend/validator.go
aditya1702: This landed with the rebase: `Prefetch` does the RPC outside any tx and returns a plan; `Apply` is DB-only. Resolving as addressed.

internal/services/blend/processor.go
aditya1702: Implemented in d6962483: config-without-data windows now go through a new `BatchUpdateConfig` partial update (config columns only, GREATEST on last_modified_ledger, duplicate guard). An invariant violation degrades to a stale data half instead of zeroed rates (this also stops zeroing `last_time`). dbtest covers config-only-preserves-rates.

internal/services/blend/processor.go
aditya1702: Done in d6962483: both drop sites now Warn and increment `wallet_ingestion_protocol_decode_failures_total{protocol_id, kind=event|entry}` (pattern of the wasm-classification counter), wired via `ProtocolDeps.MetricsService` on all paths.

internal/services/blend/entries.go
aditya1702: Fixed in 077cbd73: `auct_type > math.MaxInt32` is now a decode error, so nothing can wrap negative into the DB. Values 3..MaxInt32 stay storable on purpose — a future protocol auction type should be stored, not dropped (GraphQL already skips types it can't name).

internal/services/blend/events.go
aditya1702: Fixed in 65aabd3a: `ParseInt` + 1..100 range check, decode error outside. Rejecting rather than passing through as string: the pool contract enforces ≤100 on-chain and events only come from wasm-validated pools, so out-of-range is junk — and the GraphQL field stays `Int!`. The integration suite now asserts FillPercent end-to-end (#664).

internal/services/blend/validator.go
aditya1702: Fixed in e8403703 — the comment now cites the model's GREATEST as the non-regression guarantee, and the stale `Validate`/`ValidationInput` references are gone.

internal/services/blend/scval.go
```diff
@@ -0,0 +1,163 @@
+// Package blend implements the BLEND v2 lending protocol's on-chain
```
aristidesstaffieri, line 1: these helpers all seem pretty general, is there any reason not to have them in utils so that they can be used by other packages?

aditya1702: Looked at moving them; holding off, for two reasons:

1. Blend is the only consumer today — all 16 helpers are unexported, and half are Blend/SEP-40-shaped rather than general (`mapU32I128` decodes the Positions map shape, `mapAddrI128` the AuctionData bid/lot maps, `decodePriceData`/`buildSep40StellarAsset` are SEP-40-specific).
2. The three other places that decode ScVals (`sep41/events.go`, `indexer/processors/contracts/utils.go` + `sac.go`, `services/checkpoint.go`) all use a `(T, error)` convention with wrapped messages, while these are deliberately `(T, bool)` SDK-style. Relocating them to `internal/utils` wouldn't delete a line at any of those call sites — it would add a fourth decoding convention to the repo.

There is real duplication worth consolidating, but it's elsewhere: `extractAddressFromScVal` exists in both sep41 and contracts, and `checkpoint.go`'s balance-entry walk is a near-copy of `sac.go`'s. That consolidation (one convention, one home) is orthogonal to this PR — happy to file a follow-up issue for it.

aristidesstaffieri (approved):

PR #661, opened 2026-07-08
feat(graphql): Blend positions, pools, and earn discovery

aditya1702:
### Blend GraphQL: positions and pools

Final stacked Blend v2 PR (on #660). Adds the GraphQL read surface for Blend lending: account positions, a pool catalog, and the Blend state-change types.

**What this adds**

- `Account.blendPositions` — the account's lending/collateral/borrowing per pool and reserve, backstop deposits with queued withdrawals, claimable BLND emissions, lifetime claimed totals, interest earned/paid to date, and any open Dutch auctions on the account.
- `Query.blendPools` / `blendPool(address)` — pool catalog: per-pool status, admin, reward-zone flag, supplied/borrowed/backstop USD, weighted APYs, and every reserve's rates, sizes, and prices. `blendPools` is a Relay connection (`first/after/last/before` → `BlendPoolConnection`), keyset-paginated by pool contract address with a page cap of 50 (also the default): pool deployment is permissionless, so the catalog is unbounded and the result set must be client-bounded — and the SQL carries a real `LIMIT`, matching every other list query's pagination shape.
- Eight concrete state-change types under the schema-wide (category = object, reason = action) convention — `BlendSupplyChange`, `BlendCollateralChange`, `BlendDebtChange` (BORROW/REPAY/FLASH_LOAN/BAD_DEBT/BURN, with per-reason amount denomination documented on the type), `BlendAuctionChange` (FILL, with `auctionType`, `fillPercent`, `counterparty`, and first-class `lot`/`bid` amount lists), `BlendEmissionsClaimChange`, `BlendBackstopEmissionsClaimChange`, `BlendBackstopChange`, and `BlendBackstopQueueChange`. Every non-null field is always meaningful for its type; unmatched (category, reason) rows surface as errors, not null nodes.
- All rate and valuation math lives in `internal/services/blend/rates.go` as pure functions, verified against `blend-contracts-v2 @ ba22b487` and `blend-sdk-js @ af62659` (details below).

**Conventions**

| Convention | Value |
|---|---|
| USD / APY fields | `Float`, null when uncomputable — a missing/stale price, an APY that compounds past float64 (permissionless `ir_mod` is unbounded; non-finite values never reach the wire), a position whose reserve row hasn't landed, or a backstop shares→LP conversion whose pool balance half hasn't been folded (`lpTokens` is nullable for the same reason); pool/position totals go null rather than silently understating when any contributor is uncomputable |
| Token amounts | `String` raw integers at the token's native decimals |
| Price staleness | prices older than 24h count as missing — same window the pool contract enforces |
| `BlendPool.status` | `BlendPoolStatus` enum; null until ingested |
| Fixed-point scales | `backstopRate`, `cFactor`, `lFactor` are 7-decimal fixed point |
| APR→APY | borrow compounds daily (365), supply weekly (52) — the SDK's convention |
| netApy (position) | `(Σ supplied·supplyApy − Σ borrowed·borrowApy) / Σ supplied`, 0 when nothing supplied — the SDK/UI convention |
| Complexity | every Blend list level carries its own bound (catalog × the client-declared `first`/`last`, resolver-capped at 50 which is also the default — the priced maximum and the executable maximum coincide; per-account lists ×10, reserves and auction bid/lot ×30, q4w ×20), and the entry fields carry their real per-invocation fan-out additively — `blendPositions` +13, `blendPool` +7 (the DB queries one resolution issues) — so nested repetition like `stateChanges → account → blendPositions` is priced by what it costs; the resolver's fan-out waves are SetLimit-capped so one request can't drain the pgx pool |

**Math audit**

Every formula was diffed against the contract source (authority) and the JS SDK (cross-check), adversarially — including edge states. Where the SDK and contract disagree, the contract wins.

| Area | Result |
|---|---|
| Interest curve, supply capture, APR→APY, utilization clamp | matches the contract; exact-rational arithmetic differs from on-chain fixed-point by <1e-5 APR, always in the contract's favor |
| Rate projection to "now" | matches `Reserve::load`, including the zero-utilization short-circuit and the unclamped bad-debt bRate accrual |
| Emission claimables | stream indexes are projected `last_time → now` exactly like the distributors (reserve: over token supply; backstop: over unqueued shares); contract test vectors pinned bit-exact |
| Emission claim divisor | `10^reserve_decimals · 1e7` per the contract — the SDK's uniform `1e14` is only right for 7-decimal tokens |
| Backstop valuation | shares→LP→USD matches; queued withdrawals stay in the headline value (still slashable first-loss capital) but earn no emissions |
| Comet LP pricing | NAV per share from actual pool weights — algebraically equal to the SDK's `usdc×5/totalShares` for the 80/20 pool |
| Debt-side rounding | underlying debt truncates instead of the contract's ceil: at most 1 stroop low, display-only |

**Notes for reviewers**

- Backstop LP / BLND prices are read only from the pinned `BLEND_BACKSTOP_LP_CONTRACT_ID` oracle group — a permissionless pool can't spoof the protocol-wide prices with a self-priced row.
- `claimedBlnd` / `backstopClaimedLp` read current-state accumulator tables, not history sums, so retention can't undercount lifetime totals.
- Full-selection complexity worst cases: `blendPools` = 26,550 at its `first: 50` cap (the connection wrapper adds edges/node/cursor/pageInfo around each node), `blendPositions` = 7,572. Admitting full selections needs a deployment-side `GRAPHQL_COMPLEXITY_LIMIT` raise (recommendation: 30,000). Existing non-Blend entries are untouched, so the freighter full-detail budget is unchanged.
- The `blendPools` cursor is the base64 pool contract address (the sort key; strkey is self-validating, so no version prefix). Pages come back in ascending `pool_contract_id` order in both directions; `pkg/wbclient.GetBlendPools` walks the cursor sequence and still returns the flat `[]BlendPool` (integration tests unchanged).
- docs.blend.capital's "rate modifier has 9 decimals" is stale v1 wording; the deployed v2 contract uses 7 decimals (`SCALAR_7`), which is what this implements.

**Deploy**

- api/serve needs `BLEND_BACKSTOP_LP_CONTRACT_ID` in its env (ingest already has it).
- `GRAPHQL_COMPLEXITY_LIMIT` raise as above.
- Run order unchanged: migrate up → restart ingest → `protocol-setup --protocol-id BLEND` → `protocol-migrate history/current-state --protocol-id BLEND --start-ledger <v2 deploy ledger>`.


internal/serve/graphql/resolvers/blend_earn.go
```diff
@@ -0,0 +1,193 @@
...
+	sortEarnPoolOptions(pools)
+
+	// contract_tokens is the canonical source for token display metadata;
+	// each reserve config carries a decimals copy set at reserve creation, so
+	// it only serves as the fallback for assets the classifier never saw.
+	meta, hasMeta := d.metaByContractID[assetID]
+	decimals := reserves[0].Decimals
+	if hasMeta {
+		decimals = int32(meta.Decimals)
+	}
```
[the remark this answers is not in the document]
aditya1702: Fixed in 27a7a2ee: metadata decimals now apply only when the contract_tokens row is genuinely enriched (name set or nonzero decimals); otherwise the reserve config's decimals — written on-chain from the token itself at reserve creation — are used. Covered by a new test with an unenriched default row.

internal/serve/graphql/resolvers/blend_positions.go
```diff
@@ -0,0 +1,736 @@
...
+		now:                        time.Now().Unix(),
+	}
+
+	// Group reserve positions by pool, preserving GetByAccount's
+	// (pool_contract_id, reserve_index) order for both pool-of-first-
+	// appearance order and each pool's reserve order.
+	positionsByPool := map[string][]blenddata.Position{}
+	poolOrder := make([]string, 0, len(poolIDSet))
+	seenPool := map[string]bool{}
+	for _, p := range positions {
```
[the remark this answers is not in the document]
aditya1702: Not fixing — the premise (position rows deleted on full exit) doesn't match the on-chain behavior. In blend-contracts-v2 @ ba22b487, `pool/src/storage.rs` has no removal path for `Positions` entries (the only `remove()` in the file is temporary-storage `del_queued_reserve_set`); a full exit empties the entry's maps, which this pipeline persists by ZEROING the rows while keeping them (cost basis intact — that's the lifetime-earned design). The zeroed rows still produce the `BlendPoolPosition` node, so `claimedBlnd` remains reachable. `DeleteByPoolUser` only fires on entry removal, which the contract never performs — it exists as a defensive path.

internal/serve/graphql/resolvers/blend_positions.go
```diff
@@ -0,0 +1,975 @@
...
+	if suppliedKnown {
+		v := suppliedUsdSum
+		suppliedUsd = &v
+	}
+	if borrowedKnown {
+		v := borrowedUsdSum
+		borrowedUsd = &v
+	}
+	if suppliedKnown && borrowedKnown {
+		v := suppliedUsdSum - borrowedUsdSum
```
JiahuiWho, line 572: `buildPoolPosition` divides net earnings by net `usdValue` while blend-sdk-js ([link](https://github.com/blend-capital/blend-sdk-js/blob/main/src/pool/user_positions_est.ts#L127)) divides by `totalSupplied`, which is what the Blend UI displays. Suggest matching the SDK?

internal/serve/graphql/schema/blend.graphqls
```diff
@@ -0,0 +1,251 @@
...
+  tokenSymbol: String
+  tokenDecimals: Int
+  suppliedTokens: String!
+  collateralTokens: String!
+  borrowedTokens: String!
+  suppliedUsd: Float
+  borrowedUsd: Float
+  supplyApy: Float
+  borrowApy: Float
+  emissionsApr: Float
```
JiahuiWho: Suggestion: split `emissionsApr` into `emissionsSupplyApr` + `emissionsBorrowApr`, mirroring BlendReserve.

Blend runs two independent emission streams per reserve, and a position can hold both sides at once (supply + borrow of the same asset). With one field, the resolver picks a side — bToken stream wins whenever supply/collateral exists — so a both-sides position silently loses its borrow-stream APR, while `emissionsEarnedBlnd` on the same type sums both streams.

internal/serve/graphql/schema/blend.graphqls
```diff
@@ -0,0 +1,251 @@
...
+  interestEarned: String!
```
JiahuiWho, line 115: Logic looks solid! Can we also add the semantics to the schema doc (lifetime, in tokens, survives full exit, liquidations don't distort it). Right now that's only discoverable by reading the resolvers.

JiahuiWho:
With the null-propagation and weighting rules, can we add totalUsdValue + netApy to BlendAccountPositions, computed in the resolver from the per-pool values?

internal/serve/graphql/schema/blend.graphqls
```diff
@@ -0,0 +1,251 @@
...
+# Dutch auctions where the account is the auction owner
+# (BlendAccountPositions.activeAuctions).
+#
+# Money-shaped fields use Float for USD/APY values (nullable wherever a missing
+# oracle price makes the value uncomputable — a genuinely zero balance is still
+# 0, not null) and String! for on-chain integer amounts, kept at full precision
+# rather than lossy float64. A stored price older than 24h counts as missing:
+# the pool contract itself refuses prices past that age.
+
+extend type Query {
```
JakeUrban: From a readability perspective it would be preferable if updates to the schema was made directly to the type definitions rather then through `extend`, so readers can view the complete type in one place.

internal/serve/graphql/schema/blend.graphqls
```diff
@@ -0,0 +1,251 @@
...
+  blendPools: [BlendPool!]!
+  blendPool(address: String!): BlendPool
```
JakeUrban: These queries make sense, but I think something we should think ahead on is whether we'll want to generalize these queries in the future through something like `lendingPools: [LendingPool!]!` for example.

Also, looking at the other queries that return lists, should we use `BlendPoolConnection`? It looks like `blendPools` isn't paginated yet.

internal/serve/graphql/schema/blend.graphqls
```diff
@@ -0,0 +1,251 @@
...
+  blendEarnOptions: [BlendEarnOption!]!
```
JakeUrban: I'm not sure whether we should support this query. Currently the wallet backend's queries answer two kinds of questions:

1. What is happening on the network generally (top-level `stateChanges`, `transactions`, and `operations` queries)
2. What is happening on my account (`accountByAddress`)

This query is neither. It is a utility downstream application could utilize, but they could also build themselves. Can we determine if this is something we can implement on freighter-backend-v2 before committing to supporting this here?

internal/serve/graphql/schema/blend.graphqls
```diff
@@ -0,0 +1,251 @@
...
+}
+
+extend type Account {
+  """An account's Blend v2 lending, collateral, borrowing, and backstop positions."""
+  blendPositions: BlendAccountPositions! @goField(forceResolver: true)
```
JakeUrban: Same question here re: generalization & pagination

internal/serve/graphql/schema/blend.graphqls
```diff
@@ -0,0 +1,251 @@
...
+pool it has touched. backstopClaimedLp is lifetime backstop-emission claims in
+Comet LP tokens (7 decimals): the backstop converts claimed BLND into its LP
+token and re-deposits it, so LP tokens are what the claimer actually receives.
+Unlike claimedBlnd (per pool), the on-chain backstop claim event carries no
+pool address, so this total can only ever be reported account-wide.
+"""
+type BlendAccountPositions {
+  pools: [BlendPoolPosition!]!
+  backstop: [BlendBackstopPosition!]!
+  backstopClaimedLp: String!
```
JakeUrban, line 31: The blend.captial dapp describes backstop positions per-pool, but it seems here the backstop is separate from the pools. Do we understand the relationship between the backstop module and the pools it backs?

internal/serve/graphql/schema/blend.graphqls
```diff
@@ -0,0 +1,251 @@
...
+type BlendPool {
+  address: String!
+  name: String
+  """
+  Raw on-chain pool status: 0 Admin Active, 1 Active, 2 Admin On-Ice,
+  3 On-Ice, 4 Admin Frozen, 5 Frozen, 6 Setup. Statuses 0-3 accept supply
+  (deposits); 0-1 also allow borrowing; 4-6 reject both. Null until the
+  pool's config entry has been ingested.
+  """
+  status: Int
```
JakeUrban: Should we have an enum for this?

internal/serve/graphql/schema/enums.graphqls
JakeUrban, line 83: I've said this before but I'm still going to mention it here since it looks like this will be the last time we can consider it without doing a formal breaking change: I'd love to refactor this part of the API so that the mapping between state change types and (category, reason) pairs was defined in the schema. Right now this mapping is unknown and has to be inferred from the processing logic.

This is likely out-of-scope of this blend-specific work, but this is the one area of the API I think we could have done a bit better one.

internal/serve/graphql/schema/statechange.graphqls
```diff
@@ -149,3 +149,18 @@ type BalanceAuthorizationChange implements BaseStateChange{
...
+  ingestedAt:                 Time!
+  ledgerCreatedAt:            Time!
+  ledgerNumber:               UInt32!
+  account:                    Account! @goField(forceResolver: true)
+  operation:                  Operation @goField(forceResolver: true)
+  transaction:                Transaction! @goField(forceResolver: true)
+
+  tokenId:                    String @goField(forceResolver: true)
+  amount:                     String @goField(forceResolver: true)
+  poolId:                     String @goField(forceResolver: true)
```
JakeUrban: Continuing my point about explicitly defining the mapping between category/reason pairs and the state change variations, its not clear how these attributes are used for each pair. For example, the schema doesn't tell me for which (category, reason) pairs `tokenId` is null or non-null.

internal/serve/graphql/resolvers/blend_positions.go
aditya1702: Matched to the SDK: `netApy` now divides by total supplied USD with the SDK's zero-supply rule (`totalSupplied == 0 → 0`, debt forgiven as bad debt) — fdb735f5. The pool-catalog `netApy` already divided by supplied USD, so only the position-level rollup changed; the schema doc now states the convention.

internal/serve/graphql/schema/blend.graphqls
aditya1702: Split into `emissionsSupplyApr` + `emissionsBorrowApr`, mirroring `BlendReserve` — 88e22b00. Both report the stream's pool-wide APR at projected rates, so a both-sides position now sees both streams. `emissionsEarnedBlnd` keeps summing both, as before.

internal/serve/graphql/schema/blend.graphqls
aditya1702: Added to the schema docs — 32d60095: lifetime, token-denominated (use `priceUsd` for USD), survives a full exit, liquidation fills adjust the cost basis so seized collateral isn't reported as earnings, may be slightly negative after an exit (dust truncation / rate movement).

internal/serve/graphql/resolvers/blend_positions.go
```diff
@@ -0,0 +1,975 @@
...
+// stream is configured and the account holds a balance, the contract's
+// "user had tokens before emissions began" branch applies (distributor.rs's
+// update_user_emissions): the holder is owed balance*index/scalar the first
+// time their emissions are touched — the full index counts, from a zero
+// starting point, with nothing pre-accrued.
+//
+// configIndex is the stream's own current emission_index (from
+// blend_reserve_emissions or blend_backstop_pools.emis_index), used as the
+// projection target even past expiration — expiry freezes further real-time
+// growth but doesn't erase the last recorded index. A nil configIndex means
```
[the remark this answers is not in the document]
aditya1702: Fixed — 60580ff5. `ProjectEmissionIndex` replicates both distributors' `update_emission_data`: `index += floor((min(now, expiration) − last_time) · eps · scalar / supply)` with the contract's guards (last_time ≥ expiration/now, zero eps, zero supply). Reserve streams project over the side's raw b/dToken supply at `10^ReserveConfig.decimals`; the backstop projects over unqueued shares (`shares − q4w`) at `1e7`. The contract's own test vectors (`test_update_emission_data_past_exp`, `_rounds_down`) are pinned bit-exact in rates_test.go. The claim conversion keeps the contract divisor `10^decimals·1e7` (blend-sdk-js's uniform `1e14` is correct only for 7-decimal reserves). A negative index delta (only reachable via an ingestion gap) is clamped, mirroring the contract's `require_nonnegative` — 1daa0c2c.

internal/data/blend/oracle_prices.go
```diff
@@ -149,3 +150,95 @@ func (m *OraclePriceModel) BatchUpsert(ctx context.Context, rows []OraclePrice)
...
+// multiply matches — each output row corresponds to exactly one (self-priced
+// row, sibling row) pair, keyed by a distinct oracle_contract_id.
+func (m *OraclePriceModel) GetBackstopLPPrices(ctx context.Context) ([]OraclePrice, error) {
+	start := time.Now()
+	const query = `
+		SELECT o2.oracle_contract_id, o2.asset_contract_id, o2.price, o2.price_decimals, o2.price_timestamp, o2.updated_at
+		FROM blend_oracle_prices o1
+		JOIN blend_oracle_prices o2 ON o2.oracle_contract_id = o1.oracle_contract_id
+		WHERE o1.oracle_contract_id = o1.asset_contract_id
+		ORDER BY o2.oracle_contract_id, o2.asset_contract_id`
```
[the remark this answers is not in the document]
aditya1702: Fixed — 5af852da. `BLEND_BACKSTOP_LP_CONTRACT_ID` (the pin the snapshot writer already targets) now flows into serve, and `GetBackstopLPPrices` filters `WHERE oracle_contract_id = $pin`; `findBackstopPrices` reduces to picking the self-priced LP row and its BLND sibling inside the pinned group (ambiguity still error-logged). A second self-priced group under any other oracle is excluded by the query — covered by a new test. An unset pin returns no rows, so backstop USD fields resolve to null. Deploy note added to the PR: the api deployment needs the env var (ingest already has it).

internal/serve/serve.go
```diff
@@ -367,4 +367,34 @@ func addComplexityCalculation(config *generated.Config) {
...
+	//   blendEarnOptions: BlendEarnOption = 4 scalars + pools(1 + 4 BlendEarnPoolOption
+	//                     scalars = 5) => childComplexity 9 => 9*50 = 450
+	//   blendPositions:   BlendAccountPositions = pools(1 + [7 scalars + reserves(1+16=17)]
+	//                     = 25) + backstop(1 + [7 scalars + q4w(1+2=3)] = 11)
+	//                     + backstopClaimedLp(1) => childComplexity 37 => 37*10 = 370
+	// All comfortably under GRAPHQL_COMPLEXITY_LIMIT=6000. None of this touches
+	// AccountTransactionEdge.operations/stateChanges or any other existing entry above —
+	// those stay exactly as budgeted for the freighter full-detail query.
+	config.Complexity.Query.BlendPools = accountsListComplexityFunc
+	config.Complexity.Query.BlendEarnOptions = accountsListComplexityFunc
```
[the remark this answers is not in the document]
aditya1702: Added per-level multipliers — 6d5ebb5c: `BlendPool.reserves` and `BlendAuction.bid`/`lot` ×30 (on-chain MAX_RESERVES), `BlendBackstopPosition.q4w` ×20 (MAX_Q4W_SIZE), `BlendAccountPositions.pools`/`backstop`/`activeAuctions` ×10 (documented assumption — no on-chain cap), catalog ×50. Full-selection worst cases become blendPools = 26,150 and blendPositions = 7,572, above the current 6,000 deployment limit by design — the PR notes call out the required limit raise. Regression tests lock each multiplier. (blendEarnOptions is gone entirely — see the thread on the query's existence.)

internal/serve/serve.go
```diff
@@ -367,4 +367,34 @@ func addComplexityCalculation(config *generated.Config) {
...
+	// Query.blendPool (single pool by address) is left at the default (1 + childComplexity):
+	// same treatment as the other single-item lookups (accountByAddress, operationByID,
+	// transactionByHash) below, which also have no custom entry. It returns one object, not
+	// a list, so no page-size-style multiplier is needed.
+	config.Complexity.Account.BlendPositions = func(childComplexity int) int {
+		return childComplexity * 10
```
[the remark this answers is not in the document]
aditya1702: Same commit (6d5ebb5c): `Account.blendPositions` itself drops back to default pricing and the cardinality moved onto the individual lists — pools/backstop/activeAuctions ×10 each, nested reserves ×30, q4w ×20, bid/lot ×30 — so multi-pool accounts and the auction fan-out are priced per level instead of one flat ×10.

internal/serve/graphql/resolvers/blend_pools.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+	price := d.priceLookup(oracle, reserve.AssetContractID)
+
+	suppliedUsd := usdValueOrNil(suppliedTokens, reserve.Decimals, price)
+	borrowedUsd := usdValueOrNil(borrowedTokens, reserve.Decimals, price)
+
+	bTokenID := reserve.ReserveIndex*2 + 1
+	dTokenID := reserve.ReserveIndex * 2
+	emissionsSupplyApr := d.emissionsAPRFor(poolAddr, bTokenID, rr.BSupply, rr.BRate, reserve.Decimals, price)
+	emissionsBorrowApr := d.emissionsAPRFor(poolAddr, dTokenID, rr.DSupply, rr.DRate, reserve.Decimals, price)
```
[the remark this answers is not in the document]
aditya1702: Fixed — c3452d06: every emissions-APR call site passes the projected `PB`/`PD`, consistent with the sibling token/USD figures on the same types.

internal/serve/graphql/resolvers/blend_earn.go
```diff
@@ -0,0 +1,278 @@
...
+	oracle := types.AddressBytea("")
+	pool, ok := d.poolByID[poolAddr]
+	if ok {
+		oracle = pool.OracleContractID
+	}
+	price := d.priceLookup(oracle, reserve.AssetContractID)
+	suppliedUsd := usdValueOrNil(suppliedTokens, reserve.Decimals, price)
+
+	bTokenID := reserve.ReserveIndex*2 + 1
+	emissionsSupplyApr := d.emissionsAPRFor(poolAddr, bTokenID, rr.BSupply, rr.BRate, reserve.Decimals, price)
```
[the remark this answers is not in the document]
aditya1702: `blendEarnOptions` was removed from the schema entirely (ea1e5ad0) — earn discovery composes from `blendPools` downstream. The projected-rate fix landed on the surviving call sites in c3452d06.

internal/serve/graphql/resolvers/blend_positions.go
```diff
@@ -0,0 +1,975 @@
...
+	bTokenID := reserve.ReserveIndex*2 + 1
+	dTokenID := reserve.ReserveIndex * 2
+	bSideBalance := new(big.Int).Add(supplyB, collateralB)
+
+	var emissionsApr *float64
+	switch {
+	case bSideBalance.Sign() > 0:
+		emissionsApr = d.emissionsAPRFor(poolAddr, bTokenID, rr.BSupply, rr.BRate, reserve.Decimals, price)
+	case liabilityD.Sign() > 0:
+		emissionsApr = d.emissionsAPRFor(poolAddr, dTokenID, rr.DSupply, rr.DRate, reserve.Decimals, price)
```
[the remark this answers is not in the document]
aditya1702: Fixed in c3452d06 — projected `PB`/`PD` in every emissions-APR denominator.

internal/serve/graphql/schema/blend.graphqls
aditya1702: Removed the query — ea1e5ad0. `blendPools` carries every field the earn view derived (pool `status`, per-reserve `enabled`, `supplyApy`, `emissionsSupplyApr`, `suppliedUsd`, token metadata/decimals), so freighter-backend-v2 can compose asset-first earn discovery client-side and wallet-backend stays out of the ranking/filtering policy.

internal/serve/graphql/schema/blend.graphqls
aditya1702: The backstop is a single protocol-wide contract that holds each pool's first-loss capital separately: deposits, queued withdrawals, and emissions are all tracked per backed pool (`PoolBalance(pool)` / `UserBalance(pool, user)` on the backstop contract). That's why `BlendBackstopPosition` carries `poolAddress` and an account gets one entry per pool it backs — the relationship is per-pool; the entries live under `backstop` rather than inside `pools` because first-loss capital (slashable, Q4W-gated) is a different product than a lending position. Added a schema-doc sentence making this explicit — 32d60095.

internal/serve/graphql/schema/blend.graphqls
aditya1702: Converted — e5497fa4: `BlendPoolStatus` enum (ADMIN_ACTIVE…SETUP) with the 0-6 on-chain encoding documented on the enum and the supply/borrow permissions on the field; null until ingested and for any unrecognized on-chain value, matching the `BlendAuctionType` precedent.

internal/serve/graphql/schema/queries.graphqls
```diff
@@ -8,4 +8,10 @@ type Query {
 
     """Look up an operation by its ID (TOID)."""
     operationById(id: Int64!): Operation
+
+    """Blend v2 pool-wide catalog: every tracked pool, independent of any account."""
+    blendPools: [BlendPool!]!
```
aristidesstaffieri: `blendPools` returns every row of `blend_pools` with no bound, and I don't think the complexity limiter can compensate for it.

This is the availability angle on top of @JakeUrban's pagination question in [this thread](https://github.com/stellar/wallet-backend/pull/661#discussion_r3625577649) — worth noting that thread went outdated when `618219ff` moved these fields out of `blend.graphqls`, so it's collapsed in the Files view and never got a reply.

Three pieces:
- **Schema** — no `first`/`after` on this line, so a client can't cap the result set.
- **Data** — `PoolModel.GetAll` (`internal/data/blend/readers.go:278`) is `SELECT ... FROM blend_pools ORDER BY pool_contract_id`, no LIMIT.
- **Pricing** — `serve.go:442` prices `Query.blendPools` at `childComplexity * 50`, a constant. The ×30 `BlendPool.reserves` multiplier from `6d5ebb5c` bounds the *nested* level; the outer row count stays unpriced.

Why the constant concerns me: Blend pool deployment is permissionless, so the row count is attacker-growable rather than a function of protocol adoption. At ~20k pools, `query { blendPools { address reserves { assetContractId suppliedUsd } } }` prices at 50 × (1 + 30×2) = 3,050 — comfortably under the 10,000 default — while the resolver fans it into `Reserves.GetByPools` / `ReserveEmissions.GetByPools` / `OraclePrices.GetByOracles` across all pools and materializes ~600k reserve rows plus per-reserve `big.Rat` curve math into a single response.

Either fix works; pagination is just the one that also settles Jake's consistency question:
1. `BlendPoolConnection` with `first`/`after`, matching the other list queries, or
2. keep the flat list, add a LIMIT to `GetAll` and make the multiplier track the page size — cheaper if a schema change is unwelcome this late in the stack.

internal/serve/serve.go
```diff
@@ -407,4 +418,42 @@ func addComplexityCalculation(config *generated.Config) {
...
+	config.Complexity.BlendPoolPosition.Reserves = blendReservesBound
+	config.Complexity.BlendBackstopPosition.Q4w = func(childComplexity int) int { return childComplexity * 20 }
+	config.Complexity.BlendAuction.Bid = blendReservesBound
+	config.Complexity.BlendAuction.Lot = blendReservesBound
+	blendPerAccountBound := func(childComplexity int) int { return childComplexity * 10 }
+	config.Complexity.BlendAccountPositions.Pools = blendPerAccountBound
+	config.Complexity.BlendAccountPositions.Backstop = blendPerAccountBound
+	config.Complexity.BlendAccountPositions.ActiveAuctions = blendPerAccountBound
+	config.Complexity.Query.BlendPools = accountsListComplexityFunc
+	// Query.blendPool (single pool by address) and Account.blendPositions (one object)
```
aristidesstaffieri, line 447: Flagging the per-invocation cost here, separate from the cardinality question this comment answers.

The ×10 that `76f041b6` put on `Account.BlendPositions` was the only thing pricing the *fixed* cost of the field, and `439383b5` removed it. The cardinality argument is right — per-level multipliers do avoid double-charging list depth — but nothing replaced the flat part, and one `getBlendPositions` call is 13 queries: six in the account wave (`Positions`/`BackstopPositions`/`Emissions`/`PoolClaimed`/`BackstopClaimed`/`Auctions`), four in the pool wave (`Pools`/`Reserves`/`ReserveEmissions`/`BackstopPools`), three in the price wave (`OraclePrices`/`GetBackstopLPPrices`/`GetTokenMetadataByContractIDs`) — each an `errgroup` with no `SetLimit`. `Query.blendPool` is 6 the same way. As priced now, `blendPositions { backstopClaimedLp }` is 2 and `accountByAddress { … }` is 3.

The amplifier I'd worry about is the one lines 456-458 call out as untouched. `AccountTransactionEdge.stateChanges` (`pagination.graphqls:71`) is an unpaginated `[BaseStateChange!]!` left at default cost, and every `BaseStateChange` implementer carries `account: Account!` — so:

```graphql
accountByAddress(address: "G…") { transactions(first: 100) { edges {
  stateChanges { account { blendPositions { backstopClaimedLp } } } } } }
```

prices at roughly 100 × 4 ≈ 400 while re-resolving `blendPositions` once per state change, for what is mostly the same account. There's no dataloader on this field (intentionally, per `blend_positions.go:4-7` — correct for a single call, but it also means no dedup across resolutions within one request), so that's 100 × N × 13 queries out of a ~400-complexity request. Root-field aliasing gets there too (~3,300 aliases × 13 ≈ 43k queries under the 10,000 default), but the edge path is cheaper.

What makes it bite: `pgxpool` defaults to `MaxConns = 10` (`internal/db/db.go:20`), so those queue against 10 connections and unrelated endpoints — balances, transactions, state changes — time out behind them. It also interacts with the limit raise the PR notes ask for: admitting the 26,150 full selection scales both amplifiers proportionally.

Options, roughly in order of cost:
- Price the flat part additively rather than with a multiplier — e.g. `func(child int) int { return child + 13 }`, and `+6` for `blendPool`. Keeps the no-double-charge property while making the fixed cost visible.
- Give the field a request-scoped dataloader or singleflight keyed by address, which collapses the repeat resolutions the edge path triggers.
- `errgroup.SetLimit` on the three waves as a backstop.

Am I missing a reason the `stateChanges → account → blendPositions` path isn't a concern here?

internal/services/blend/rates.go
```diff
@@ -0,0 +1,440 @@
...
+// (1+apr/periods)^periods - 1. Pass BorrowAPYCompoundingPeriods or
+// SupplyAPYCompoundingPeriods per blend-sdk-js's asymmetric convention
+// (see package doc). float64 is precise enough here: this is a display
+// estimate, not a settlement amount. If apr is negative enough that the
+// per-period base (1+apr/periods) is non-positive — more than a full
+// period's worth of loss compounded, which has no sane percentage-return
+// interpretation and can even flip sign depending on whether periods is
+// odd or even — ToAPY clamps to -1 (total loss) rather than propagate that
+// nonsensical result.
+func ToAPY(apr *big.Rat, periods int64) float64 {
```
aristidesstaffieri: The `base <= 0` guard covers one direction; the other one overflows.

With `periods = 365`, `math.Pow` returns `+Inf` once `base > ~6.99` — i.e. **borrowAPR > 2,186.68** (218,668% APR). Measured:

| borrowAPR | borrowApy (365) | supplyApy (52) |
|---|---|---|
| 2,000 | 1.63e296 | 1.00e83 |
| 2,186 | 1.63e308 | 9.13e84 |
| **2,187** | **+Inf** | 9.35e84 |

The supply side is fine in practice (52 periods needs APR > ~4.4e7 for `Inf`), though it's already returning meaningless finite values well before that.

Reaching that APR doesn't take a crazy curve on its own — the `int32` 7-decimal curve fields cap the curve around 644. The unbounded factor is `irMod`: it's read as an i128 string (`entries.go:616`) and carried to `BorrowAPR` with no clamp anywhere in the pipeline, which the resolver comment acknowledges ("ir_mod isn't guaranteed to stay within int32 range", `blend_positions.go:39-41`). With `irMod` around 100, a curve sum over ~21.9 is enough, and pool deployment plus reserve config are permissionless.

Where it lands: gqlgen's `MarshalFloatContext` rejects Inf/NaN (`graphql/float.go:38-45`) and `contextMarshalerAdapter.MarshalGQL` (`graphql/jsonw.go:51-57`) calls `AddError` then writes `null`. Since `borrowApy`/`netApy` are nullable there's no null-propagation cascade — the field goes null and `errors[]` picks up `cannot marshal infinite no NaN float values`. `data` survives, but clients on a strict error policy (Apollo's default `errorPolicy: 'none'`) drop the whole response. The part I'd weigh most: `blendPools` returns every pool, so one adversarial pool degrades the shared catalog response for every client, not just its own row.

There's a NaN path as well — `blend_positions.go:528` does `borrowApyNumerator += *rp.BorrowedUsd * *rp.BorrowApy`, and `0 × +Inf = NaN`, so a supply-only position in a reserve whose `borrowApy` is `+Inf` yields `netApy = NaN` and the same marshal failure.

Both `BlendReserve` (`blend_pools.go:61`) and `BlendReservePosition` are affected since they share `computeReserveRates`.

Cheapest fix is symmetry with the existing clamp — either cap the positive side, or return a sentinel the caller maps to `nil` so the field legitimately reads null rather than erroring. Worth a `rates_test.go` case either way; there's no Inf/overflow vector there today.

internal/serve/graphql/resolvers/blend_positions.go
```diff
@@ -0,0 +1,1007 @@
...
+	poolAddr := string(bp.PoolContractID)
+
+	shares, err := parseBigInt(bp.Shares)
+	if err != nil {
+		return nil, fmt.Errorf("parsing blend backstop position shares: %w", err)
+	}
+
+	poolShares, poolTokens := big.NewInt(0), big.NewInt(0)
+	var emisIndex *big.Int
+	if backstopPool, ok := d.backstopPoolByID[poolAddr]; ok {
```
aristidesstaffieri, line 716: The godoc above frames this as "the pool's `blend_backstop_pools` row has not been ingested yet" — a map miss. I think the more common shape is a row that *does* exist with zeroed balance columns, which this `ok` check can't distinguish.

`BatchUpsertEmissions` (`backstop_pools.go:143`) inserts only the emission columns:

```sql
INSERT INTO blend_backstop_pools (pool_contract_id, emis_eps, emis_index, emis_expiration, emis_last_time, last_modified_ledger)
```

and the DDL defaults the rest (`2026-07-08.5-blend_backstop_pools.sql:6-8`):

```sql
shares TEXT NOT NULL DEFAULT '0',
tokens TEXT NOT NULL DEFAULT '0',
```

So an emissions-only window creates the row with `shares='0'`, `tokens='0'` — and `stageBackstopBEmisData` (`processor.go:811-822`) stages the emission half "independently of any balance half already staged", by design. `BEmisData` moves on the backstop's emission-distribution crank, which is independent of deposits (`PoolBalance` only changes on deposit/withdraw/donate), so a `current-state --start-ledger N` window that catches a distribution before any backstop deposit for that pool lands zeros for every reward-zone pool. It then persists until something touches that pool's `PoolBalance` — for a quiet pool, potentially days.

Downstream of that, `ok` is **true**, so `poolShares = 0` flows into `BackstopLPTokens` → `0`, and `usdValueOrNil` short-circuits on `amount.Sign() == 0` before it ever consults the price, returning `&0.0` rather than nil. A live deposit renders:

```json
{ "shares": "49999...", "lpTokens": "0", "usdValue": 0 }
```

indistinguishable from a closed position, and the same applies per q4w entry. `projectedBackstopEmissionIndex` also runs with `poolShares = 0` on that row, so emissions under-report alongside.

Two things make me want more than a doc note here:

- It reads as a confident zero rather than a missing one, which inverts the convention stated at `blend.graphqls:12-16` (null when uncomputable, 0 only when there's genuinely nothing to price). Here there is something to price.
- It's silent — no error, no log, no metric on the zero/missing pool row, so there's no signal that a figure is provisional.

Also relevant to @JiahuiWho's request for `totalUsdValue` on `BlendAccountPositions`: a rollup would absorb this zero into an account-wide total with no indication, which is harder to notice than a single suspicious row.

Options, cheapest first: treat `poolShares == 0 && userShares > 0` as uncomputable and return nil for `lpTokens`/`usdValue` rather than 0; or take blend-sdk-js's `sharesToBackstopTokens` fallback (shares unchanged) which is at least a defensible bootstrap value; or, if the current behavior is deliberate, log/count it so the gap is observable. Happy to be told the deviation is settled — mostly flagging that the "not ingested yet" framing undersells how the row actually arrives.

internal/serve/graphql/resolvers/blend_positions.go
```diff
@@ -0,0 +1,1007 @@
...
+	suppliedUsdSum, borrowedUsdSum := 0.0, 0.0
+	supplyApyNumerator, borrowApyNumerator := 0.0, 0.0
+	suppliedKnown, borrowedKnown := true, true
+
+	for _, p := range positions {
+		rp, err := d.buildReservePosition(p)
+		if err != nil {
+			return nil, err
+		}
+		if rp == nil {
```
aristidesstaffieri, line 530: **TL;DR** — this `continue` skips a reserve position without clearing `suppliedKnown`/`borrowedKnown`, so a dropped reserve silently *shrinks* the pool total instead of nulling it. That's the opposite of what the missing-price branch six lines below does, and of what this function's own doc comment promises. Worst case a live position renders `suppliedUsd: 0, usdValue: 0, netApy: 0, reserves: []` — indistinguishable from a closed position.

### Possible fix

Mirror the missing-price branch:

```go
if rp == nil {
    // A reserve we can't resolve makes the pool-wide totals uncomputable,
    // not smaller — same rule as a missing price below.
    suppliedKnown, borrowedKnown = false, false
    continue
}
```

`usdValue`/`netApy` then fall out as nil via the existing `suppliedKnown && borrowedKnown` gate, and `reserves` still omits the row. Worth a test alongside the existing missing-price case, since the two branches are supposed to behave identically and nothing currently pins that.

### Details

The doc comment on `buildPoolPosition` states the invariant:

> suppliedUsd/borrowedUsd/usdValue/netApy become nil (**not a silently-understated sum**) as soon as any contributing reserve's own suppliedUsd/borrowedUsd is nil, since a missing price on one reserve makes the pool-wide total genuinely uncomputable, not just smaller.

Two distinct "can't compute this" cases reach the loop, and they're handled inconsistently:

- **Missing price** → `rp.SuppliedUsd == nil` → `suppliedKnown = false` → totals nil. Correct.
- **Missing reserve config** → `buildReservePosition` returns `(nil, nil)` at `:379-382` → this `continue` fires *before* those checks → `suppliedKnown` stays `true` → the position's USD contribution is dropped from the sum while the sum still presents itself as complete.

On reachability, I want to be upfront that I don't think this is routine. Both `blend_positions` and `blend_reserves` are written by the same current-state pass in the same transaction, so there's no history-vs-current-state ordering window. A row is created only when one window stages both reserve halves:

| Staged halves | Writer | Creates a row? |
|---|---|---|
| `config` + `data` | `BatchUpsert` — `INSERT … ON CONFLICT` (`reserves.go:184`) | yes |
| `data` only | `BatchUpdateData` — plain `UPDATE` (`:293`) | no, silent no-op |
| `config` only | `BatchUpdateConfig` — plain `UPDATE` (`:382`) | no — `processor.go:1039` calls this out: "a skip until its ResData arrives" |

Since `set_reserve` writes both entries together, the creation ledger produces the row, and a migration started at the protocol's deployment ledger per `running-a-data-migration.md:98` observes every creation. So in correct operation there's no gap.

What's left is operational rather than routine:

- `--start-ledger` chosen too late. "The ledger where the first contract implementing your protocol was deployed" is ambiguous for a multi-contract protocol (factory + pools + backstop), and any pool whose reserves were created before the chosen ledger never gets rows — permanently, since a mature pool may never see another `set_reserve`.
- Eviction/archival of a `ResConfig` entry.

So I'd frame this as defending an invariant the code already claims rather than a live bug. The reason I still think it's worth the two lines: the failure mode is silent and self-concealing — with `reserves: []` and `usdValue: 0` there's no signal distinguishing "incomplete data" from "position closed", so if it ever does happen it surfaces as a wrong number in Freighter rather than as an error. `blendPools` isn't affected: it builds its reserve list *from* the reserves table (`blend_pools.go:333`), so a missing row means the reserve is simply absent with no total claiming to include it.

internal/serve/graphql/schema/queries.graphqls
aditya1702: Agreed nothing bounds the response but the row count — we're deliberately accepting that for now. Mainnet Blend pool count is ~20 today and grows slowly (a pool without backstop deposits carries no reserves/positions to index, so mass-deployed empty pools stay near the base row cost), and the constant ×50 overprices today's catalog rather than underpricing it.

If pool count grows past a few hundred, `BlendPoolConnection` with keyset pagination on `pool_contract_id` (GetAll already orders by it) is the fix we'd reach for, and it settles the pagination-consistency question at the same time. Deferring the schema change until the data justifies it.

internal/serve/serve.go
aditya1702: No — the `stateChanges → account → blendPositions` path was a real gap. Fixed in 6fc5af80 with the additive option: `Account.blendPositions` prices at `childComplexity + 13` and `Query.blendPool` at `childComplexity + 7` (recounting the pool path from the code: `Pools.GetByIDs` + the 5-query catalog wave + the token-metadata read — one more than 6). The amplified shape now prices at 1,701 instead of ~500, pinned by a new complexity_test case using exactly that query. The three errgroup waves also got `SetLimit(4/4/3)` as the backstop, so one resolution holds at most 4 of the pool's 10 connections.

Dataloader/singleflight skipped for now: with the flat cost the edge path is priced by its real fan-out well before it can queue meaningful load, and the single-call design stays dataloader-free per the file's design note. Root-field aliasing still scales with the deployed limit, but blendPositions is now priced on par per point with everything else it competes against.

internal/services/blend/rates.go
aditya1702: Fixed in f26bb058. `ToAPY` reports non-finite results via an `ok` flag — both directions: `math.Pow` overflow past borrow APR ~2,186, and an apr whose big.Rat exceeds float64 range making the base itself +Inf. `computeReserveRates` carries `*float64` APYs (nil when unrepresentable), and both USD-weighted aggregates treat a nil APY on a priced reserve as uncomputable — netApy/interestApy go null rather than weighting that reserve at 0%, which also closes the 0×Inf→NaN path on the positions and catalog sides.

Tests: boundary pin at 2,186 (finite) / 2,187 (non-finite) in rates_test, plus resolver tests seeding a poisoned i128-scale ir_mod and asserting null APYs with no GraphQL error and intact USD fields.

internal/serve/graphql/resolvers/blend_positions.go
aditya1702: The emissions-only-row analysis holds — the row's existence proved nothing. Fixed in 762fc65f with the first option plus the schema change it needs: the shares→LP conversion keys on the rate being known (`poolShares > 0`; a user holding shares proves deposits exist, so zero pool shares is always the gap, whether the row is absent or emissions-only), `lpTokens` becomes nullable and reads null alongside `usdValue`, each Q4W entry follows the same rule, and `emissionsEarnedBlnd` keeps the stored-index floor rather than nulling (`ProjectEmissionIndex` already returns the index unchanged at zero supply, so the figure is a legitimate accrued-so-far lower bound).

`backstopUsdForPool` applies the matching catalog rule: zero balances beside a populated emission half read null (backstop emissions only crank for reward-zone pools, which requires deposits), a missing row stays a genuine $0, and zero tokens with nonzero shares (a backstop fully drawn for bad debt) stays a genuine $0. Regression tests cover the emissions-only row, the missing row, and the catalog split.

The deeper fix — nullable balance columns so the two states are distinguishable in storage rather than by rule — is a reasonable follow-up if the catalog-side heuristic proves too clever; it also affects the totalUsdValue rollup discussion.

internal/serve/graphql/resolvers/blend_positions.go
aditya1702: Fixed in 1f56c36c, essentially the two lines: the `rp == nil` branch clears both known flags, so the totals go null under the same rule as a missing price. `buildReservePosition`'s doc drops the "shouldn't happen in practice" framing for the real mechanism (a config-only window on a brand-new reserve writes no row — `BatchUpdateConfig` is a plain UPDATE, per persistReserves' ResData-skip comment), and the sibling regression test sits next to the missing-price case, asserting the reserve drops out of `reserves` while suppliedUsd/borrowedUsd/usdValue/netApy all read null. Mutation-verified: restoring the bare `continue` fails it.

internal/serve/graphql/schema/queries.graphqls
aditya1702: Implemented as option 1 — the connection, since it also settles Jake's consistency question (ea784538 + 4b985291 on this branch; wbclient adaptation in c7561b5c on the integration-tests branch).

**TL;DR:** `blendPools(first, after, last, before) → BlendPoolConnection!`, keyset-paginated by pool contract address with a real SQL LIMIT, priced by the client-declared page size.

- **Schema/resolver:** same Relay four + shared PageInfo as every other list query, same parser/helpers. Cursor = base64 pool address (the sort key; strkey self-validates, so no version prefix — non-address cursors fail closed with BAD_USER_INPUT).
- **Data:** `PoolModel.GetPage` (keyset on the PK, exclusive cursor, LIMIT); `GetAll` deleted — the resolver was its only caller.
- **Pricing:** `Query.BlendPools` now uses the shared paginated complexity func, so the ×50 constant became the actual `first`/`last`. The cap deliberately equals DefaultPageLimit (50): each node fans ×30 reserves, so the omitted-args page is already the heaviest priced query — priced max == executable max. Full-selection worst case is 26,550 at first:50 (connection wrapper adds 8 around the 523-cost node); your 20k-pool example now prices at its true per-page cost and can't materialize the table.
- **Ordering nuance:** the connection helper trims the +1 sentinel from the front on backward pages, so GetPage returns ascending rows for both directions (the transactions.go convention) — locked in by a last/before test.
- **SDK:** `GetBlendPools` walks the cursor sequence in pages of 50 and still returns the flat `[]BlendPool`, so the integration suite is untouched.

aristidesstaffieri (approved):

PR #662, opened 2026-07-09
fix(graphql): derive account existence (is_funded) from native balance

aditya1702:
## Problem

`accountByAddress` (GraphQL) returned a non-null `Account` for **any** syntactically valid address, with no existence check. freighter-backend-v2 derives `is_funded` from whether this query returns null (`ErrAccountNotFound`), so v2 reported **`is_funded=true` for every valid address** — never-created, merged, and contract-token-only accounts alike. Most damaging: a freshly generated, never-funded account (the case Freighter's "fund your account" UX depends on) reports funded.

This surfaced in the dev freighter balance validator as a cluster of `is_funded` mismatches (v1=false, v2=true), all on addresses with no classic account.

## Fix

Gate existence on the `native_balances` row. That row is created when the account is created and **deleted on merge** (`token_ingestion.go`, `native_balances.go`), and every funded Stellar account holds ≥ base reserve — so its presence is a faithful "the classic account exists" signal, identical to Horizon/v1's definition of `isFunded`.

`AccountByAddress` now returns `nil, nil` (→ `accountByAddress: null` → v2 `is_funded=false` + empty balances) for a **G-address with no native row**. Contract (`C...`) addresses are not classic accounts and skip the gate. No freighter-backend-v2 change is needed — its `ErrAccountNotFound` branch already sets `is_funded=false` with empty balances.

## Semantics ("Match v1")

- Never-funded / merged / contract-token-only G-addresses → `is_funded=false`, no balances. Matches v1 exactly (v1 can't even simulate `balance()` for an unfunded account, so it returns `{}`).
- **No merge-time cascade deletion of other balances (deliberate):** trustlines/LP shares are account subentries and are already removed before a merge can occur (`AccountMerge` requires `numSubEntries == 0`); SAC/SEP-41 are Soroban contract-data that survive a merge and are legitimately still held. The null gate hides them for display without destroying data.

## Testing

- Updated `TestQueryResolver_Account`: funded G-address → account; G-address with no native row → nil (covers never-funded / merged / contract-token-only); C-address → account (gate skipped); invalid → `INVALID_ADDRESS`.
- `make check` (golangci-lint 0 issues; vet/shadow/deadcode/gql-validate clean) and the resolvers package `go test -race` are green.

### DIFF
--- a/internal/serve/graphql/resolvers/queries.resolvers.go
+++ b/internal/serve/graphql/resolvers/queries.resolvers.go
@@ -69,6 +69,14 @@ func (r *queryResolver) Transactions(ctx context.Context, first *int32, after *s
 }
 
 // AccountByAddress is the resolver for the accountByAddress field.
+//
+// For a classic (G...) address the account exists on-chain only if it has a
+// native_balances row: that row is written when the account ledger entry is created
+// and removed when the account is merged. Returning null (rather than a synthesized
+// Account) when the row is absent lets callers distinguish an existing account from a
+// never-created, merged, or contract-token-only address — the nullable accountByAddress
+// field is the account-existence signal. Contract (C...) addresses are not classic
+// accounts, so the existence gate does not apply to them.
 func (r *queryResolver) AccountByAddress(ctx context.Context, address string) (*types.Account, error) {
 	if !utils.IsValidStellarAddress(address) {
 		return nil, &gqlerror.Error{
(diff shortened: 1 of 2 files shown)


aditya1702:
Superseded by stellar/freighter-backend-v2#130. Nulling `accountByAddress` for addresses without a native balance also nulled their transactions/state-changes (every account-scoped query gates on `accountByAddress == nil`), breaking the account view for merged accounts, which retain history but have no classic account. The fix moved to the v2 bridge: derive is_funded from native-balance presence and hide balances when unfunded, leaving history intact.

PR #665, opened 2026-07-15
Pre-production hardening: schema, queries, ingestion, and API

aditya1702:
Pre-production hardening pass across the database schema, data-layer queries, live ingestion, and the GraphQL API. The service is pre-release, so breaking changes land directly in the migration files (clean CREATEs for launch) and in data semantics where correctness demanded it.

## Correctness fixes

- **Dataloader batching served wrong data under aliasing**: the generic loaders applied the first key's columns/limit/sort/cursor to every key batched in the 5ms window. Aliased fields with different sub-selections got silently under-fetched columns, and nested pages were clamped to a hidden cap of 10 with dropped cursors and a wrong `hasNextPage` whenever two or more parents batched together. Batches are now grouped by query shape; nested pages honor a documented 100 cap.
- **No graceful shutdown**: SIGTERM only stopped the HTTP servers; the ingest loop ran on `context.Background()` until the kubelet SIGKILLed it after the full grace period every deploy. The root context now comes from `signal.NotifyContext`, shutdown finishes the in-flight ledger, and teardown closes the ledger backend, wasm extractor (previously never closed — leaked the wazero runtime), and DB pool in order.
- **Split-brain defenses**: a CNPG failover silently releases the ingestion advisory lock while the loop keeps writing through other pool connections. The loop now probes the lock-holding session every ledger and exits when it dies, and the latest-ledger cursor advances via a guarded update that refuses values it doesn't own — cursor regression is impossible even with two concurrent ingesters.
- **Deterministic state change IDs**: `state_change_id` was a random int63 per row, so re-processing a ledger silently inserted duplicates. IDs are now deterministic ordinals per `(to_id, operation_id)`, namespaced per emitter (indexer / SEP-41 / Blend write independently), so re-ingest produces byte-identical rows and accidental duplicates fail loudly on the PK.

## Query performance (measured on a mainnet-data environment)

- Five dataloader batch queries carried no time predicate and hash-joined against full hypertable decompressions: 26–40s per batch on 9 days of data, scaling linearly with retention. Rewritten as UNNEST + `CROSS JOIN LATERAL` per-key probes with the partition column pinned (plus an `OFFSET 0` fence on the ordered variants so chunk exclusion survives the ORDER BY/LIMIT): **2.5–18ms for 100-key batches with all historical chunks excluded at runtime**, constant with retention.
- Root `transactions`/`operations`/`stateChanges` first pages did a full seq scan + heapsort of the day chunk (8–23s). They now have composite indexes matching the API sort keys, accept `since`/`until` (defaulting to the last 7 days), cap pages at 100, and fail closed on invalid limits (one overflow path previously stripped the LIMIT entirely).
- Backfill gap detection scans only the requested ledger window (was: full-hypertable DISTINCT per run).
- RPC calls no longer run inside DB transactions: protocol classification splits into Match/Prefetch/Apply phases (Apply's signature carries no RPC service — compile-checked) with the plan computed once per ledger and reused across persist retries, and checkpoint SAC-metadata enrichment moved to a follow-up transaction after the archive load commits.

## Schema & TimescaleDB

- Storage parameters retuned per table: fillfactor 90 on low-churn upsert tables (80 stays only where every row churns daily), the inert `autovacuum_vacuum_cost_limit` removed (cost_delay=0 disables cost-based throttling), the aggressive autovacuum block dropped from protocol-bounded tiny tables, and two dead `sep41_allowances` indexes removed.
- Numeric amounts/rates/prices stored as TEXT become unconstrained NUMERIC (write-time integrity, native accumulation arithmetic); the GraphQL String contract is unchanged.
- Runtime TimescaleDB policy config now applies in every ingestion mode (a backfill-first bootstrap previously ran the auto-created compression policy unbounded) and converges retention/reconcile jobs in place so job IDs and run history survive restarts.
- TimescaleDB pins move 2.25.0 → 2.28.2 everywhere; `docs/operations.md` (new) carries the upgrade runbook, production database checklist, alerting guidance, and connection-topology constraints.

## API hardening

Request/write/idle timeouts; a query depth limit (15); internal error masking (unknown errors return a generic `INTERNAL_SERVER_ERROR`; pagination/cursor mistakes are coded `BAD_USER_INPUT`); introspection gated behind `--graphql-introspection-enabled` (default off); Prometheus operation labels bounded to schema root fields; tolerant `UInt32` input parsing. Two regression tests lock the freighter constraint: the full-detail queries compute 3901/5601 against the 6000 complexity limit, and `AccountTransactionEdge.operations/stateChanges` must never gain a complexity multiplier.

## Retries & observability

Persist retries classify SQLSTATEs (data/integrity/schema errors fail fast instead of crash-looping through 5 retries); a dead datastore buffer exits immediately; the never-initialized-protocol-cursor case no longer floods the `cursor_missing` metric, making it alertable for real cursor loss; ledger-fetch duration/retries are now observed; worker pools are explicitly bounded.

## Verification

Full race-enabled unit suite and the integration suite (fresh DB, all migrations applied from scratch) pass; `make check` green at every commit. Query rewrites were verified with `EXPLAIN (ANALYZE, BUFFERS)` against production-scale data before/after.

## Breaking changes (pre-release)

Column types (TEXT→NUMERIC), state change ID values/ordering, root connections default to a 7-day window and cap pages at 100, introspection defaults off, and internal error text is no longer returned to clients.

---

### Update — NUMERIC claimed tables

Brings the new `blend_pool_claimed` / `blend_backstop_claimed` tables in line with the other Blend amount columns: `claimed_blnd` / `claimed_lp` become `NUMERIC` (native accumulation, no text round-trips), `fillfactor = 90`, vacuum cost limit dropped; readers cast back to text so the Go/GraphQL `String` contract is unchanged.


aditya1702:
Superseded by the main-based hardening stack #667 → #668 → #669 → #670 (all non-Blend hardening ported off the Blend stack so it can merge independently). The `pre-prod-hardening` branch stays as the reference for re-applying the Blend-specific residue onto the Blend PRs after they rebase.

PR #667, opened 2026-07-16
Pre-prod hardening 1/4: schema & TimescaleDB foundation

aditya1702:
## What

First of four pre-production hardening PRs (1/4 → this one, then #668 loadtest removal, #669 ingestion, #670 GraphQL/docs). This PR carries every **destructive schema change** — existing migration files are edited in place, so it must land before the production database is created. Everything here requires a fresh DB; the remaining PRs in the stack do not.

## Why

wallet-backend is pre-release: this is the last window to fix schema-level issues without ALTER migrations. These changes come out of a full data-layer/TimescaleDB audit against a mainnet-data dev cluster (EXPLAIN plans and pg_stat/timescaledb_information measurements referenced in the commit messages).

## Changes

- **Storage retune + idempotent TSDB policy config**: balance-table storage blocks retuned from measured churn (fillfactor 80→90, drop the `autovacuum_vacuum_cost_limit` no-op — per-table `cost_delay=0` already exempts the worker from cost balancing); retention/compression policy setup made idempotent and applied in **live mode only** — a backfill runs with ad-hoc flags and must not mutate or remove the shared retention/compression policies the live deployment relies on (and an active retention policy would drop the very history a backfill writes).
- **Time-pinned lateral probes for batch queries**: the five dataloader batch queries carried no partition-column predicate, so each call decompressed all chunks (26–40s per batch on 9 days of mainnet data, growing with retention). They are now UNNEST + LATERAL per-key probes pinned to the parent's ledger time behind an `OFFSET 0` fence: 2.5–18ms measured, constant with retention. Because relationship resolvers pin on the parent row's `ledger_created_at` and key their dataloaders on `(to_id, operation_id, state_change_id)`, every transaction/operation/state-change projection forces those columns regardless of the client's field selection (regression-tested on minimal projections, end to end through the pinned child lookup). Also: bound array/limit parameters replace tuple IN-lists and interpolated LIMITs, root GetAll queries fail closed on non-positive limits, `GetLedgerGaps` scans only the requested window via chunk skipping.
- **NUMERIC amount columns + deterministic state_change IDs**: TEXT numeric columns (sac/sep41 balances, allowance amounts) become NUMERIC — writes cast once at the boundary, accumulation SQL runs natively, readers cast back so the GraphQL String contract is unchanged. `state_change_id` switches from crypto/rand to deterministic per-`(to_id, operation_id)` ordinals in namespaced ranges, so re-ingesting a ledger produces byte-identical rows and duplicates fail loudly on the PK. Ordinals are keyed on the full `(to_id, operation_id)` pair, so a window-scoped emitter numbering `operation_id=0` rows across several transactions stays window-independent.
- **Per-processor sub-namespaces for state_change IDs**: each emitting indexer processor (token transfers, effects, contract deploys, SAC events) numbers its own emissions inside its own sub-range (`k<<28` within the indexer's namespace), declared via `StateChangeSubBase()` on the processor itself. IDs therefore never depend on processor registration/invocation order — adding or reordering processors cannot shift existing IDs, which would otherwise silently defeat the duplicate-insert PK backstop on replays. Within one processor, emission order derives from transaction meta (canonical on-chain data).
- **Deterministic emission order for multi-change groups**: two processors emitted several state changes for one `(to_id, operation_id)` in Go map-iteration order, which broke ID reproducibility exactly where the PK backstop matters. A `SetOptions` op updating several thresholds now emits them in fixed low → medium → high order; an `InvokeHostFunction` op deploying several contracts emits them in canonical XDR walk order (root host function, then auth invocations depth-first) with first-wins dedup by contract ID. Exact-order regression tests pin both.
- **Namespace-base enforcement for future protocol emitters**: `ProtocolProcessor` gains `StateChangeOrdinalBase()`, and `BuildProcessors` — the single funnel both boot entry points use — validates every base is a positive multiple of `StateChangeOrdinalNamespaceWidth` (`1<<40`) and unique across the built set, so a future protocol shipped with a stale or duplicated base fails fast at startup instead of silently colliding IDs on disk; a registry-wide test re-checks the invariant in CI as new protocols register. The indexer's sub-base registry gets the same treatment: `NewIndexer` validates every processor's `StateChangeSubBase()` is a positive multiple of `StateChangeSubNamespaceWidth` (`1<<28`), fits inside the indexer's emitter namespace, and is unique across streams — with the token-transfer slot (sub-base 0) reserved up front since that stream is emitted outside the processors slice — so a duplicated sub-base also fails at startup instead of as a PK violation on the first ledger where two processors emit for the same operation. The determinism contract is documented on the emitting interfaces (intra-group order must be reproducible from canonical on-chain data; cross-group parallelism/map iteration is explicitly fine; processors that can't emit a group in canonical order must sort it by a deterministic total key before ordinal assignment), and the base registry documents the bit-40/bit-28 split rationale (2^23 emitter namespaces × 2^40 ordinals per group; 4096 indexer sub-streams × 2^28 — every dimension orders of magnitude beyond transaction-meta size limits).
- **Chunk skipping on `transactions.ledger_number`** for windowed gap detection.
- **Hypertable PK consolidation**: PK column order matches the batch queries' access shape (equality column first, then the walk's sort key), making the separate API-sort indexes redundant — they're dropped; the TimescaleDB default partition-column indexes with zero consumers are dropped too.
- **TimescaleDB pinned to 2.28.2** (Dockerfile, CI workflows, docker-compose).

## Deploy notes

- Requires a **fresh database** (migrations edited in place; existing envs must be wiped + re-initialized).
- Fresh-apply is exercised end-to-end by the integration suite (fresh TimescaleDB container + all migrations).

### DIFF
--- a/internal/data/statechanges.go
+++ b/internal/data/statechanges.go
@@ -545,7 +545,7 @@ func (m *StateChangeModel) BatchGetAccountStateChangesByToIDs(ctx context.Contex
 			hi = t
 		}
 	}
-	columns = prepareColumnsWithID(columns, types.StateChange{}, "sc", "to_id", "ledger_created_at")
+	columns = prepareColumnsWithID(columns, types.StateChange{}, "sc", "to_id", "operation_id", "state_change_id", "account_id", "ledger_created_at")
 	query := fmt.Sprintf(`
 		SELECT %s
 		FROM state_changes sc
--- a/internal/data/statechanges.go
+++ b/internal/data/statechanges.go
@@ -2,8 +2,6 @@ package data
 
 import (
 	"context"
-	"crypto/rand"
-	"encoding/binary"
 	"fmt"
 	"strings"
 	"time"
@@ -190,9 +188,16 @@ func (m *StateChangeModel) GetAll(ctx context.Context, columns string, limit *in
 // BatchCopy inserts state changes using pgx's binary COPY protocol.
 // Uses native pgtype types for optimal performance (see https://github.com/jackc/pgx/issues/763).
 //
+// Each StateChange must already carry its final, deterministic StateChangeID
+// (see types.AssignStateChangeOrdinals) — BatchCopy writes it as-is and never
+// generates one.
+//
 // IMPORTANT: BatchCopy will FAIL if any duplicate records exist. The PostgreSQL COPY
 // protocol does not support conflict handling. Callers must ensure no duplicates exist
-// before calling this method, or handle the unique constraint violation error appropriately.
+// before calling this method, or handle the unique constraint violation error
+// appropriately. Because state_change_id is now deterministic, this is also the
+// idempotency backstop: re-copying an already-committed ledger's rows collides on the
+// primary key and fails loudly instead of inserting duplicates.
 func (m *StateChangeModel) BatchCopy(
 	ctx context.Context,
 	pgxTx pgx.Tx,
@@ -220,13 +225,6 @@ func (m *StateChangeModel) BatchCopy(
 		},
 		pgx.CopyFromSlice(len(stateChanges), func(i int) ([]any, error) {
 			sc := stateChanges[i]
-
-			// Generate a fresh random ID at insertion time so retries on PK collision
-			// produce new IDs automatically.
-			stateChangeID, err := generateStateChangeID()
-			if err != nil {
-				return nil, err
-			}
 
 			// Convert account_id to BYTEA (required field)
 			accountBytes, err := sc.AccountID.Value()
@@ -270,7 +268,7 @@ func (m *StateChangeModel) BatchCopy(
 
 			return []any{
 				pgtype.Int8{Int64: sc.ToID, Valid: true},
-				pgtype.Int8{Int64: stateChangeID, Valid: true},
+				pgtype.Int8{Int64: sc.StateChangeID, Valid: true},
 				pgtype.Text{String: string(sc.StateChangeCategory), Valid: true},
 				pgtypeTextFromReason(sc.StateChangeReason),
 				pgtype.Timestamptz{Time: sc.LedgerCreatedAt, Valid: true},
@@ -324,16 +322,6 @@ func (m *StateChangeModel) BatchCopy(
 	m.Metrics.QueriesTotal.WithLabelValues("BatchCopy", "state_changes").Inc()
 
 	return len(stateChanges), nil
-}
-
-// generateStateChangeID produces a random positive int64 from crypto/rand.
-// Full 63 bits of entropy, masked to ensure a positive value.
-func generateStateChangeID() (int64, error) {
-	var buf [8]byte
-	if _, err := rand.Read(buf[:]); err != nil {
-		return 0, fmt.Errorf("generating state change ID: %w", err)
-	}
-	return int64(binary.BigEndian.Uint64(buf[:]) & 0x7FFFFFFFFFFFFFFF), nil
 }
 
 // BatchGetByToID gets state changes for a single transaction with pagination support, pinned to
--- a/internal/data/statechanges.go
+++ b/internal/data/statechanges.go
@@ -32,7 +32,7 @@ var _ StateChangeWriter = (*StateChangeModel)(nil)
 // BatchGetByAccountAddress gets the state changes that are associated with the given account address.
 // Optional filters: txHash, operationID, category, and reason can be used to further filter results.
 func (m *StateChangeModel) BatchGetByAccountAddress(ctx context.Context, accountAddress string, txHash *string, operationID *int64, category *string, reason *string, columns string, limit *int32, cursor *types.StateChangeCursor, sortOrder SortOrder, timeRange *TimeRange) ([]*types.StateChangeWithCursor, error) {
-	columns = prepareColumnsWithID(columns, types.StateChange{}, "", "to_id", "operation_id", "state_change_id", "account_id")
+	columns = prepareColumnsWithID(columns, types.StateChange{}, "", "to_id", "operation_id", "state_change_id", "account_id", "ledger_created_at")
 	var queryBuilder strings.Builder
 	args := []interface{}{types.AddressBytea(accountAddress)}
 	argIndex := 2
@@ -128,7 +128,7 @@ func (m *StateChangeModel) GetAll(ctx context.Context, columns string, limit *in
 	if err := validatePositiveLimit(limit); err != nil {
 		return nil, err
 	}
-	columns = prepareColumnsWithID(columns, types.StateChange{}, "", "to_id", "operation_id", "state_change_id", "account_id")
+	columns = prepareColumnsWithID(columns, types.StateChange{}, "", "to_id", "operation_id", "state_change_id", "account_id", "ledger_created_at")
 	var queryBuilder strings.Builder
 	var args []interface{}
 	argIndex := 1
@@ -327,7 +327,7 @@ func (m *StateChangeModel) BatchCopy(
 // BatchGetByToID gets state changes for a single transaction with pagination support, pinned to
 // the parent transaction's ledger_created_at for partition-column chunk exclusion.
 func (m *StateChangeModel) BatchGetByToID(ctx context.Context, toID int64, ledgerCreatedAt time.Time, columns string, limit *int32, cursor *types.StateChangeCursor, sortOrder SortOrder) ([]*types.StateChangeWithCursor, error) {
-	columns = prepareColumnsWithID(columns, types.StateChange{}, "", "to_id", "operation_id", "state_change_id", "account_id")
+	columns = prepareColumnsWithID(columns, types.StateChange{}, "", "to_id", "operation_id", "state_change_id", "account_id", "ledger_created_at")
 	var queryBuilder strings.Builder
 	fmt.Fprintf(&queryBuilder, `
 		SELECT %s, ledger_created_at as cursor_ledger_created_at, to_id as cursor_to_id, operation_id as cursor_operation_id, state_change_id as cursor_state_change_id
@@ -391,7 +391,7 @@ func (m *StateChangeModel) BatchGetByToIDs(ctx context.Context, toIDs []int64, l
 	if len(toIDs) != len(ledgerCreatedAts) {
 		return nil, fmt.Errorf("toIDs and ledgerCreatedAts must be parallel arrays of equal length, got %d and %d", len(toIDs), len(ledgerCreatedAts))
 	}
-	columns = prepareColumnsWithID(columns, types.StateChange{}, "sc", "to_id", "operation_id", "state_change_id", "account_id")
+	columns = prepareColumnsWithID(columns, types.StateChange{}, "sc", "to_id", "operation_id", "state_change_id", "account_id", "ledger_created_at")
 
 	// The ORDER BY + LIMIT live inside the LATERAL (a per-transaction top-N), replacing the old
 	// ROW_NUMBER()-over-everything CTE with an equivalent per-to_id cap; a subquery containing
@@ -458,7 +458,7 @@ func (m *StateChangeModel) BatchGetByToIDs(ctx context.Context, toIDs []int64, l
 // BatchGetByOperationID gets state changes for a single operation with pagination support,
 // pinned to the parent operation's ledger_created_at for partition-column chunk exclusion.
 func (m *StateChangeModel) BatchGetByOperationID(ctx context.Context, operationID int64, ledgerCreatedAt time.Time, columns string, limit *int32, cursor *types.StateChangeCursor, sortOrder SortOrder) ([]*types.StateChangeWithCursor, error) {
-	columns = prepareColumnsWithID(columns, types.StateChange{}, "", "to_id", "operation_id", "state_change_id", "account_id")
+	columns = prepareColumnsWithID(columns, types.StateChange{}, "", "to_id", "operation_id", "state_change_id", "account_id", "ledger_created_at")
 	var queryBuilder strings.Builder
 	fmt.Fprintf(&queryBuilder, `
 		SELECT %s, ledger_created_at as cursor_ledger_created_at, to_id as cursor_to_id, operation_id as cursor_operation_id, state_change_id as cursor_state_change_id
@@ -545,7 +545,7 @@ func (m *StateChangeModel) BatchGetAccountStateChangesByToIDs(ctx context.Contex
 			hi = t
 		}
 	}
-	columns = prepareColumnsWithID(columns, types.StateChange{}, "sc", "to_id")
+	columns = prepareColumnsWithID(columns, types.StateChange{}, "sc", "to_id", "ledger_created_at")
 	query := fmt.Sprintf(`
 		SELECT %s
 		FROM state_changes sc
@@ -577,7 +577,7 @@ func (m *StateChangeModel) BatchGetByOperationIDs(ctx context.Context, operation
 	if len(operationIDs) != len(ledgerCreatedAts) {
 		return nil, fmt.Errorf("operationIDs and ledgerCreatedAts must be parallel arrays of equal length, got %d and %d", len(operationIDs), len(ledgerCreatedAts))
 	}
-	columns = prepareColumnsWithID(columns, types.StateChange{}, "sc", "to_id", "operation_id", "state_change_id", "account_id")
+	columns = prepareColumnsWithID(columns, types.StateChange{}, "sc", "to_id", "operation_id", "state_change_id", "account_id", "ledger_created_at")
 
 	// The ORDER BY + LIMIT live inside the LATERAL (a per-operation top-N), replacing the old
 	// ROW_NUMBER()-over-everything CTE with an equivalent per-operation cap; a subquery
--- a/internal/data/statechanges.go
+++ b/internal/data/statechanges.go
@@ -48,9 +48,11 @@ func (m *StateChangeModel) BatchGetByAccountAddress(ctx context.Context, account
 	// Time range filter: enables TimescaleDB chunk pruning on the state_changes hypertable
 	args, argIndex = appendTimeRangeConditions(&queryBuilder, "ledger_created_at", timeRange, args, argIndex)
 
-	// Add transaction hash filter if provided (uses subquery to find to_id by hash)
+	// Add transaction hash filter if provided (uses subquery to find to_id(s) by hash).
+	// idx_transactions_hash is non-unique (TimescaleDB can't enforce unique(hash) on a
+	// hypertable), so this must tolerate more than one matching row instead of erroring.
 	if txHash != nil {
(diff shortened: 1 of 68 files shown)


internal/serve/graphql/resolvers/transaction.resolvers.go
```diff
@@ -34,11 +34,12 @@ func (r *transactionResolver) Operations(ctx context.Context, obj *types.Transac
...
-		Columns:   strings.Join(dbColumns, ", "),
-		Limit:     &queryLimit,
-		Cursor:    params.Cursor,
-		SortOrder: params.SortOrder,
+		ToID:            obj.ToID,
+		Columns:         strings.Join(dbColumns, ", "),
+		Limit:           &queryLimit,
+		Cursor:          params.Cursor,
+		SortOrder:       params.SortOrder,
+		LedgerCreatedAt: obj.LedgerCreatedAt,
```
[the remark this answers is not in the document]
aditya1702: Confirmed and fixed in b6fe8ac8: every transaction/operation/state-change projection now forces `ledger_created_at` (the same treatment the key columns and `BatchGetByAccountAddress` already had), so relationship pins never depend on the client's scalar selection. Regression tests cover minimal projections on every fetch path plus the pinned child lookup end to end (`Test*_MinimalProjectionHydratesLedgerCreatedAt`), verified red before the fix and green after.

internal/serve/graphql/resolvers/transaction.resolvers.go
```diff
@@ -98,11 +99,12 @@ func (r *transactionResolver) StateChanges(ctx context.Context, obj *types.Trans
...
-		Columns:   strings.Join(dbColumns, ", "),
-		Limit:     &queryLimit,
-		Cursor:    params.StateChangeCursor,
-		SortOrder: params.SortOrder,
+		ToID:            obj.ToID,
+		Columns:         strings.Join(dbColumns, ", "),
+		Limit:           &queryLimit,
+		Cursor:          params.StateChangeCursor,
+		SortOrder:       params.SortOrder,
+		LedgerCreatedAt: obj.LedgerCreatedAt,
```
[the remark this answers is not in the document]
aditya1702: Confirmed and fixed in b6fe8ac8: every transaction/operation/state-change projection now forces `ledger_created_at` (the same treatment the key columns and `BatchGetByAccountAddress` already had), so relationship pins never depend on the client's scalar selection. Regression tests cover minimal projections on every fetch path plus the pinned child lookup end to end (`Test*_MinimalProjectionHydratesLedgerCreatedAt`), verified red before the fix and green after.

internal/serve/graphql/resolvers/operation.resolvers.go
```diff
@@ -82,11 +82,12 @@ func (r *operationResolver) StateChanges(ctx context.Context, obj *types.Operati
...
-		Columns:     strings.Join(dbColumns, ", "),
-		Limit:       &queryLimit,
-		Cursor:      params.StateChangeCursor,
-		SortOrder:   params.SortOrder,
+		OperationID:     obj.ID,
+		Columns:         strings.Join(dbColumns, ", "),
+		Limit:           &queryLimit,
+		Cursor:          params.StateChangeCursor,
+		SortOrder:       params.SortOrder,
+		LedgerCreatedAt: obj.LedgerCreatedAt,
```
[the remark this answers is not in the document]
aditya1702: Confirmed and fixed in b6fe8ac8: every transaction/operation/state-change projection now forces `ledger_created_at` (the same treatment the key columns and `BatchGetByAccountAddress` already had), so relationship pins never depend on the client's scalar selection. Regression tests cover minimal projections on every fetch path plus the pinned child lookup end to end (`Test*_MinimalProjectionHydratesLedgerCreatedAt`), verified red before the fix and green after.

internal/serve/graphql/resolvers/resolver.go
```diff
@@ -128,14 +129,15 @@ func (r *Resolver) resolveStateChangeAccount(accountID types.AddressBytea) (*typ
...
 	loaders := ctx.Value(middleware.LoadersKey).(*dataloaders.Dataloaders)
 	dbColumns := GetDBColumnsForFields(ctx, types.Operation{})
 
 	stateChangeKey := fmt.Sprintf("%d-%d-%d", toID, operationID, stateChangeID)
 	loaderKey := dataloaders.OperationColumnsKey{
-		StateChangeID: stateChangeKey,
-		Columns:       strings.Join(dbColumns, ", "),
+		StateChangeID:   stateChangeKey,
+		Columns:         strings.Join(dbColumns, ", "),
+		LedgerCreatedAt: ledgerCreatedAt,
```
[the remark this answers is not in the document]
aditya1702: Confirmed and fixed in b6fe8ac8: every transaction/operation/state-change projection now forces `ledger_created_at` (the same treatment the key columns and `BatchGetByAccountAddress` already had), so relationship pins never depend on the client's scalar selection. Regression tests cover minimal projections on every fetch path plus the pinned child lookup end to end (`Test*_MinimalProjectionHydratesLedgerCreatedAt`), verified red before the fix and green after.

internal/serve/graphql/resolvers/resolver.go
```diff
@@ -146,14 +148,15 @@ func (r *Resolver) resolveStateChangeOperation(ctx context.Context, toID int64,
...
 	loaders := ctx.Value(middleware.LoadersKey).(*dataloaders.Dataloaders)
 	dbColumns := GetDBColumnsForFields(ctx, types.Transaction{})
 
 	stateChangeKey := fmt.Sprintf("%d-%d-%d", toID, operationID, stateChangeID)
 	loaderKey := dataloaders.TransactionColumnsKey{
-		StateChangeID: stateChangeKey,
-		Columns:       strings.Join(dbColumns, ", "),
+		StateChangeID:   stateChangeKey,
+		Columns:         strings.Join(dbColumns, ", "),
+		LedgerCreatedAt: ledgerCreatedAt,
```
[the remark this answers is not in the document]
aditya1702: Confirmed and fixed in b6fe8ac8: every transaction/operation/state-change projection now forces `ledger_created_at` (the same treatment the key columns and `BatchGetByAccountAddress` already had), so relationship pins never depend on the client's scalar selection. Regression tests cover minimal projections on every fetch path plus the pinned child lookup end to end (`Test*_MinimalProjectionHydratesLedgerCreatedAt`), verified red before the fix and green after.

internal/indexer/types/types.go
```diff
@@ -687,6 +687,75 @@ type StateChangeCursorGetter interface {
...
+// emitter's namespace base — plus its processor's sub-base for multi-stream
+// emitters like the indexer. Processing the same input twice (e.g. a
+// re-ingested ledger) yields byte-identical IDs, so a duplicate BatchCopy
+// fails loudly on the state_changes primary key instead of inserting
+// duplicate rows.
+//
+// Callers must pass one emission stream at a time — the final, filtered slice
+// actually handed to BatchCopy — so ordinals come out contiguous (1..N, no
+// gaps) per group within the stream's namespace.
+func AssignStateChangeOrdinals(changes []StateChange, base int64) {
```
aristidesstaffieri, line 775: **Grouping ordinals by `OperationID` alone doesn't match the contract the registry promises — and the gap silently defeats the PK backstop for window-scoped emitters.**

The namespace registry comment above says each emitter "numbers its own state changes 1..N in deterministic emission order within an *(to_id, operation_id)* group (see AssignStateChangeOrdinals)", but the implementation keys the counter map on `OperationID` only:

```go
next := make(map[int64]int64, len(changes))
for i := range changes {
    opID := changes[i].OperationID
    next[opID]++
    changes[i].StateChangeID = base + next[opID]
}
```

Both current emitters are safe only incidentally:

- the indexer assigns per-transaction stream, so `ToID` is constant within a call;
- SEP-41 stages across a multi-ledger window, but its `opID = toid(ledger, txIdx, opIdx+1)` embeds the transaction and is never 0, so an `OperationID` group can never span `ToID`s.

Nothing enforces either property at the call sites. The failing usage is exactly what the documented contract invites: a future window-scoped emitter (the Blend base is already reserved) staging transaction-level changes with `OperationID = 0` for several transactions in one window. Those all share one counter bucket, so a given row's ordinal depends on which *other* ledgers happen to be in the window. Re-ingesting the same ledger under different window bounds (live vs. backfill, or a re-run migration) then produces *different* `state_change_id`s for the same logical rows — `BatchCopy` inserts silent duplicates instead of failing on the PK, which is precisely the corruption this scheme exists to surface.

Fix is one line of shape change — key the map on the pair, which also makes the code match its own doc:

```go
type ordinalKey struct{ toID, opID int64 }

next := make(map[ordinalKey]int64, len(changes))
for i := range changes {
    k := ordinalKey{changes[i].ToID, changes[i].OperationID}
    next[k]++
    changes[i].StateChangeID = base + next[k]
}
```

(The function doc's "within each distinct OperationID group" wording would want the same update.)

Worth doing in this PR rather than later: once rows exist, any emitter that ever hit the single-key bucket has persisted IDs that a re-run under the fixed grouping won't reproduce — cheap now, messy after.

internal/ingest/ingest.go
```diff
@@ -134,10 +134,12 @@ func setupDeps(cfg Configs) (services.IngestService, error) {
...
 
-	if cfg.IngestionMode == services.IngestionModeLive {
-		if err := configureHypertableSettings(ctx, dbConnectionPool, cfg.ChunkInterval, cfg.RetentionPeriod, cfg.OldestLedgerCursorName, cfg.CompressionScheduleInterval, cfg.CompressAfter, cfg.MaxChunksToCompress); err != nil {
-			return nil, fmt.Errorf("configuring hypertable settings: %w", err)
-		}
+	// Configured regardless of ingestion mode: a backfill-first bootstrap must not
+	// run under TimescaleDB's auto-created columnstore policy defaults (unlimited
+	// maxchunks, no retention). All settings applied here are idempotent, so
+	// running this on every start (live or backfill) is safe.
+	if err := configureHypertableSettings(ctx, dbConnectionPool, cfg.ChunkInterval, cfg.RetentionPeriod, cfg.OldestLedgerCursorName, cfg.CompressionScheduleInterval, cfg.CompressAfter, cfg.MaxChunksToCompress); err != nil {
```
aristidesstaffieri: **Removing the `IngestionModeLive` guard here trades a per-invocation property (idempotence) for one this call doesn't have: convergence across processes with different flags.**

The new comment says "All settings applied here are idempotent, so running this on every start (live or backfill) is safe." That's true for one process re-running with the same flags — but the policies live in shared database state, and every process start now imposes *its own* flag values on them. A backfill job is typically launched ad hoc, without the live deployment's flag set, and the empty/zero defaults are interpreted as "disable"/"skip" rather than "leave unchanged". Three concrete failure modes fall out:

1. **Backfill without `--retention-period` strips live retention.** The flag defaults to `""`, and `configureRetentionPolicy` treats empty as disable: `remove_retention_policy(...)` runs on every hypertable, and `configureReconciliationJob("")` executes `delete_job` on `reconcile_oldest_cursor`. A one-off backfill on a production database silently disables retention and cursor reconciliation until the next live pod restart re-creates them — meanwhile chunks are never dropped and `oldest_ingest_ledger` goes stale.

2. **Backfill *with* retention drops the history it's writing.** The policy is now installed at startup, before the backfill writes. Backfilling a range older than `RetentionPeriod` (e.g. 90 days of history under a 30-day policy — the bootstrap case the new comment is motivated by) runs under an active `policy_retention` job that drops the freshly written chunks: the backfill logs success, ~60 days of it vanish. The progressive recompressor's `compress_chunk($chunk)` calls also race the concurrent chunk drops and error out, failing batches.

3. **A backfill's compression cap ratchets live jobs down permanently.** `--compression-max-chunks 10` writes `maxchunks_to_compress = 10` into every compression job's config, but the write is gated on `maxChunksToCompress > 0` — so when live restarts with its default of 0, the block is skipped and the cap is never removed. Live compression is permanently limited to 10 chunks per run until someone `alter_job`s by hand.

One fix shape covers all three. Either:

- **restore the mode guard** (backfill never touches policies), which gives up the stated motivation — a backfill-first bootstrap running under TimescaleDB's auto-created policy defaults; or
- **keep the unconditional call but make it convergent**: empty/zero flag values mean "leave existing configuration unchanged" instead of the current asymmetry where `""` means *disable* for retention but `0` means *skip* for maxchunks. The bootstrap concern is then handled explicitly — e.g. only apply defaults-sensitive settings when the policy/job doesn't exist yet, so a backfill can create missing policies but never mutate or remove ones the live deployment configured.

The second keeps the bootstrap win and removes the operational footgun; the first is smaller but reintroduces the "backfill-first bootstrap runs uncapped" gap this PR set out to close.

internal/data/statechanges.go
```diff
@@ -530,7 +545,7 @@ func (m *StateChangeModel) BatchGetAccountStateChangesByToIDs(ctx context.Contex
 			hi = t
 		}
 	}
-	columns = prepareColumnsWithID(columns, types.StateChange{}, "sc", "to_id")
+	columns = prepareColumnsWithID(columns, types.StateChange{}, "sc", "to_id", "ledger_created_at")
```
aristidesstaffieri: **The forced-column set here is missing the id columns the state-change resolvers build their loader keys from — nested `operation { ... }` silently resolves to null on minimal selections.**

This line (correctly) forces `to_id` and `ledger_created_at` into the projection for the new time-pinned child lookups, but its sibling `BatchGetByAccountAddress` forces the full key set:

```go
// BatchGetByAccountAddress (line ~35)
columns = prepareColumnsWithID(columns, types.StateChange{}, "", "to_id", "operation_id", "state_change_id", "account_id", "ledger_created_at")

// BatchGetAccountStateChangesByToIDs (this line)
columns = prepareColumnsWithID(columns, types.StateChange{}, "sc", "to_id", "ledger_created_at")
```

`operation_id` and `state_change_id` aren't just data fields — every state-change resolver feeds them into its dataloader key (`resolver.go:136`):

```go
stateChangeKey := fmt.Sprintf("%d-%d-%d", toID, operationID, stateChangeID)
```

So for an account-scoped query whose selection doesn't happen to include a field mapping to `operation_id` — e.g. `account → transactions → stateChanges { type, amount, operation { id } }` — the rows come back with `OperationID = 0`, the loader key is `"<toID>-0-0"`, and the operation lookup probes `o.id = 0`: no match, `operation` resolves to null even though it exists. As a second-order effect, *all* state changes of that transaction collapse onto the same `"<toID>-0-0"` key, so the dataloader dedups them into one shared (null) result.

Fix is to match the sibling's forced set on this line:

```go
columns = prepareColumnsWithID(columns, types.StateChange{}, "sc", "to_id", "operation_id", "state_change_id", "account_id", "ledger_created_at")
```

Since this PR already added regression tests for minimal projections end-to-end through the pinned child lookup, extending one of those to assert `operation { id }` is non-null on a minimal account-scoped state-change selection would lock this in.

internal/indexer/types/types.go
aditya1702: Fixed in 150ec416. `AssignStateChangeOrdinals` now keys its counter map on the `(to_id, operation_id)` pair, so it matches the namespace-registry contract and a window-scoped emitter staging `operation_id=0` rows across several transactions gets window-independent ordinals (no shared bucket, no re-ingest divergence, PK backstop intact).

Added `TestAssignStateChangeOrdinals_GroupsByToIDAndOperationID`: `operation_id=0` across two distinct `to_id`s must yield independent `1..N` per pair (`1,2,1,2`); the old single-key code produced `1,2,3,4`. The existing ordinal tests stay green since pair-keying is output-identical for the current emitters (indexer streams per transaction, SEP-41's `operation_id` embeds the tx index).

internal/ingest/ingest.go
aditya1702: Fixed in 1ef14505 by restoring the `IngestionModeLive` guard around `configureHypertableSettings`. A backfill no longer touches shared retention/compression policy state, which closes all three failure modes at once: an empty `--retention-period` stripping retention + the reconciliation job; a backfill dropping the history it is writing under an active retention policy; and `--compression-max-chunks` ratcheting live compression jobs down permanently (the `>0` write is never reset by live's default of `0`).

Tradeoff accepted: a backfill-first bootstrap runs under TimescaleDB's auto-created columnstore defaults until the first live pod configures policies — a bounded, one-time operational step rather than a silent data-loss path. `timescaledb.go` is unchanged; only the call site is gated.

internal/data/statechanges.go
aditya1702: Fixed in c0124fe4. `BatchGetAccountStateChangesByToIDs` now forces the full key set (`to_id, operation_id, state_change_id, account_id, ledger_created_at`), matching its sibling `BatchGetByAccountAddress`, so a minimal account-scoped selection no longer comes back with `operation_id`/`state_change_id` zeroed and the loader key no longer collapses to `<to_id>-0-0` (which nulled nested `operation`/`transaction` and deduped every row into one).

Extended `TestStateChangeModel_BatchGetAccountStateChangesByToIDs` with a minimal-projection sub-test asserting `operation_id`/`state_change_id`/`account_id` are hydrated when the client selects only `state_change_category`; verified red before the fix (both id columns came back `0`) and green after.

internal/indexer/types/types.go
```diff
@@ -687,6 +687,95 @@ type StateChangeCursorGetter interface {
...
+// binding constraint. Protocol emitters (SEP-41, Blend) are single-stream and use their
+// base directly; a future multi-stream emitter subdivides its own namespace
+// the same way. Like the emitter bases, sub-bases must never change once
+// rows exist with them; new processors take the next unused slot.
+const (
+	StateChangeSubBaseTokenTransfer  int64 = 0 << 28
+	StateChangeSubBaseEffects        int64 = 1 << 28
+	StateChangeSubBaseContractDeploy int64 = 2 << 28
+	StateChangeSubBaseSACEvents      int64 = 3 << 28
+)
```
JiahuiWho, line 761: Would it be worth validating sub-base uniqueness in the `Indexer` constructor, like BuildProcessors does for protocol bases? 
As is, a duplicated sub-base wouldn't fail at startup, it'd surface as a PK violation on the first ledger where both processors emit for the same operation.

JiahuiWho (approved):

internal/indexer/types/types.go
aditya1702: Added in 918f38c3. `NewIndexer` now validates the sub-base registry the same way `BuildProcessors` validates protocol bases: every processor's `StateChangeSubBase()` must be a positive multiple of the (newly named) `types.StateChangeSubNamespaceWidth`, fit inside the indexer's emitter namespace, and be unique across streams — with the token-transfer slot (sub-base 0) reserved up front since that stream is emitted outside the processors slice. A duplicated or misaligned sub-base now fails at startup instead of as a PK violation on the first ledger where two processors emit for the same operation. Covered by `TestValidateStateChangeSubBases` (duplicate, reserved-zero, misaligned, negative, and out-of-namespace cases).

PR #668, opened 2026-07-16
Pre-prod hardening 2/4: remove the old loadtest backend

aditya1702:
## What

Pre-production hardening 2/4 (stacked on #667). Removes the old loadtest backend: the `loadtest` CLI command, `internal/loadtest/` (including a ~100MB ledger fixture and a 323k-line seed SQL file), its ledger-backend hook, and the exported `PersistLedgerData` wrapper that existed only for it. Two now-unused go.mod dependencies drop out. Comments and the CAS-gating test name that referenced the exported wrapper now use the unexported `persistLedgerData`.

Pure deletion — no behavior change for serve/ingest. 324k deleted lines are almost entirely testdata.


internal/services/ingest_live.go
```diff
@@ -28,14 +28,9 @@ const (
 	lagMetricUpdateInterval            = 1 * time.Second
 )
 
-// PersistLedgerData persists processed ledger data to the database in a single atomic transaction.
-// This is the shared core used by both live ingestion and loadtest.
-// It handles: trustline assets, contract tokens, filtered data insertion,
-// token changes, and cursor update.
-func (m *ingestService) PersistLedgerData(ctx context.Context, ledgerSeq uint32, buffer *indexer.IndexerBuffer, cursorName string) (int, int, error) {
```
JiahuiWho, line 35: Since the exported wrapper is gone, could we also lowercase the three comment references to PersistLedgerData (ingest_store.go:30, indexer_buffer.go:693, protocol_wasms.go:27)?

JiahuiWho (approved):

internal/services/ingest_live.go
aditya1702: Done in 35050238. Lowercased those three plus a fourth comment reference the grep turned up (`ingest.go:71`, the `ProtocolValidators` config doc), and renamed `Test_PersistLedgerData_ProtocolCASGating` → `Test_persistLedgerData_ProtocolCASGating` (with its mention in `data_migration_test.go`) to match the repo's `Test_lowercase` convention for unexported functions — `rg PersistLedgerData` now comes back empty. Note: the `ingest_store.go:30` reference conflicts with #669, which rewrites that paragraph to describe the new `casProtocolCursor` flow; the rebase keeps #669's version there.

PR #670, opened 2026-07-16
Pre-prod hardening 4/4: GraphQL API hardening + ops docs

aditya1702:
## What

Pre-production hardening 4/4 (stacked on #669). GraphQL API hardening — including trimming the API surface to the three root queries a wallet backend actually needs — plus the production operations reference.

## Changes

- **Removed the unbounded root list queries** `transactions`, `operations`, and `stateChanges`: a wallet backend has no need for cross-account history scans. The API now exposes only `transactionByHash`, `accountByAddress`, and `operationById` (with their nested and account-scoped connections, which are unchanged). Breaking change, acceptable pre-release. The removal carries its full dead-code closure with it — the three resolvers, the data-layer `GetAll` methods and `validatePositiveLimit`, the root pagination/time-window helpers, the client SDK `GetTransactions`/`GetOperations`/`GetStateChanges` methods and their builders/response structs, and the two DB sort indexes (`idx_transactions_created_at_sort`, `idx_operations_created_at_sort`) that existed only to serve the root sort. Those two indexes were introduced in #667 and removed here, both within this unmerged stack, so no deployed database ever created them; the remaining `ledger_created_at`-ordered lookups (`GetOldestLedger`, the oldest-cursor reconcile job) are served by TimescaleDB's per-chunk columnstore metadata indexes, confirmed with `EXPLAIN`. `make check` (including `deadcode`) is green.
- **Dataloader batches grouped by query shape**: loaders applied `keys[0]`'s columns/limit/cursor to the whole batch, so two aliased fields with different arguments in one query could silently get each other's data (or panic on a nil limit). Batches are now grouped by (columns, limit, cursor, sort) shape; nil/non-positive limits fail closed. Regression tests cover the aliased-query cases.
- **Complexity regression test**: locks the freighter full-detail query under the production `GRAPHQL_COMPLEXITY_LIMIT=6000` so future schema/complexity changes can't silently break the primary consumer.
- **API hardening**: per-request timeouts, query depth limit, error masking (internal errors no longer leak SQL/driver text to clients), root-query time bounds, introspection env-gating, and `UInt32` scalar accepting all numeric wire types. The error mask allowlists protocol error codes (e.g. gqlgen APQ's `PERSISTED_QUERY_NOT_FOUND`) so a persisted-query client still receives the signal to retry with the full query on a cold-cache hash miss.
- **1ms dataloader collection window** (up from the 16µs default) with per-loader batching metrics — cuts per-request query count for the freighter full-detail shape.
- **Docs**: production operations reference (`docs/operations.md`) and GraphQL README synced to the actual schema — documented-but-nonexistent mutations removed, the three root queries documented, and all root list-query examples (pagination, DataLoader, APQ, field-selection) repointed to their account-scoped equivalents. Two cursor-CAS SQL casts widened `::integer` → `::bigint`.


internal/serve/graphql/utils.go
```diff
@@ -8,14 +8,40 @@ import (
...
 const DefaultPageLimit int32 = 50
 
-// CustomErrorPresenter provides more detailed error messages for GraphQL validation errors
+// clientSafeErrorCodes enumerates every GraphQL error "code" extension the API deliberately
+// constructs (or that gqlgen itself assigns) for client consumption: validation/parse errors,
+// the complexity and depth limits, and the app's own coded resolver errors (see
+// resolvers/errors.go and the badUserInputError/balanceBadUserInputError helpers). Any error
+// presented without one of these codes is treated as an internal failure by CustomErrorPresenter
+// and masked before it reaches the client.
+var clientSafeErrorCodes = map[string]bool{
```
JiahuiWho, line 27: Should `PERSISTED_QUERY_NOT_FOUND` be in this list? gqlgen's APQ extension returns it on a hash-only cache miss (apq.go sets that code), and clients need it verbatim to retry with the full query

internal/db/migrations/2025-06-10.2-transactions.sql
```diff
@@ -25,14 +25,11 @@ SELECT enable_chunk_skipping('transactions', 'to_id');
...
--- sorts by (ledger_created_at DESC, to_id DESC), which the default index can't serve
--- for a top-N first page (it still needs a heapsort on to_id within each timestamp).
--- This composite index matches the API sort key directly and makes the auto-created
--- one redundant.
+-- TimescaleDB's default single-column index on the partition column. Retention drops chunks by
+-- range metadata, and no query path filters this table by bare ledger_created_at (reads go through
+-- the primary key, idx_transactions_hash, or the transactions_accounts table), so it has zero
+-- consumers.
 DROP INDEX IF EXISTS transactions_ledger_created_at_idx;
-CREATE INDEX idx_transactions_created_at_sort ON transactions (ledger_created_at DESC, to_id DESC);
```
JiahuiWho, line 35: I still see 2 queries ordering by bare `ledger_created_at` on this table: `GetOldestLedger` (ingest_store.go:198) and the `reconcile_oldest_cursor` job (timescaledb.go:79), do they still plan okay?

internal/serve/graphql/utils.go
aditya1702: Added in `83e9c5ea`. `PERSISTED_QUERY_NOT_FOUND` is now in the client-safe allowlist, so gqlgen's APQ hash-only cache miss reaches the client verbatim instead of being masked to `INTERNAL_SERVER_ERROR`, and the client retries with the full query. Covered by a dedicated test for the cold-cache path (plus the allowlist pass-through table).

internal/db/migrations/2025-06-10.2-transactions.sql
aditya1702: Verified on the dev DB with read-only `EXPLAIN` — both queries still plan fine without these indexes.

The `ORDER BY ledger_created_at ASC, {to_id|id} ASC LIMIT 1` is served by TimescaleDB's per-chunk **columnstore metadata** indexes (`compress_hyper_*__ts_meta_min_1__ts_meta_max_1__ts_idx`), which come from `tsdb.orderby` and exist independently of any btree — neither `idx_*_created_at_sort` nor the default `*_ledger_created_at_idx` appears in the plan even when present. The `Custom Scan (ChunkAppend)` streams chunks oldest-first and the `LIMIT 1` short-circuits on the oldest chunk (the large recent chunk is never touched).

Measured plan cost: ~10 with the indexes present, ~98 (transactions) / ~197 (operations) with all index scans force-disabled — negligible either way for these low-frequency queries.

internal/serve/graphql/utils.go
```diff
@@ -8,14 +8,40 @@ import (
...
+	"INVALID_TRANSACTION_HASH":  true,
+	"INVALID_ADDRESS":           true,
+	"INTERNAL_ERROR":            true, // resolvers/account_balances.go: already a sanitized, generic message
+	"GRAPHQL_VALIDATION_FAILED": true, // gqlparser query validation (unknown field, bad args, ...)
+	"GRAPHQL_PARSE_FAILED":      true, // gqlparser query parse errors
+	"COMPLEXITY_LIMIT_EXCEEDED": true, // gqlgen extension.FixedComplexityLimit
+	"QUERY_TOO_DEEP":            true, // query depth limit extension
+	"UNAUTHENTICATED":           true,
+	"FORBIDDEN":                 true,
+}
```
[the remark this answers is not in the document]
aditya1702: Added in `83e9c5ea`. `PERSISTED_QUERY_NOT_FOUND` is now in the client-safe allowlist, so gqlgen's APQ hash-only cache miss reaches the client verbatim instead of being masked to `INTERNAL_SERVER_ERROR`, and the client retries with the full query. Covered by a dedicated test for the cold-cache path (plus the allowlist pass-through table).

internal/db/migrations/2025-06-10.2-transactions.sql
```diff
@@ -25,14 +25,11 @@ SELECT enable_chunk_skipping('transactions', 'to_id');
...
--- auto-creates on the hypertable's partition column. The root GraphQL connection
--- sorts by (ledger_created_at DESC, to_id DESC), which the default index can't serve
--- for a top-N first page (it still needs a heapsort on to_id within each timestamp).
--- This composite index matches the API sort key directly and makes the auto-created
--- one redundant.
+-- TimescaleDB's default single-column index on the partition column. Retention drops chunks by
+-- range metadata, and no query path filters this table by bare ledger_created_at (reads go through
+-- the primary key, idx_transactions_hash, or the transactions_accounts table), so it has zero
+-- consumers.
 DROP INDEX IF EXISTS transactions_ledger_created_at_idx;
```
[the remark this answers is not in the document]
aditya1702: This one doesn't apply here — no forward `DROP` migration is needed.

`idx_transactions_created_at_sort` / `idx_operations_created_at_sort` were **introduced in #667 and removed in this PR — both within the same unmerged hardening stack**. They were never on `main` (confirmed via `git grep` on `main`) and are absent from every deployed database (verified directly against the dev DB, which has neither). So no upgraded deployment ever ran the `CREATE INDEX`, and a forward `DROP INDEX IF EXISTS` would be a permanent no-op.

And per the `EXPLAIN` check on the sibling thread, the remaining `ledger_created_at`-ordered queries (`GetOldestLedger`, the `reconcile_oldest_cursor` job) plan fine without them — they're served by TimescaleDB's columnstore metadata indexes. Leaving the removal as the edit to the original migration.

internal/db/migrations/2025-06-10.3-operations.sql
```diff
@@ -34,14 +34,10 @@ CREATE TABLE operations (
...
--- Replaces the default single-column (ledger_created_at DESC) index TimescaleDB
--- auto-creates on the hypertable's partition column. The root GraphQL connection
--- sorts by (ledger_created_at DESC, id DESC), which the default index can't serve
--- for a top-N first page (it still needs a heapsort on id within each timestamp).
--- This composite index matches the API sort key directly and makes the auto-created
--- one redundant.
+-- TimescaleDB's default single-column index on the partition column. Retention drops chunks by
+-- range metadata, and no query path filters this table by bare ledger_created_at (reads go through
+-- the primary key or the operations_accounts table), so it has zero consumers.
 DROP INDEX IF EXISTS operations_ledger_created_at_idx;
```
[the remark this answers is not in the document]
aditya1702: This one doesn't apply here — no forward `DROP` migration is needed.

`idx_transactions_created_at_sort` / `idx_operations_created_at_sort` were **introduced in #667 and removed in this PR — both within the same unmerged hardening stack**. They were never on `main` (confirmed via `git grep` on `main`) and are absent from every deployed database (verified directly against the dev DB, which has neither). So no upgraded deployment ever ran the `CREATE INDEX`, and a forward `DROP INDEX IF EXISTS` would be a permanent no-op.

And per the `EXPLAIN` check on the sibling thread, the remaining `ledger_created_at`-ordered queries (`GetOldestLedger`, the `reconcile_oldest_cursor` job) plan fine without them — they're served by TimescaleDB's columnstore metadata indexes. Leaving the removal as the edit to the original migration.

JiahuiWho (approved):

PR #672, opened 2026-07-24
GraphQL API: encode state-change variants as 19 concrete types + schema conventions cleanup

aditya1702:
Breaking refactor of the GraphQL API (we're pre-release, so this is the last free window). The core problem: which `(category, reason)` pairs exist and which fields are set for each was **invisible in the schema** — you had to read processor code to know that e.g. `signerWeights.old` is null for `ADD`. This PR encodes the full variant structure in the schema, plus a holistic conventions cleanup.

## State changes: 9 opaque types → 19 exact types

`BaseStateChange.type` is renamed **`category`** (matches the enum + filter input). Types are split wherever nullability differs by reason, so every field's nullability is now exact and documented in SDL docstrings:

| Type | (category, reason) | Own fields |
|---|---|---|
| `BalanceChange` | BALANCE × DEBIT/CREDIT/MINT/BURN | `tokenId!`, `amount!`, `toMuxedId` — covers transaction fees: fee rows are (BALANCE, DEBIT) with `operation` null |
| `AccountCreatedChange` | ACCOUNT × CREATE | `creatorAddress!` — `account` is the created G-address or deployed C-address; `creatorAddress` is the funder or deployer |
| `AccountMergedChange` | ACCOUNT × MERGE | `destinationAddress!` |
| `SignerAddedChange` | SIGNER × ADD | `signerAddress!`, `newWeight!` |
| `SignerUpdatedChange` | SIGNER × UPDATE | `signerAddress!`, `oldWeight!`, `newWeight!` |
| `SignerRemovedChange` | SIGNER × REMOVE | `signerAddress!`, `oldWeight!` |
| `ThresholdChange` | SIGNATURE_THRESHOLD × UPDATE | `threshold: ThresholdLevel!` (LOW/MEDIUM/HIGH — which one changed), `oldThreshold!`, `newThreshold!`; one row per changed threshold |
| `AccountFlagsChange` | FLAGS × SET/CLEAR | `flags: [AccountFlag!]!` |
| `HomeDomainSetChange` | HOME_DOMAIN × SET | `homeDomain!` |
| `HomeDomainUpdatedChange` | HOME_DOMAIN × UPDATE | `oldHomeDomain!`, `newHomeDomain!` |
| `HomeDomainClearedChange` | HOME_DOMAIN × CLEAR | `oldHomeDomain!` |
| `DataEntryAddedChange` | DATA_ENTRY × ADD | `name!`, `value!` (base64) |
| `DataEntryUpdatedChange` | DATA_ENTRY × UPDATE | `name!`, `oldValue!`, `newValue!` (base64) |
| `DataEntryRemovedChange` | DATA_ENTRY × REMOVE | `name!`, `oldValue!` (base64) |
| `AllowanceChange` | ALLOWANCE × UPDATE (SEP-41 approve) | `tokenId!`, `spender!`, `amount!`, `expirationLedger!` |
| `TrustlineAddedChange` | TRUSTLINE × ADD | `tokenId`⊕`liquidityPoolId`, `limit!` |
| `TrustlineUpdatedChange` | TRUSTLINE × UPDATE | `tokenId`⊕`liquidityPoolId`, `oldLimit!`, `newLimit!` |
| `TrustlineRemovedChange` | TRUSTLINE × REMOVE | `tokenId`⊕`liquidityPoolId` |
| `BalanceAuthorizationChange` | BALANCE_AUTHORIZATION × SET/CLEAR | `tokenId`⊕`liquidityPoolId`, `flags: [TrustlineFlag!]` (null for SAC holders) |

(⊕ = exactly one set, documented in the docstrings.)

Highlights:
- **JSON-in-String is gone.** `signerWeights`/`thresholds`/`limit` blobs and `MetadataChange.keyValue` are typed fields now. SEP-41 approve persists the spender in the (previously unused) `spender_account_id` column.
- **Fees are `BalanceChange`.** One type covers the complete set of balance movements; transaction-fee rows are `(BALANCE, DEBIT)` with `operation` null (fees are per-transaction) and `toMuxedId` null. A CREDIT fee row cannot exist: the processor nets Soroban refunds into the charge, and a refund never exceeds the fee. `operation` is non-null on the other 14 types.
- **The (category, reason) convention is uniform: category = the entity, reason = the kind of update.** Thresholds no longer encode *which* threshold in the reason (that's the typed `threshold` field, backed by a new column); home domains and data entries no longer encode the *entity* in the reason — they get their own `HOME_DOMAIN` and `DATA_ENTRY` categories, each split by transition so nullability is exact, and the `METADATA` category is gone. The data-entry name moves into a `data_entry_name` column (projectable) and `key_value` payloads flatten to only the meaningful side of the transition — **no field in the API means anything by being empty any more**, and the empty string never reaches the database.
- **No-op writes no longer produce rows.** The Horizon home-domain effect fires whenever `SetOptions` carries a domain, with no check that the value changed, so re-writing the same domain used to emit a row — and on an account with no domain it emitted one claiming reason SET with an empty `newHomeDomain`. Emission now skips unchanged domains, matching how flags already skip no-op rows.
- **Allowances get their own `ALLOWANCE` category.** They aren't key-value metadata, so they no longer share METADATA with home domains and data entries; `(METADATA, UPDATE)` is now an invalid pair.
- **`SponsorshipChange` and the RESERVES category are removed end to end** — including the effects-processor emission and the three sponsorship-only DB columns. Sponsorship is an attribute of a reserve change, not a state change in its own right; a `ReservesChange`/`MinimumBalanceChange` type tracking required reserves can be added back once designed. Sponsorship *operations* are still indexed.
- **Contract deployments are `AccountCreatedChange`.** One type for all account creation: `account` distinguishes classic (G-address) vs contract (C-address), and a single non-null `creatorAddress` replaces `funderAddress`/`deployerAddress` (the two DB columns unify into `creator_account_id`).
- Type resolution dispatches purely on (category, reason) — no column discriminators — and **errors** on unknown pairs instead of returning null nodes. The discriminator columns (category, reason) are always selected regardless of client field selection.
- `flags` are typed enums (`AccountFlag`/`TrustlineFlag`) with deterministic order (the old string decoder iterated a map — order was random per request). A NULL `flags` column on `AccountFlagsChange` is a data-integrity error, never an empty list.
- **Semantically guaranteed old values are non-null** (`oldWeight`, `oldThreshold`, `oldHomeDomain`): the effects processor rejects effects whose ledger-entry pre-image is missing (malformed meta, dropped + logged) instead of degrading to a null field, and a locked master key's prior weight decodes as its true value, 0. A null `DataEntryChange.oldValue` now always means "created", never "couldn't reconstruct".
- **Pre-image lookups are entity-matched** (fixes a latent bug in the old-value extraction): the lookup previously returned the *first* pre-image of the requested entry type — multi-entry operations (merges carry two account pre-images) could source old values from the wrong entity. Accounts now match by address, trustlines by trustor + asset/pool, data entries by owner + name; regression tests pin all three.

## Other breaking changes

| Before | After |
|---|---|
| `Operation.operationType` | `Operation.type` |
| Filter `category`/`reason: String` | `StateChangeCategory` / `StateChangeReason` enums |
| Connections/edges nullable inconsistently | `edges: [X!]!`, `node: X!`, connection fields non-null everywhere |
| `TrustlineBalance.type: String` (`"credit_alphanum4"`) | `assetType: AssetType!` enum |
| `LiquidityPoolBalance.liquidityPoolId` | removed (`tokenId` is the pool id) |
| `ReservesChange`/`SponsorshipChange`, `RESERVES`, `SPONSOR`/`UNSPONSOR` | removed (see above) |
| `ContractDeployedChange`, `funderAddress`/`deployerAddress` | merged into `AccountCreatedChange.creatorAddress!` |
| `AllowanceChange` under `METADATA` | own `ALLOWANCE` category |
| `(SIGNATURE_THRESHOLD, LOW/MEDIUM/HIGH)` | `(SIGNATURE_THRESHOLD, UPDATE)` + `threshold: ThresholdLevel!` field |
| `(METADATA, HOME_DOMAIN/DATA_ENTRY)`, `METADATA` category | `HOME_DOMAIN` and `DATA_ENTRY` categories; home domains split into Set/Updated/Cleared and data entries into Added/Updated/Removed |
| `AUTHORIZATION` category (never emitted) | removed from Go enum + migration CHECK |
| `#` comments in SDL | `"""docstrings"""` (introspectable, visible in GraphiQL) |

## Complexity limit: default 5000 → 10000

gqlgen **sums** mutually exclusive inline fragments, so the full-detail account-history query selecting all 19 fragments measures **7301** at `first:100` (balances: 3801). `AccountTransactionEdge.operations/stateChanges` still have **no multiplier** (guard test unchanged). Regression tests pin both queries under 10000.

## Client (`pkg/wbclient`) — breaking API cleanup

- Mirrors the schema 1:1 (19 fragments/structs, `__typename` switch, enum consts). An offline test validates **every** client query against the SDL, so client/schema drift fails unit tests.
- `types.Operation.Type` mirrors the schema field name (`json:"type"`, no aliasing), so custom `QueryOptions.OperationFields` selecting the real schema field unmarshal correctly.
- Pagination/time/filter params move from positional args to structs (nil = defaults): `Page{First,Last,After,Before}`, `TimeRange{Since,Until}`, `StateChangeFilter{TransactionHash,OperationID,Category,Reason}` — e.g. `GetAccountStateChanges(ctx, addr, filter, timeRange, page)` instead of 12 positional params. Filter category/reason are enum-typed.
- GraphQL response errors are a typed `GraphQLErrors` slice (`errors.As`-able, `extensions.code` preserved, all messages joined) instead of just the first message string.
- Connections enforce the non-null contract end to end: missing/null `edges` or `node` is an unmarshal error, and a null connection field on an existing entity is rejected — never a silent nil.
- **Not-found is always a sentinel**: `ErrAccountNotFound`, `ErrTransactionNotFound`, `ErrOperationNotFound` — transaction/operation-scoped queries distinguish "doesn't exist" from "malformed response" instead of returning `(nil, nil)`.
- `TrustlineBalance.Code/Issuer` are non-null `string` (the SDL declares `code`/`issuer` non-null).

## Tests

- Dispatch-matrix unit tests cover every (category, reason, discriminator) → type, including invalid-pair error cases (dispatch rejects e.g. `(BALANCE, MERGE)`, `(METADATA, UPDATE)`); required fields (`amount`, trustline limits, `flags`) error on NULL backing data instead of returning `""`/`[]`.
- A column audit confirms no `state_changes` column is written without being exposed and none is exposed without a writer; the one orphan (`claimable_balance_id`, a sponsorship leftover) is dropped along with two pieces of dead builder plumbing.
- Integration validators rewritten to typed assertions (enum flags fail at compile time now, not string-compare); sponsorship ops keep their operation-level and non-sponsorship state-change coverage.
- All 14 GraphQL examples in the README validate against the SDL (SameResponseShape included).

## Deploy notes

| Action | Why |
|---|---|
| Raise `GRAPHQL_COMPLEXITY_LIMIT` env 6000 → 10000 | freighter full-detail query measures 7301 |
| Re-ingest dev/staging | migration edited in place (sql-migrate tracks by filename): CHECK values changed (−AUTHORIZATION, −RESERVES/SPONSOR/UNSPONSOR, +ALLOWANCE/HOME_DOMAIN/DATA_ENTRY, −METADATA), columns added (`threshold`, `data_entry_name`, `creator_account_id`) and dropped (sponsorship columns, `funder_account_id`/`deployer_account_id`, `claimable_balance_id`), `spender_account_id` newly written |
| freighter-backend-v2 adapts renamed fields when it adopts this API | `category`, `Operation.type`, split types — stellar/freighter-backend-v2#143 |

## Verification

`make check` + `make unit-test` (race, 32 packages) green; complexity regression suite green; integration suite compile-verified (full Docker run pending).


internal/serve/graphql/resolvers/utils.go
```diff
@@ -108,61 +108,87 @@ func NewConnectionWithRelayPagination[T any, C int64 | string](nodes []T, params
...
+// violate the non-null StateChangeEdge.node contract.
+func convertStateChangeTypes(stateChange types.StateChange) (generated.BaseStateChange, error) {
 	switch stateChange.StateChangeCategory {
 	case types.StateChangeCategoryBalance:
-		return &types.StandardBalanceStateChangeModel{
-			StateChange: stateChange,
+		if stateChange.OperationID == 0 {
+			return &types.FeeChangeModel{StateChange: stateChange}, nil
 		}
+		return &types.BalanceChangeModel{StateChange: stateChange}, nil
```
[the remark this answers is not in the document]
aditya1702: Fixed in 8fc58898: every dispatch arm now switches on reason — fee rows are restricted to DEBIT/CREDIT, operation-backed balance rows to DEBIT/CREDIT/MINT/BURN, and SIGNATURE_THRESHOLD/FLAGS/RESERVES/BALANCE_AUTHORIZATION each accept only their valid reasons. All other combinations reach the integrity error; dispatch-matrix tests cover the new error cases.

internal/serve/graphql/schema/pagination.graphqls
```diff
@@ -1,55 +1,69 @@
+"""Relay-style page of operations."""
 type OperationConnection {
-    edges: [OperationEdge!]
+    edges: [OperationEdge!]!
```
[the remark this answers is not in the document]
aditya1702: Fixed in 8fc58898: OperationConnection/OperationEdge (and every other connection type) now enforce the non-null contract — missing/null edges, null edge entries, and null nodes are unmarshal errors. Comments and tests updated to the strict contract.

internal/serve/graphql/schema/pagination.graphqls
```diff
@@ -1,55 +1,69 @@
...
 type OperationEdge {
-    node: Operation
+    node: Operation!
     cursor: String!
 }
 
+"""Relay-style page of state changes."""
 type StateChangeConnection {
-    edges: [StateChangeEdge!]
+    edges: [StateChangeEdge!]!
```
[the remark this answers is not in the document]
aditya1702: Fixed in 8fc58898: StateChangeConnection/StateChangeEdge reject missing/null edges and null nodes as malformed responses; tests rewritten accordingly.

internal/serve/graphql/schema/filters.graphqls
```diff
@@ -1,19 +1,14 @@
...
-  """Filter by state change category - returns only state changes with this category"""
-  category: String
-  
-  """Filter by state change reason - returns only state changes with this reason"""
-  reason: String
+  """Only state changes with this category."""
+  category: StateChangeCategory
+
+  """Only state changes with this reason."""
+  reason: StateChangeReason
```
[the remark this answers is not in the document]
aditya1702: Fixed in 8fc58898: GetAccountStateChanges now takes *types.StateChangeCategory / *types.StateChangeReason; callers and tests updated.

internal/serve/graphql/resolvers/resolver.go
```diff
@@ -100,17 +99,32 @@ func (r *Resolver) resolveRequiredString(field sql.NullString) string {
 	return ""
 }
 
-// resolveRequiredJSONBField resolves JSONB fields that return required strings
-// Marshals Go object to JSON string, returns empty string if field is nil
-func (r *Resolver) resolveRequiredJSONBField(field interface{}) (string, error) {
-	if field == nil {
-		return "", nil
+// resolveRequiredAddress resolves a required address field from the database.
+// Returns an error when the field is null, since the GraphQL schema declares it non-nullable.
```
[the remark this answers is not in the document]
aditya1702: Fixed in 8fc58898: resolveRequiredString returns an error on SQL NULL like the address/int16 helpers; amount and trustline limit resolvers propagate it, with tests for each null path.

internal/data/statechanges.go
```diff
@@ -29,10 +29,16 @@ type StateChangeWriter interface {
 
 var _ StateChangeWriter = (*StateChangeModel)(nil)
 
+// stateChangeMandatoryColumns is always selected regardless of the GraphQL field
+// projection: the cursor/key columns, plus the discriminators (state_change_category,
+// state_change_reason, deployer_account_id) that convertStateChangeTypes in the
+// resolvers package needs to pick the concrete GraphQL type for each row.
+var stateChangeMandatoryColumns = []string{"to_id", "operation_id", "state_change_id", "account_id", "ledger_created_at", "state_change_category", "state_change_reason", "deployer_account_id"}
```
JakeUrban: Why is `deployer_account_id` included? All the other columns are general-puropose

internal/serve/graphql/schema/statechange.graphqls
```diff
@@ -1,41 +1,71 @@
...
 
-type AccountChange implements BaseStateChange {
-  type:                       StateChangeCategory! @goField(forceResolver: true)
+"""
+The transaction fee debited from (or refunded to) the fee-paying account.
+Pairs: (BALANCE, DEBIT) for a net fee charge, (BALANCE, CREDIT) for a net
+refund. The only state change with no associated operation; `operation` is
+always null.
+"""
+type FeeChange implements BaseStateChange {
```
JakeUrban: Fee refunds are an implementation detail of the protocol right? We would never actually index a FeeChange with a `CREDIT` reason would we?

JakeUrban: Separately, what is the rationale for `FeeChange` being a distinct change type, instead of it being a variation of `BalanceChange`? I recognize that you have stronger typing (`BalanceChange` has a non-null `operation` field, `FeeChange` omits `toMuxedId`) but I'm not sure if this is worth the potential confusion it may cause. If we instead make `BalanceChange` include the transaction fee case, then `BalanceChange` is the only type required to get the complete set of balance changes.

internal/serve/graphql/schema/statechange.graphqls
```diff
@@ -44,108 +74,359 @@ type AccountChange implements BaseStateChange {
...
+  funderAddress:              String! @goField(forceResolver: true)
 }
 
-type SignerThresholdsChange implements BaseStateChange {
-  type:                       StateChangeCategory! @goField(forceResolver: true)
+"""
+A smart-contract deployment. `account` is the deployed contract address.
+Pair: (ACCOUNT, CREATE).
+"""
+type ContractDeployedChange implements BaseStateChange {
```
JakeUrban: It may be worth considering making contract deployment a variation of `AccountCreatedChange`, rather than having a distinct type. Again, the tradeoff is weak typing (`deployerAddress` and `funderAddress` have to be nullable) but the benefit is having one change type for creation of all account types.

I'm less opinionated about this one compared to `FeeChange`. I think it could also be confusing to encounter contract deployments when trying to monitor stellar account creation, so I'll defer to you.

JakeUrban: Another thought: right now we treat all contracts as potential smart accounts. Instead, we could only treat contracts as accounts if they include `__check_auth()` in their implementation.

Maybe this is out-of-scope of this PR, but worth discussing further.

internal/serve/graphql/schema/statechange.graphqls
```diff
@@ -44,108 +74,359 @@ type AccountChange implements BaseStateChange {
...
+  """New value, base64-encoded; null on removal."""
+  newValue:                   String @goField(forceResolver: true)
+}
+
+"""
+A SEP-41 allowance approval: `account` (the token holder) authorized `spender`
+to transfer up to `amount` of the token on its behalf.
+Pair: (METADATA, UPDATE).
+"""
+type AllowanceChange implements BaseStateChange {
```
JakeUrban, line 332: It may make more sense for allowances to have their own category. `METADATA` so far is just classic account data entries + the account's home domain which is basically a first-class data entry. Allowances aren't key-val pairs like data entries are.

internal/serve/graphql/schema/statechange.graphqls
```diff
@@ -44,108 +74,359 @@ type AccountChange implements BaseStateChange {
...
+"""
+A base-reserve sponsorship established or released. Each sponsorship event
+emits one state change per perspective: on the sponsoring account's change
+`sponsoredAddress` is set (account and trustline sponsorships only); on the
+sponsored account's change `sponsorAddress` is set. At most one of the entity
+fields (tokenId, liquidityPoolId, claimableBalanceId, dataName, signerAddress)
+identifies what is sponsored; all are null for whole-account sponsorships.
+Pairs: (RESERVES, SPONSOR), (RESERVES, UNSPONSOR).
+"""
+type SponsorshipChange implements BaseStateChange {
```
JakeUrban: If we're preparing for a v1.0 API release I think we should remove this change type.

I've historically advocated for replacing this change with a `ReservesChange` type, where the state it tracks is the number of reserves an account requires. Alternatively we could have a `MinimumBalanceChange` that tracks the minimum balance an account requires. Both are different ways to represent the same thing.

Sponsorship is an attribute of these changes, not the change itself. Meaning, when an account's reserves / minimum balance increases, it is either because it is sponsoring a subentry for another account or the account itself has a new subentry. In the same way, an account's reserves / minimum balance decreases when it revokes sponsorship from another account's subentry or the account itself no longer has the subentry.

Lets add one of these change types back later, wdyt?

aditya1702: @JakeUrban  Agreed, we can always add new state changes later so I will remove this for now and we will redesign it cleanly

pkg/wbclient/queries.go
```diff
@@ -22,7 +22,7 @@ const (
 
 	defaultOperationFields = `
 		id
-		operationType
+		operationType: type
```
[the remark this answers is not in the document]
aditya1702: Fixed in 74d182b0: the client field is renamed to `Type` (`json:"type"`) mirroring the schema exactly, and the `operationType:` alias is dropped from the default selection — custom `OperationFields` selecting `type` now unmarshal correctly.

internal/serve/graphql/schema/transaction.graphqls
```diff
@@ -1,22 +1,26 @@
...
-  
-  # GraphQL Relationships - these fields require resolvers
-  # @goField(forceResolver: true) tells gqlgen to always generate a resolver
-  # even if the Go struct has a matching field
-  operations(first: Int, after: String, last: Int, before: String):     OperationConnection
-  
-  # Related accounts - uses resolver with dataloader for efficiency
+
+  """Operations contained in this transaction."""
+  operations(first: Int, after: String, last: Int, before: String): OperationConnection!
```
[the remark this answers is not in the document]
aditya1702: Fixed in 74d182b0: `TransactionByHash` is a nullable root now — a null root returns `ErrTransactionNotFound`, and a null `operations`/`stateChanges` connection on an existing transaction is a malformed-response error, matching the account methods. New client tests cover all paths.

internal/serve/graphql/schema/operation.graphqls
```diff
@@ -1,22 +1,28 @@
...
-  # Related accounts - uses resolver with dataloader for efficiency
+
+  """Accounts that participated in this operation."""
   accounts:        [Account!]! @goField(forceResolver: true)
-  
-  # Related state changes - uses resolver to fetch associated changes
-  stateChanges(first: Int, after: String, last: Int, before: String):    StateChangeConnection
+
+  """State changes produced by this operation."""
+  stateChanges(first: Int, after: String, last: Int, before: String): StateChangeConnection!
```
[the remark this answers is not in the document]
aditya1702: Fixed in 74d182b0: `OperationByID` is nullable with an `ErrOperationNotFound` sentinel; a nil `stateChanges` connection on an existing operation is rejected. `GetTransactionByHash`/`GetOperationByID` also return the sentinels instead of (nil, nil) for consistency.

internal/serve/graphql/resolvers/statechange.resolvers.go
```diff
@@ -8,457 +8,831 @@ package resolvers
...
+}
+
+// Flags is the resolver for the flags field. Null when the row carries no flags;
+// otherwise the decoded account flags.
+func (r *accountFlagsChangeResolver) Flags(ctx context.Context, obj *types.AccountFlagsChangeModel) ([]types.AccountFlag, error) {
+	if !obj.Flags.Valid {
+		return nil, nil
+	}
+	return types.DecodeAccountFlags(obj.Flags.Int16), nil
+}
```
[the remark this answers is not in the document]
aditya1702: Fixed in 74d182b0: a NULL flags column on `AccountFlagsChange` now returns a data-integrity error naming the row identifiers instead of serializing an empty list; test added for the null path. (`BalanceAuthorizationChange.flags` stays nullable — SAC holders legitimately have none.)

internal/serve/graphql/README.md
```diff
@@ -573,33 +649,49 @@ query GetBalanceChanges {
...
-          ... on SignerChange {
+          ... on SignerUpdatedChange {
             signerAddress
-            signerWeights
+            oldWeight
+            newWeight
+          }
+
+          ... on TrustlineAddedChange {
+            tokenId
```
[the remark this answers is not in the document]
aditya1702: Fixed in 74d182b0: the example aliases the `TrustlineAddedChange` occurrence (`trustlineTokenId: tokenId`) so it satisfies SameResponseShape. All 14 GraphQL examples in the README were validated against the SDL with gqlparser; this was the only violation.

internal/data/statechanges.go
aditya1702: This was originally included in order to generate the ContractDeployed state change. However I unified the account creation and contract deployed change so this is now not required

internal/serve/graphql/schema/statechange.graphqls
aditya1702: @JakeUrban Yeah my original intention to break it up was to have strict non-nullable columns however you are right that we dont need to be that granular. I removed the FeeChange state change type

internal/serve/graphql/schema/statechange.graphqls
aditya1702: @JakeUrban I combined the two state changes into 1 by having a `createrAccountID` that is the funder address for classic and deployer account for contract deployments. I think that is cleaner here

internal/db/migrations/2025-06-10.4-statechanges.sql
```diff
@@ -9,15 +9,14 @@ CREATE TABLE state_changes (
     state_change_category TEXT NOT NULL CHECK (
         state_change_category IN (
             'BALANCE', 'ACCOUNT', 'SIGNER', 'SIGNATURE_THRESHOLD',
-            'METADATA', 'FLAGS', 'TRUSTLINE', 'RESERVES',
-            'BALANCE_AUTHORIZATION', 'AUTHORIZATION'
+            'DATA_ENTRY', 'HOME_DOMAIN', 'ALLOWANCE', 'FLAGS', 'TRUSTLINE',
```
JiahuiWho, line 12: Just checking: Are you editing this migration in place because it hasn't been deployed to prod yet?

pkg/wbclient/client.go
```diff
@@ -252,16 +264,23 @@ func (c *Client) GetTransactionByHash(ctx context.Context, hash string, opts ...
...
+	if data.TransactionByHash == nil {
+		return nil, fmt.Errorf("%w: %s", ErrTransactionNotFound, hash)
+	}
+
 	return data.TransactionByHash, nil
 }
 
+// GetAccountByAddress fetches a single account by its address. Pass a
+// *QueryOptions to restrict the account fields fetched; omit it (or pass nil)
+// for the default field set.
```
JiahuiWho: `GetAccountByAddress` still returns `(nil, nil)` on not-found, while every sibling lookup in this PR was updated to return `ErrAccountNotFound`. Should check `data.AccountByAddress == nil` here too for the sentinel-error guarantee to actually hold everywhere.

internal/serve/graphql/resolvers/statechange_resolvers_test.go
```diff
@@ -19,299 +18,745 @@ import (
...
+			sc:   types.StateChange{StateChangeCategory: types.StateChangeCategoryTrustline, StateChangeReason: types.StateChangeReasonUpdate},
+			want: &types.TrustlineUpdatedChangeModel{},
+		},
+		{
+			name: "TRUSTLINE remove",
+			sc:   types.StateChange{StateChangeCategory: types.StateChangeCategoryTrustline, StateChangeReason: types.StateChangeReasonRemove},
+			want: &types.TrustlineRemovedChangeModel{},
+		},
+		{
+			name: "BALANCE_AUTHORIZATION set",
```
JiahuiWho, line 168: Missing the `CLEAR` case?  `utils.go` maps both `SET` and `CLEAR` to `BalanceAuthorizationChangeModel`. Let's add it so the dispatch table stays exhaustive.

internal/indexer/processors/state_change_builder.go
```diff
@@ -147,36 +154,24 @@ func (b *StateChangeBuilder) WithTokenType(tokenType types.ContractType) *StateC
...
-}
-
 // WithOperationID sets the operation ID
 func (b *StateChangeBuilder) WithOperationID(operationID int64) *StateChangeBuilder {
 	b.base.OperationID = operationID
 	return b
 }
 
-// WithClaimableBalanceID sets the claimable balance ID for sponsorship state changes
+// WithClaimableBalanceID sets the claimable balance ID
```
JiahuiWho: `ClaimableBalanceID`/`WithClaimableBalanceID` look dead now that sponsorship state changes are gone, only a unit test still calls the setter. Considering drop it?

internal/indexer/processors/effects_test.go
```diff
@@ -249,90 +250,10 @@ func TestEffects_ProcessTransaction(t *testing.T) {
 		assert.Equal(t, toid.New(12345, 1, 1).ToInt64(), changes[0].OperationID)
 		assert.Equal(t, uint32(12345), changes[0].LedgerNumber)
 		assert.Equal(t, time.Unix(12345*100, 0), changes[0].LedgerCreatedAt)
-		assert.Equal(t, types.StateChangeCategoryMetadata, changes[0].StateChangeCategory)
-		assert.Equal(t, types.StateChangeReasonDataEntry, changes[0].StateChangeReason)
-		assert.Equal(t, types.NullableJSONB{"hello": map[string]any{"old": ""}}, changes[0].KeyValue)
-	})
-	t.Run("Sponsorship - reserves sponsorship created/updated/revoked", func(t *testing.T) {
```
JiahuiWho, line 256: This is also the only test using a real `CreateAccount` XDR fixture, which asserted the synthesized `SIGNER`/`ADD` effect. Consider keeping a trimmed non-sponsorship assertion here?

internal/indexer/processors/effects.go
```diff
@@ -143,19 +150,33 @@ func (p *EffectsProcessor) ProcessOperation(_ context.Context, opWrapper *Transa
 		// Threshold effects: track changes to account signature thresholds (low/medium/high)
 		case EffectAccountThresholdsUpdated:
 			changeBuilder = changeBuilder.WithCategory(types.StateChangeCategorySignatureThreshold)
-			stateChanges = append(stateChanges, p.parseThresholds(changeBuilder, &effect, changes)...)
+			thresholdChanges, err := p.parseThresholds(changeBuilder, &effect, changes)
+			if err != nil {
+				log.Debugf("processor: %s: failed to parse threshold effects: effectType: %s, address: %s, txHash: %s, opID: %d, err: %v", p.Name(), effect.TypeString, effect.Address, txHash, opWrapper.ID(), err)
+				continue
+			}
```
JiahuiWho, line 157: These "reject malformed pre-image" paths (line 169, 226 etc) only `log.Debugf` + `continue`. 

Given that:
>  ... rejects effects whose ledger-entry pre-image is missing (malformed meta, dropped + logged) instead of degrading to a null field

 a silent drop in production is operationally indistinguishable from the old bug if it's ever hit on real ledger data. Worth adding a counter to make it observable?

internal/db/migrations/2025-06-10.4-statechanges.sql
aditya1702: @JiahuiWho Yes directly making change to schema since WB is not yet deployed - this is the last time we will directly edit the schema

pkg/wbclient/client.go
aditya1702: Good catch — fixed in f53a702e. It was the only one of twelve client query methods still doing this, so the sentinel guarantee didn't actually hold everywhere. It returns `ErrAccountNotFound` now and the doc comment says so; added a not-found test alongside the other single-entity getters. Worth noting it had zero callers, so nothing depended on the old `(nil, nil)` contract.

internal/serve/graphql/resolvers/statechange_resolvers_test.go
aditya1702: Fixed in 61662b8a — and the audit turned up a second gap you didn't hit: `(BALANCE, BURN)` was missing too, so the table covered 24 of the 26 pairs `convertStateChangeTypes` accepts. Both added, and I corrected the section comment above them, which claimed SIGNATURE_THRESHOLD/FLAGS/BALANCE_AUTHORIZATION accept "any reason" — they each accept only specific ones.

internal/indexer/processors/state_change_builder.go
aditya1702: Confirmed and removed end to end in b3fadfff — setter, struct field, the BatchCopy column and value, and the column itself (in place, since this branch already requires a re-ingest). It was writing SQL NULL on every single row.

I extended the check to every other `state_changes` column as well, since one orphan suggested there might be more. Result: this was the only one. Nothing else is written-without-being-exposed, and nothing is exposed-without-a-writer — every GraphQL field on all 19 types resolves from a column with a real production writer. The three symbols adjacent to this one are all live: `WithSigner` (effects.go), `WithLiquidityPoolID` (trustlines), `WithSpender` (sep41).

The audit did find two non-column leftovers of the same kind, removed in the same commit: `ContractType` was computed per event and then discarded on `Build()` (no db tag, `json:"-"`, no reader), and the builder's `IngestedAt` assignment was dead because BatchCopy omits that column and Postgres `DEFAULT NOW()` supplies the value.

internal/indexer/processors/effects_test.go
aditya1702: You're right that real coverage was lost — restored in 61662b8a. One nuance: SIGNER/ADD itself is still asserted in the `SetOption` subtest, so what actually went missing was specifically the *account-creation* path into it, which was left verified only by the Docker integration suite.

The new subtest asserts that change with a minimal hand-built CreateAccount meta fixture rather than restoring the opaque base64 group the removal took with it — the effect derives from the operation's destination, so the meta only needs valid operation changes. It produces exactly one change now (the eight sponsorship rows are gone with the category), and the test pins the signer address, weight 1, and the absent old weight.

internal/indexer/processors/effects.go
aditya1702: Agreed, and you put your finger on the real problem: those sites log at **debug**, which is off in production, so a drop left no trace at all — the PR description's claim that they're "dropped + logged" was only true in development. All seven are `log.Warnf` now (f53a702e), so the error text and the account/tx/op context actually reach prod logs.

We decided against a counter for the moment: each site guards several distinct causes (the signer one alone covers weight conversion, a missing pre-image, and a signer absent from the pre-image), so a per-site `reason` label would have been imprecise, and the warn line carries the specific error. Easy to add later if these turn out to fire on real ledger data. Left the two intentional bare `continue`s alone — the home-domain no-op and the untracked-effect `default` — since those aren't drops.

pkg/wbclient/client.go
```diff
@@ -273,13 +292,16 @@ func (c *Client) GetAccountByAddress(ctx context.Context, address string, opts .
 	return data.AccountByAddress, nil
```
aristidesstaffieri, line 297: **`GetAccountByAddress` can still return `(nil, nil)` for a missing account.**

This PR gives every other lookup a typed not-found error — `GetTransactionByHash` returns `ErrTransactionNotFound`, `GetOperationByID` returns `ErrOperationNotFound`, and all the account-scoped queries (`GetAccountTransactions`, `GetAccountStateChanges`, …) return `ErrAccountNotFound` when `data.AccountByAddress == nil`. This is the one getter left that returns a nil pointer with a nil error.

The `ErrAccountNotFound` doc comment also documents the contract as "returned by account-scoped queries when the account does not exist … Use `errors.Is(err, wbclient.ErrAccountNotFound)`", so a caller following the package's own convention will panic:

```go
acct, err := c.GetAccountByAddress(ctx, addr)
if err != nil {
    return err
}
use(acct.Address) // nil-pointer dereference when the account doesn't exist
```

Suggest adding the same guard the siblings have:

```go
if data.AccountByAddress == nil {
    return nil, fmt.Errorf("%w: %s", ErrAccountNotFound, address)
}
```

(and noting it in the method's doc comment like the others do).

aditya1702: @aristidesstaffieri this has already been resolved: https://github.com/stellar/wallet-backend/pull/672/changes/f53a702e46558fd77e4307cbe26c7d9415c84bca

aristidesstaffieri (approved):

PR #678, opened 2026-08-06
Add streaming-loadtest ledger backend for apply-load replay

aditya1702:
### What
- New dev-only `LEDGER_BACKEND_TYPE=streaming-loadtest`: reads ledger meta from named pipes written by `stellar-core apply-load` (one FIFO per tx profile), merges the streams into one ledger sequence, renumbers so restarts on either side never need a DB reset, and stamps advancing close times (apply-load emits closeTime 0).
- Two flags: `--loadtest-meta-pipe-paths`, `--loadtest-ledger-close-duration` (pacing; FIFO backpressure throttles the generators to match).
- This backend type skips the history-archive connect and checkpoint bootstrap (apply-load publishes no archive); ingestion starts from ledger 1 on an empty DB.

### Why
Load-test ingestion at rates the dev loadtest network can't produce (thousands of soroban TPL, sub-second ledgers). Only the ledger backend is swapped — the live ingest loop, processors, and persistence are unchanged, so results predict production behavior.

### Testing
- Unit tests for renumbering, 3-pipe merge, per-pipe generator restart, truncated frames, pacing, retry idempotence (green under `-race`).
- Opt-in corpus test (`STREAMING_LOADTEST_CORPUS`) replayed real v27 sac/custom_token/soroswap apply-load output: 200 merged ledgers, ~57k txs, all parsed by the production transaction reader.
- `rpc`/`datastore` behavior unchanged (validation branches exercised manually).

### DIFF
--- a/cmd/ingest.go
+++ b/cmd/ingest.go
@@ -85,11 +85,11 @@ func (c *ingestCmd) Command() *cobra.Command {
 		},
 		{
 			Name:        "archive-url",
-			Usage:       "Archive URL for history archives",
+			Usage:       "Archive URL for history archives. Required for every backend except 'streaming-loadtest', which runs without a history archive.",
 			OptType:     types.String,
 			ConfigKey:   &cfg.ArchiveURL,
 			FlagDefault: "https://history.stellar.org/prd/core-testnet/core_testnet_001/",
-			Required:    true,
+			Required:    false,
 		},
 		{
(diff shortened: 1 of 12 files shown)


aditya1702:
Superseded by #679, which targets blend/pr6-integration-tests (what dev currently runs) so the backend can be tested against the live dev deployment.

PR #684, opened 2026-08-13
Pipeline live ingestion into fetch/process/persist with seven parallel write streams

aditya1702:
**How to review (~2 h):**

1. Read "Seven sibling write streams" below — it's the core design.
2. Read the commits in order. The last 11 are an adversarial review round: 4 correctness, 3 deploy-safety, 4 hygiene.

**What this does:** the three ingestion stages ran one after another, so a ledger cost the *sum* of the stages (7.7–8.4 s on the rig; 8.2 s of that was serialized inserts). Now the stages overlap and persist writes through seven parallel transactions. A ledger costs only the *slowest* stage.

Also fixed along the way: one silent transaction-loss bug, three pathological DB write patterns, an undetectable ingestion-hang class, two batched-mode silent-loss bugs.

**Why now:** block time is headed 2 s → 1 s → 600 ms at constant volume. The contract: the slowest stage's p99 stays under block time.

## The pipeline

While ledger N persists, N+1 processes and N+2 fetches.

- Persist is the only stage that writes. Strictly in ledger order (cursor + CAS advance N-1 → N). Rejects ledger 0 at entry (`seq-1` underflows unsigned).
- Buffers rotate through a free channel, cleared only at take. A persist still reading is never torn.
- The process→persist queue carries only the buffer, the close time, and timings. The decoded `LedgerCloseMeta` and materialized transactions never ride it.
- Any stage error stops the pipeline. Restart resumes from the cursor.

Metrics:

- `wallet_ingestion_duration_seconds` sums stage times (wall time would count queue wait).
- `..._phase_duration_seconds{phase="insert_into_db"}` records the **true wall time of each persist commit** — never divided by batch size. Its 0.6 s / 1.0 s grading buckets read real commit latency exactly when persist is behind.
- `persist_batch_size` carries the coalescing.

## Seven sibling write streams

Each sibling owns its own tables. Every FK is checked inside the transaction that writes its parent rows.

| Sibling | Tables |
|---|---|
| 5 × COPY | `transactions`, `transactions_accounts`, `operations`, `operations_accounts`, `state_changes` |
| balances | `native_balances`, `liquidity_pools`, `liquidity_pool_balances` |
| trustlines | `trustline_assets`, `trustline_balances` |
| coordinator | contracts, classification, protocol current state, SAC balances, cursor |

The five bulk tables have one authoritative definition: `data.BulkCopyTables`. Siblings, startup reconciliation, hypertable settings, and the backfill recompressor all consume it. A test pins the sibling set to the list — a table missing from reconciliation would crash-loop re-ingest on PK collisions.

**`state_changes` has two producers, one transaction.** Protocol history rows are `state_changes` rows too:

- The ledger's own rows stream from the sibling goroutine.
- Protocol history streams from the coordinator goroutine (CAS-gated).
- Both write through the *one* `state_changes` sibling transaction, under a mutex.

Why: two of our own transactions inserting into the same hypertable can deadlock **undetectably** at a chunk boundary. TimescaleDB serializes chunk creation while the coordinator is blocked in Go — where Postgres's deadlock detector can't see the cycle. The CAS stays on the coordinator: siblings commit strictly first, so a committed protocol cursor still implies committed history rows.

The commit order is the whole safety story:

1. Nothing commits until every stream succeeds.
2. Siblings commit first (sub-ms). The coordinator commits strictly last — its cursor decides which ledgers exist.
3. Failure after the first sibling commit is fatal (`ErrPartialPersist`). COPY has no `ON CONFLICT`; a retry would collide.
4. The barrier runs on `context.WithoutCancel`. A SIGTERM mid-barrier cannot split the batch into a fake fatal `ErrPartialPersist` — a batch commits fully or not at all.
5. Crash recovery is one startup step: `DeleteRowsAboveLedger` removes bulk rows past the cursor (protocol history included — same table, same TOID predicate). Balance upserts are idempotent. The rare unbounded-scan fallback now logs itself.

Three deliberate trades:

- Siblings run `synchronous_commit = off`. Durability unchanged — the coordinator's synchronous commit flushes all earlier WAL. Removes up to 7 serialized fsync waits per persist.
- SAC balances stay on the coordinator. Classification writes their parent table there; a second writer could deadlock at the commit barrier.
- **Torn reads across the barrier are accepted.** Commits are sequential, so a reader can briefly see a transaction without its participant links. The window is one commit round-trip per sibling, the cursor still commits last, re-reads converge. Closing it would mean bounding every read path by the cursor.

## Batching (off by default)

`--live-persist-max-batch-size`, **default 1 = exactly the old behavior**. Mainnet/testnet keep 1; only the loadtest rig raises it. When persist falls behind, up to N ledgers fold into one commit set.

Four rules keep it sound:

1. A ledger with *unseen* classification inputs opens its own batch — its plan must see the previous commit. The seen-set is **binding-aware**: a contract re-observed bound to a *different* wasm counts as unseen, even if that wasm is already classified.
2. Mid-batch ledgers run with no plan. No plan means **"no reclassification", not "nothing is classified"** — rule 1 guarantees their buffered contracts are pure re-observations, so committed membership stands. (The bug: dropping it silently discarded a re-observed contract's events. Regression-tested at unit and batched-persist level.)
3. The rotation holds 2N+1 buffers — a buffer stays checked out until its batch commits.
4. Membership lookups read the coordinating transaction, not the pool — a contract classified at the batch head is still uncommitted while a mid-batch ledger carries its events. Regression-tested.

Housekeeping: the seen-sets are skipped entirely at batch size 1 and bounded by cap-and-clear (a miss only makes one cut conservative).

One accepted trade, documented in code: best-effort RPC enrichment (SEP-41 token metadata) retries on classification passes, which pause under a sustained backlog and resume when the pipeline catches up. Gating batches on enrichment would disable batching wholesale on RPC-less deployments like the rig.

## DB write-pattern fixes

1. **Balance upserts sorted + guarded.** Were map-ordered and rewrote unchanged rows (60+ buffer touches per row; the table sat at 97% free space). Now PK-sorted with `IS DISTINCT FROM`: an identical row writes nothing.
2. **State-change lookups seek the PK.** TOID makes `to_id` derivable from `operation_id`, so lookups bind the PK prefix: **7 buffers instead of 307**. Index swap ships as a real forward migration — see the migration note below.
3. **Startup reconciliation bounded by close time.** Was >10 min of chunk scans on a deep DB. Close times are monotone, so the cursor's close time prunes old chunks statically.
4. **New deferred FK `liquidity_pool_balances.pool_id → liquidity_pools`.** An orphaned pool-share balance is always an ingestion bug; now it fails loudly at write time. Applied deploy-safely: orphan-cleanup `DELETE` → `ADD CONSTRAINT ... NOT VALID` (brief metadata-only lock) → `VALIDATE` in its own transaction, so ingestion keeps writing through the scan and a pre-existing orphan can't halt a deploy. The checkpoint loader was made symmetric: share rows of pools the pool capture skips are dropped with their pool.

## Correctness fix: transactions were keyed by hash

Participants were keyed by ToID, transactions by hash. Merged loadtest ledgers can repeat one envelope at several positions — distinct ToIDs, one hash. The duplicates were silently dropped and their participant links COPYed a zero timestamp into a year-0001 chunk.

Fix: both maps share the ToID key, and `BatchCopyAccounts` (now one shared helper for both link tables) errors on a link row with no parent. Real networks can't repeat a hash — latent in production, live only on the rig.

## Also in here

- Indexer pools sized off `GOMAXPROCS`, not `NumCPU` — a 6-core pod on a 36-core node ran 72 workers on 6 threads.
- DB pool default 10 → 12. Persist pins 9 connections (7 siblings + coordinator + advisory lock) and `pgxpool.Acquire` has no timeout, so live ingestion now **fails fast at startup** if `--db-max-conns` can't fit the barrier — instead of self-deadlocking silently.
- Phase histogram re-bracketed onto 0.5/0.6/0.75/1 s — the old buckets merged 0.9 s and 1.9 s stages at p99.
- Removed: dead fields, an inert flag, a hand-rolled retry ladder, never-read COPY row counts. Link-table COPY metrics get their own labels.

## Measured

| | Before | After |
|---|---|---|
| Insert phase | 8.2 s serialized | 3.4 s (largest single COPY) |
| Pipelined work/ledger, fresh DB | — | ~1.4–1.5 s |
| `BatchGetByOperationID` buffers | 307 | 7 |
| Startup reconciliation, deep DB | >10 min | partition-bounded seek |

Backfill is untouched (keeps the single-transaction path).

> [!NOTE]
> **No manual DDL needed.** `2025-06-10.4-statechanges.sql` (already shipped) is edited in place so the baseline file stays the single readable statement of the table's final shape. A new idempotent migration (`2026-08-21.0`) converges environments that applied the old shape: creates `idx_state_changes_account_id`, drops the two superseded indexes, no-ops on fresh databases and on the loadtest DB (already swapped by hand).
>
> One caveat: TimescaleDB rejects `CREATE INDEX CONCURRENTLY` on hypertables, so the CREATE takes a write-blocking lock for the build. On an environment with a deep `state_changes`, run the migration step in a maintenance window.

---

Second of four PRs from the loadtest campaign (replaces #682). **Review order: #679 → this → #685; #686 is independent.** Combined rig result: **18,458 tx/s, process p99 0.992 s**; PRs not separately re-benchmarked. This PR took the insert phase off the critical path.


aditya1702:
Superseded by six smaller PRs. Same work, same final tree — `persist/6-hardening` is byte-identical to this branch's tip.

Review them in order:

1. #704 `persist/1-schema` — state_changes index reshape + pool-share FK (6 files)
2. #705 `persist/2-fixes` — three independent fixes (5 files)
3. #706 `persist/3-pipeline` — fetch/process/persist pipeline (12 files)
4. #707 `persist/4-seams` — sibling COPY streams + batched commits (9 files)
5. #708 `persist/5-write-families` — one sibling per write family (25 files)
6. #709 `persist/6-hardening` — correctness, deploy safety, hygiene (13 files)

The streaming-loadtest ledger backend was removed before the split (it now runs from its own deployment), which is why this PR shows 44 files rather than the original 51.

The branch stays put — #685 is still based on it.

PR #686, opened 2026-08-13
Fix the DB query-duration buckets, the oldest-ledger index path, and compression scheduling

aditya1702:
**Four independent fixes, one commit each — reviewable in any order. Review time: ~20 min. One manual-DDL warning below.**

Found while profiling live ingestion: a metric that clipped exactly the queries being optimized, a missing index that made an hourly job 45% of all DB disk reads, five compression policies firing at the same instant, and a phase histogram whose counts stopped matching its siblings.

## 1. Query-duration buckets clipped at 0.38 s

`wallet_db_query_duration_seconds` topped out at 0.38 s: every multi-second bulk COPY landed in `+Inf` and p99 panels flattened. Widened to 0.1 ms – ~17.7 s.

## 2. Oldest-ledger lookup had no ordered path

`SELECT ledger_number FROM transactions ORDER BY ledger_created_at ASC LIMIT 1` (backfill gap detection + the hourly `reconcile_oldest_cursor` job) had no index to walk — an earlier audit dropped TimescaleDB's default partition-column index after counting only WHERE-clause consumers, missing that this query consumes the index's *ordering*. The planner pulled the first row from every chunk: **12–187 s on an 87 GB DB; the hourly job was 45% of all DB disk-read time.**

Fix: keep the default index; drop the `to_id` tie-break from both query sites (close times increase strictly — the tie-break could only force an extra sort). Tested by running the reconciliation job against out-of-order chunks.

> [!WARNING]
> **Manual DDL on already-migrated databases** (fresh DBs get the index from `create_hypertable`). Applied on the loadtest DB; **pending on dev/staging/prod**:
>
> ```sql
> CREATE INDEX transactions_ledger_created_at_idx ON transactions (ledger_created_at DESC);
> ```

> [!NOTE]
> `2025-06-10.2-transactions.sql` (already shipped) is edited in place, not superseded. Deliberate: wallet-backend is pre-release, and the baseline files stay the single readable statement of each table's final shape.

## 3. Compression policies all fired at once

All five hypertables' columnstore policies were auto-created on identical schedules and compressed their just-closed chunks concurrently — an I/O storm that took **persist p50 from 0.68 s to 3.0 s for 18 of every 60 minutes** on the rig. Each policy is now anchored to a distinct slot offset on the schedule grid, so at most one fires at a time; jobs already on their slot are untouched across restarts.

## 4. insert_into_db now observes once per ledger, at full commit duration

The phase histogram recorded one observation per persist *commit*, so under batching its counts tracked commits while every other phase tracked ledgers — panels comparing phases (or grading p99 against ledger close time) mixed units.

Each ledger now observes the **full, undivided wall time of the commit that carried it**:

- counts are per-ledger again, comparable with `process_ledger`;
- slow batched commits stay fully visible (never divided by batch size — dividing is what used to hide overload: a 4 s commit for 8 ledgers read as eight healthy 0.5 s shares);
- a slow commit weighs in once per ledger it delayed, which is the right weighting for the 0.6 s / 1.0 s grading buckets.

`persist_batch_size` still carries the coalescing alongside.

---

Stacked: **#679 → #684 → #685 → this → #699.** Combined rig result: **18,458 tx/s, process p99 0.992 s**; PRs not separately re-benchmarked.


internal/data/ingest_store.go
```diff
@@ -252,10 +279,19 @@ func (m *IngestStoreModel) GetLedgerGaps(ctx context.Context, startLedger, endLe
...
+// stops instead of reading the first row of every chunk.
+// The sort key is ledger_created_at alone: stellar-core enforces strictly increasing ledger close
+// times, so every row sharing a ledger_created_at belongs to one ledger and carries the same
+// ledger_number. A to_id tie-break could not change the result, and would cost an incremental
+// sort on top of the index's pathkeys.
 func (m *IngestStoreModel) GetOldestLedger(ctx context.Context) (uint32, error) {
 	start := time.Now()
 	oldest, err := db.QueryOne[uint32](ctx, m.DB,
-		`SELECT ledger_number FROM transactions ORDER BY ledger_created_at ASC, to_id ASC LIMIT 1`)
+		`SELECT ledger_number FROM transactions ORDER BY ledger_created_at ASC LIMIT 1`)
```
leighmcculloch: The doc comment above (line 284) claims the hypertable's partition-column index gives "every uncompressed chunk an ordered path" for this query, but the oldest chunks — exactly what `GetOldestLedger`/`reconcile_oldest_cursor` need to read — are the ones this same PR's compression-scheduling fix will have already compressed by the time this query runs in production (compress_after is typically well under the retention window on an 87GB table like the one described in this PR).

If TimescaleDB doesn't extend the ordered ChunkAppend path to compressed/columnstore chunks the same way it does for uncompressed ones, `ORDER BY ledger_created_at ASC LIMIT 1` can't get an ordered path into the chunks holding the oldest row, and the hourly reconcile job's cost may not actually improve for real data.

The new test `reconciliation_job_advances_cursor_to_oldest_ledger` (internal/ingest/timescaledb_test.go) only inserts fresh, uncompressed rows and never compresses a chunk before invoking the job, so it would pass either way and doesn't validate the claim against the compressed case this PR's own numbers describe. Worth adding a test that compresses the oldest chunk first (or confirming via `EXPLAIN` that the plan stays an ordered ChunkAppend post-compression) before relying on this for the production dataset.

---

internal/ingest/timescaledb.go
```diff
@@ -63,71 +64,93 @@ func configureHypertableSettings(ctx context.Context, pool *pgxpool.Pool, chunkI
...
 	}
 
+	// The first run lands on the job's slot within the NEXT grid interval,
+	// which is always in the future; fixed_schedule keeps every later run on
+	// initial_start + n*interval, i.e. on the slot.
+	if _, err := pool.Exec(ctx,
+		`SELECT alter_job($1,
+			schedule_interval => $2::interval,
+			fixed_schedule    => true,
+			initial_start     => date_bin($2::interval, now(), 'epoch'::timestamptz)
```
leighmcculloch: `date_bin()` (here and in the `onSlot` check at line 126) rejects interval strings containing months/years ("interval time zones cannot contain months or years"). The comment above already flags this ("date_bin limits the schedule interval to month-free values (minutes, hours, days)"), but nothing enforces or surfaces it — `configureHypertableSettings` just returns whatever Postgres error comes back.

Previously, `compressionScheduleInterval` went straight into `alter_job($1, schedule_interval => $2::interval)` with no such restriction, so a value like `"1 month"` (a plausible low-churn `--compression-schedule-interval` setting) worked fine. After this change, that same operator-supplied value now hard-fails `configureHypertableSettings` on every process start — a silent regression for anyone with month-granularity config already in place.

Since this is a known, deliberate limitation of the staggering approach, it'd be worth either validating the interval up front with a clear error message, or documenting the month/year restriction on the flag itself, rather than surfacing a raw Postgres error at startup.

---

internal/ingest/timescaledb.go
```diff
@@ -53,8 +54,8 @@ func configureHypertableSettings(ctx context.Context, pool *pgxpool.Pool, chunkI
 		return fmt.Errorf("configuring reconciliation job: %w", err)
 	}
 
-	if compressionScheduleInterval != "" {
-		for _, table := range hypertables {
+	if compressionScheduleInterval != "" || compressAfter != "" || maxChunksToCompress > 0 {
+		for i, table := range hypertables {
```
leighmcculloch: Each table's stagger slot (`i, len(hypertables)` passed into `staggerCompressionJob`) is derived purely from its position in the `hypertables` slice and the slice's current length, not from any stable per-table identifier.

If a future change adds, removes, or reorders an entry in `hypertables` (e.g. a 6th hypertable added later), `total` changes for every table on the next restart, so every existing table's `slotOffset` (`index/total`) shifts simultaneously. `onSlot` then evaluates false for all of them, and this restart reschedules every compression job's `initial_start` at once — recreating, for that one transition, exactly the concurrent-compression I/O storm this PR is meant to fix, and potentially delaying compression on some tables by up to a full schedule interval.

Keying the slot off something stable per table (e.g. a hash of the table name mod some fixed bucket count) instead of slice position/length would decouple existing tables' schedules from future additions/removals to the list.

---

Issue #618, opened 2026-05-29
Feature Request: Simulation Query


JakeUrban:
### What problem does your feature solve?

The Wallet Backend currently indexes state changes for transactions included in closed ledgers, but many wallets including Freighter want to know the state changes for transactions prior to signing them.

### What would you like to see?

A GraphQL query that accepts a transaction, simulates it using RPC's simulationTransaction method, and then uses the data provided in the response in addition to the transaction itself to derive and return the state changes that would have been indexed by the wallet backend if the transaction had been submitted to the network and included in a ledger.

### What alternatives are there?

N/A.

PR #691, opened 2026-08-19
feat(graphql): 1/n - simulateStateChanges GraphQL query Phase 0 scaffolding + walking skeleton 

JiahuiWho:
### What

1. **GraphQL schema + codegen** - `simulateStateChanges(transactionXdr: String!): SimulatedStateChanges!` and the new `BaseSimulatedStateChange` interface.

2. **`TransactionSimulationService` skeleton + resolver wiring** — decodes the envelope and branches on transaction type.

3. **Walking skeleton test** — synthesizes a native-SAC transfer and asserts the pipeline emits the sender DEBIT and receiver CREDIT with correct accounts, amounts, and token ID. Also covers the error contract.

### Why

Phase 0 of simulated transaction state changes: the API surface, service skeleton, and a walking skeleton proving the core assumption

Design doc: [Simulated Transaction State Changes](https://docs.google.com/document/d/1lVAni2zAe7hdCL8j7GR5I6lnwGaUy_Pp-UD19l2kVQ4/edit)

### Known limitations

N/A

https://github.com/stellar/wallet-backend/issues/618

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/serve/graphql/resolvers/simulation.go
+++ b/internal/serve/graphql/resolvers/simulation.go
@@ -115,6 +115,11 @@ func (r *Resolver) convertToSimulatedStateChange(sc types.StateChange) (graphql1
 			}, nil
 		default: // invalid reason for BALANCE_AUTHORIZATION; falls through to the error below
 		}
+	case types.StateChangeCategorySigner, types.StateChangeCategorySignatureThreshold,
+		types.StateChangeCategoryDataEntry, types.StateChangeCategoryHomeDomain,
+		types.StateChangeCategoryFlags, types.StateChangeCategoryTrustline:
+		// Classic-source variants: the simulated schema doesn't expose them yet.
+		// They land with the classic derivation source; falls through to the error below.
 	}
 	return nil, fmt.Errorf("state change has no simulated GraphQL type for (category=%s, reason=%s)",
 		sc.StateChangeCategory, sc.StateChangeReason)
--- a/internal/serve/graphql/resolvers/simulation.go
+++ b/internal/serve/graphql/resolvers/simulation.go
@@ -125,16 +125,24 @@ func (r *Resolver) convertToSimulatedStateChange(sc types.StateChange) (graphql1
 		sc.StateChangeCategory, sc.StateChangeReason)
 }
 
-// keyValueUint32 reads a uint32 from a KeyValue JSONB payload, where JSON
-// unmarshal exposes numbers as float64.
+// keyValueUint32 reads a uint32 from a KeyValue payload. The value can arrive in
+// two representations: a real uint32 when the state change is built in memory
+// (the simulation path, e.g. sep41.Processor writes live_until_ledger directly),
+// or a float64 when it has been round-tripped through JSONB from the database.
 func keyValueUint32(kv types.NullableJSONB, key string) (uint32, bool) {
 	raw, ok := kv[key]
 	if !ok {
 		return 0, false
 	}
-	f, ok := raw.(float64)
-	if !ok || f < 0 || f > math.MaxUint32 {
+	switch v := raw.(type) {
+	case uint32:
+		return v, true
+	case float64:
+		if v < 0 || v > math.MaxUint32 {
+			return 0, false
+		}
+		return uint32(v), true
+	default:
 		return 0, false
 	}
-	return uint32(f), true
 }
--- a/internal/serve/graphql/resolvers/simulation.go
+++ b/internal/serve/graphql/resolvers/simulation.go
@@ -1,0 +1,135 @@
+package resolvers
+
+import (
+	"fmt"
+	"math"
+
+	"github.com/stellar/wallet-backend/internal/indexer/types"
+	graphql1 "github.com/stellar/wallet-backend/internal/serve/graphql/generated"
+)
+
+// convertToSimulatedStateChanges resolves the concrete simulated GraphQL type for
+// each state change produced by the simulation pipeline. Unlike the history types
+// (convertStateChangeTypes), simulated types are plain structs built eagerly: the
+// data never touches the DB, so there are no lazy column-backed resolvers.
+func (r *Resolver) convertToSimulatedStateChanges(stateChanges []types.StateChange) ([]graphql1.BaseSimulatedStateChange, error) {
+	converted := make([]graphql1.BaseSimulatedStateChange, len(stateChanges))
+	for i, stateChange := range stateChanges {
+		c, err := r.convertToSimulatedStateChange(stateChange)
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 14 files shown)


internal/services/transaction_simulation.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+}
+
+// stateChangesForTransaction runs a synthesized ledger transaction through the
+// ingestion pipeline into an in-memory buffer — the same processors real
+// ingestion uses, with no persistence — and returns the resulting state changes.
+func (s *transactionSimulationService) stateChangesForTransaction(ctx context.Context, tx ingest.LedgerTransaction) ([]types.StateChange, error) {
+	buffer := indexer.NewIndexerBuffer()
+	if _, err := s.ledgerIndexer.ProcessLedgerTransactions(ctx, []ingest.LedgerTransaction{tx}, buffer); err != nil {
+		return nil, fmt.Errorf("processing transaction through indexer: %w", err)
```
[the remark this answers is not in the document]
JiahuiWho: This is out of scope for Phase 0 (scaffolding only) and Phase 1 (which targets SAC via RPC simulate). Tracking it as a follow-up for SEP-41 simulation coverage rather than expanding this scaffold PR.

internal/serve/graphql/resolvers/simulation.go
```diff
@@ -0,0 +1,148 @@
...
+		converted[i] = c
+	}
+	return converted, nil
+}
+
+// convertToSimulatedStateChange dispatches on (category, reason), mirroring
+// convertStateChangeTypes. A change matching no arm means the simulation pipeline
+// emitted a variant the simulated schema does not expose yet, and surfaces as an
+// error rather than a silently dropped row.
+func (r *Resolver) convertToSimulatedStateChange(sc types.StateChange) (graphql1.BaseSimulatedStateChange, error) {
```
aditya1702, line 31: **Nothing tests this, and #719 makes it live.**

`git grep convertToSimulatedStateChange -- '*_test.go'` → no matches, on either branch. #719 adds the success path (`transaction_simulation.go:102`) but its 195 new test lines are all `package services`. The walking skeleton can't cover it either — that's `package services`, this is `package resolvers`.

A table test needs no setup, since `resolveRequired*` are pure (`resolver.go:78-111`):

```go
got, err := (&Resolver{}).convertToSimulatedStateChanges(fixtures)
```

Fine by me if it lands in #719 instead of here.

internal/serve/graphql/resolvers/simulation.go
```diff
@@ -0,0 +1,148 @@
...
+			}
+			return graphql1.SimulatedAccountCreatedChange{
+				Category:       sc.StateChangeCategory,
+				Reason:         sc.StateChangeReason,
+				AccountAddress: accountAddress,
+				CreatorAddress: creatorAddress,
+			}, nil
+		default: // invalid reason for ACCOUNT; falls through to the error below
+		}
+	case types.StateChangeCategoryAllowance:
```
aditya1702, line 71: **Unreachable — needs the sep41 wiring #719 says isn't there.**

#719, `transaction_simulation.go:396-400`: *"only runs the core Indexer processors, not the separate protocol processors… that processor is not wired in here yet."*

Simulation runs `NewIndexer`'s 3 processors (`indexer.go:131-135`); the only `StateChangeCategoryAllowance` emitter is `sep41/processor.go:257`, which registers separately via `ingest.go:28`.

`SimulatedBalanceChange.toMuxedId` is the same — only set at `sep41/processor.go:180,201`, and `token_transfer.go` has zero muxed handling.

The other two arms are reachable, leave them: `ACCOUNT` ← `contract_deploy.go:54`, `BALANCE_AUTHORIZATION` ← `contracts/sac.go:191`.

Either drop this arm and `toMuxedId` until sep41 lands (that also deletes `keyValueUint32`), or put the limitation in the schema docs so clients don't build against an always-null field.

If it stays: `keyValueUint32`'s `default:` at L144 returns `(0, false)`, which L88 reports as "missing" — a present but wrong-typed value looks absent. `statechange.resolvers.go:165` tells them apart.

internal/serve/graphql/resolvers/statechange.resolvers.go
```diff
@@ -865,24 +865,22 @@ func (r *Resolver) TrustlineUpdatedChange() graphql1.TrustlineUpdatedChangeResol
...
-	homeDomainUpdatedChangeResolver    struct{ *Resolver }
-	signerAddedChangeResolver          struct{ *Resolver }
-	signerRemovedChangeResolver        struct{ *Resolver }
-	signerUpdatedChangeResolver        struct{ *Resolver }
-	thresholdChangeResolver            struct{ *Resolver }
-	trustlineAddedChangeResolver       struct{ *Resolver }
-	trustlineRemovedChangeResolver     struct{ *Resolver }
-	trustlineUpdatedChangeResolver     struct{ *Resolver }
-)
+type accountCreatedChangeResolver struct{ *Resolver }
```
aditya1702, line 868: **No action — flagging why 40 unrelated lines are in the diff.**

`gqlgen`'s `resolver.gotpl` emits one bare `type X struct{ *Resolver }` per object and can't emit a group, so `make gql-generate` ungroups. `gofumpt@v0.9.2` (via `make tidy`) wants 2+ adjacent type decls grouped, so it re-groups. This is the only resolver file with more than one, so it's the only one that flips.

`make tidy` gives you what `make check` produces, if you want the smaller diff — it'll flip back next regen.

internal/serve/graphql/resolvers/simulation.go
JiahuiWho: Added  in #719 as 2b9eb085. The same commit also adds the `SIMULATION_FAILED` rows that were missing from the resolver error-mapping table and the presenter pass-through table.

internal/serve/graphql/resolvers/statechange.resolvers.go
JiahuiWho: Good catch on the cause. I'll run `make tidy` after `make gql-generate` going forward so the diff stays in the gofumpt shape.

internal/serve/graphql/resolvers/simulation.go
JiahuiWho: it's unreachable in #719. The follow-up PR (branch feat/618-simulate-sep41-tokens, opening shortly) wires the SEP-41 processor into simulation, which makes both the `ALLOWANCE` arm and `toMuxedId` real. So I'd rather keep the arm than delete it and add it back one PR later.

The `keyValueUint32` point is fixed in 0116c926 on #719 with a test for the wrong-type case.

aditya1702 (approved):

PR #692, opened 2026-08-20
feat(migrate): --rebuild — wipe and re-derive a protocol's indexed state

aditya1702:
## TL;DR

When a bug fix changes what a protocol derives, the rows already in the database stay wrong — and nothing on-chain can fix them, because things like SEP-41 balances and Blend cost basis are running totals. This PR adds a `--rebuild` flag to both migrate commands that throws the protocol's rows away and replays them through the fixed code.

```
protocol-migrate current-state --protocol-id SEP41 --start-ledger <first ledger> --rebuild
protocol-migrate history       --protocol-id SEP41 --rebuild
```

One command, no separate wipe step. Live ingestion keeps running.

## Why wipe and replay, instead of something surgical

Replaying every event from the start is the only fix that works for *every* protocol. The alternatives each fix some cases and not others:

| Alternative | Why not |
|---|---|
| Ask the chain for the right value | Only some columns have an on-chain answer. A tool that fixes some columns is worse than one that fixes all of them. |
| Recompute from `state_changes` | Retention drops old rows, and a decode bug corrupts those rows too. |
| Auto-repair on restart | A rebuild can run for a day. That is a human decision, not something a pod restart should trigger. |

## How it works

Both rebuilds are the same four steps: **validate → lock → wipe → run the normal migration.** The migrate engine itself is untouched by this PR.

**Current-state rebuild**
1. Refuse if the migration is marked `in_progress`.
2. One transaction: reset the cursor, reset the status, truncate the tables.
3. Run the migration from `--start-ledger`. It hands off to live ingestion at the tip, like any migration.

**History rebuild**
1. Refuse if the migration is marked `in_progress`.
2. Reset the cursor and status **first**. That stops live from writing this protocol's history, so the deletes that follow race nothing.
3. Delete the protocol's rows in 10k-ledger slices.
4. Run the migration over the retained window. Live resumes at handoff.

### Why the two order the wipe differently

They're solving different problems, so they can't share an order.

Current-state has to reset and wipe **together**, or live could fold onto a half-empty table. Everything happens in one transaction.

History can't do that: `state_changes` is compressed, so deleting decompresses whole batches, and a full-window delete inside one transaction would decompress unbounded data. So it resets first — which switches live off for this protocol — then deletes in slices, one transaction each.

### Why the wipe truncates

Live ingestion writes **every protocol in a single transaction per ledger**, and that transaction CASes each protocol's cursor row. The current-state wipe holds that same cursor row.

So the wipe's duration is a stall for *all* ingestion, not just the protocol being rebuilt. `TRUNCATE` costs the same whether it discards ten rows or ten million, which is what keeps that stall bounded — a `DELETE` would not. Safe here: nothing foreign-keys these tables, and none are hypertables.

The transaction also sets `lock_timeout = '5s'`. `TRUNCATE` needs `ACCESS EXCLUSIVE`, so a long API read on one of these tables makes it wait — and Postgres queues later lock requests behind the waiting `TRUNCATE`, so readers pile up too. A failed run the operator can retry beats an open-ended stall.

### Why the history delete reads its upper bound late

The delete bound is read **after** the cursor reset commits, not before.

A live writer that already passed its CAS holds the cursor row, so the reset waits for it — and live bumps `latest_ingest_ledger` in the same transaction it writes history. Writers *after* the reset fail their CAS and write nothing. So a bound read at that moment covers every row that can exist.

Reading it earlier would leave live-written rows above the bound undeleted, for the re-migration to then re-derive at the same `state_change_id`.

## Safety

- `contract_tokens`, `protocol_wasms` and `protocol_contracts` are **never** touched. Nothing rebuilds classification, and the re-migration needs it to know which contracts to fold.
- The cursor row is updated, never deleted — live treats a missing cursor row as a fatal incident.
- The compressed-table delete prunes batches using existing metadata. No new index.
- Reruns are safe: the wipes are idempotent and re-derived rows land at deterministic `state_change_id`s.
- An unusable history window (no cursors yet, or latest below oldest) is refused **before any cursor is reset**, so a rejected run leaves the protocol exactly as it found it.

## Locks

One advisory lock per protocol per strategy. Only one current-state run and one history run can touch a protocol at a time. History migrations gain mutual exclusion they didn't have before.

The lock IDs are a wire-level key — two binaries in a rolling deployment must derive the same one — so the tests pin them to golden values rather than checking the function against itself.

## Trade-offs (accepted on purpose)

1. Nothing verifies our numbers against the chain. A rebuild re-runs our own code.
2. Rebuild time grows with the chain (~hundreds of ledgers/s from the lake).
3. The protocol serves partial data during its rebuild. History writes pause until handoff.
4. A history fix always re-derives the full retained window. Range-scoped rebuilds were dropped for simplicity; they come back if history migration becomes a backfill.

## Testing

**Unit** — wipe/reset semantics, rollback on a failed wipe, the `in_progress` refusal, delete slice edges, the inverted-window refusal, and lock behaviour including a failed multi-protocol acquire freeing its locks.

**Data layer** — both wipes run against real tables. The Blend test asserts coverage by filenode: `TRUNCATE` rewrites a table's file, so a changed filenode proves the wipe reached that table and an unchanged one proves it didn't. That catches a current-state table added later and forgotten — which hand-seeded rows would not.

**Integration** — the Blend suite corrupts one row's amount, deletes another, runs `history --rebuild`, and asserts both come back correct while other namespaces, `contract_tokens` and statuses stay untouched.

`make check` green.

## Docs

`docs/data-migrations/running-a-data-migration.md` gains a **Rebuilding a Protocol** section, and loses the `--oldest-ledger-cursor-name` flag table — that flag is removed here, so the documented command would have failed with "unknown flag".

## After merge

Run both rebuilds for SEP41 in dev and confirm the known bad balance (CAZXR…EXSND) is fixed.


aristidesstaffieri:
We should update the migration documentation to include the parts needed for the rebuild, like the wipe logic.

internal/data/blend/wipe.go
```diff
@@ -0,0 +1,33 @@
...
+		"blend_backstop_positions",
+		"blend_backstop_pools",
+		"blend_reserve_emissions",
+		"blend_emissions",
+		"blend_pool_claimed",
+		"blend_backstop_claimed",
+		"blend_oracle_prices",
+		"blend_auctions",
+	} {
+		if _, err := dbTx.Exec(ctx, "DELETE FROM "+table); err != nil {
```
aristidesstaffieri: any reason not to just `truncate` here?

Since the `update` at the beginning of the transaction in `protocol_migrate_current_state_rebuild.go` takes a lock, it blocks ingestion for all protocols until the delete completes which could be expensive for some protocols and is unbounded, blocking indefinitely.

internal/services/protocol_migrate_history_rebuild.go
```diff
@@ -0,0 +1,225 @@
...
+		if p.HistoryMigrationStatus == data.StatusInProgress {
+			return fmt.Errorf("protocol %q history migration is marked in_progress; investigate the dead run before rebuilding", pid)
+		}
+	}
+	return nil
+}
+
+// retainedWindow returns the wipe bounds: oldest retained ledger through
+// live ingestion's committed tip.
+func (s *protocolHistoryRebuildService) retainedWindow(ctx context.Context) (uint32, uint32, error) {
```
aristidesstaffieri: can this also check `start > latest break` or `oldest <= latest` to avoid operator mistakes that throw this into a loop?

internal/services/protocol_migrate_lock_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/stellar/wallet-backend/internal/db"
+)
+
+func TestMigrateAdvisoryLockID(t *testing.T) {
+	assert.Equal(t, migrateAdvisoryLockID(lockScopeCurrentState, "SEP41"), migrateAdvisoryLockID(lockScopeCurrentState, "SEP41"),
+		"the lock ID must be deterministic — every writer must derive the same key for a protocol")
```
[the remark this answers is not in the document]
aditya1702: Fixed in `ab19252f`. Both IDs are now pinned to hardcoded values, so changing the prefix, hash, or cast fails the test.

internal/data/blend/wipe.go
```diff
@@ -0,0 +1,33 @@
...
+// WipeCurrentState deletes every Blend current-state row, in the same
+// transaction as the caller's cursor reset. Classification tables
+// (protocol_wasms, protocol_contracts) are deliberately untouched: nothing
+// rebuilds them.
+func WipeCurrentState(ctx context.Context, dbTx pgx.Tx) error {
+	for _, table := range []string{
+		"blend_pools",
+		"blend_positions",
+		"blend_reserves",
+		"blend_backstop_positions",
```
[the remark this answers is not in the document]
aditya1702: Fixed in `18aaf5a5` — `internal/data/blend/wipe_test.go` runs the wipe against real tables and checks `protocol_wasms`/`protocol_contracts` survive.

Coverage is checked by filenode instead of seeding all 11 tables. TRUNCATE rewrites a table's file, so a changed filenode proves the wipe hit it. This catches the case you named — a table added later and forgotten — which seeded rows would not, since the seed list would be stale too.

Confirmed it fails: removing `blend_auctions` from the statement makes the test fail naming that table.

internal/services/protocol_migrate_history_rebuild.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+		if statusErr := s.engine.strategy.UpdateMigrationStatus(ctx, dbTx, []string{protocolID}, data.StatusNotStarted); statusErr != nil {
+			return fmt.Errorf("resetting migration status: %w", statusErr)
+		}
+		return nil
+	}); txErr != nil {
+		return fmt.Errorf("resetting history migration state for %s: %w", protocolID, txErr)
+	}
+
+	base := s.engine.processors[protocolID].StateChangeOrdinalBase()
```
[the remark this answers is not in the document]
aditya1702: Real bug, fixed in `78d3fe31`. `latest_ingest_ledger` is now read per protocol, after that protocol's cursor reset commits.

Why that is safe: a live writer that already passed its CAS holds the cursor row, so the reset waits for it — and live writes history and bumps `latest_ingest_ledger` in the same transaction. Writers after the reset fail their CAS and write nothing.

internal/data/sep41/wipe.go
```diff
@@ -0,0 +1,20 @@
...
+
+	"github.com/jackc/pgx/v5"
+)
+
+// WipeCurrentState deletes every SEP-41 current-state row, in the same
+// transaction as the caller's cursor reset. contract_tokens is deliberately
+// untouched: classification owns it and nothing rebuilds it.
+func WipeCurrentState(ctx context.Context, dbTx pgx.Tx) error {
+	for _, table := range []string{"sep41_balances", "sep41_allowances"} {
+		if _, err := dbTx.Exec(ctx, "DELETE FROM "+table); err != nil {
```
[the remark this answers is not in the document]
aditya1702: Fixed in `18aaf5a5` — `internal/data/sep41/wipe_test.go` seeds balances, allowances and a `contract_tokens` row, runs the wipe, and checks the token row survives.

cmd/protocol_migrate.go
```diff
@@ -278,17 +278,43 @@ func runMigration(
...
 		"history",
 		"Backfill protocol history state from oldest to latest ingested ledger",
-		"Processes historical ledgers from oldest_ingest_ledger to the tip, producing protocol state changes and converging with live ingestion via CAS-gated cursors.",
+		"Processes historical ledgers from oldest_ingest_ledger to the tip, producing protocol state changes and converging with live ingestion via CAS-gated cursors. "+
+			"Takes the protocol's advisory lock, so only one history migration or rebuild can run per protocol at a time. "+
+			"--rebuild re-runs an already-migrated protocol from scratch and is destructive: it deletes the protocol's history rows before folding.",
 		func(cmd *cobra.Command, opts *migrationCommandOpts) {
-			cmd.Flags().StringVar(&oldestLedgerCursorName, "oldest-ledger-cursor-name", data.OldestLedgerCursorName, "Name of the oldest ledger cursor in the ingest store. Must match the value used by the ingest service.")
+			cmd.Flags().BoolVar(&rebuild, "rebuild", false, "Delete the protocol's history rows and rebuild them from the oldest retained ledger. Destructive: the protocol's history serves empty/partial data until the rebuild reaches the live frontier.")
```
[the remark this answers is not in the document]
aditya1702: Fixed in `61fba538`. The flag table is gone and the guide has a new "Rebuilding a Protocol" section.

internal/data/blend/wipe.go
aditya1702: No reason not to — switched to TRUNCATE in `b0ca6d1e`.

You are right about the blocking. Live writes every protocol in one transaction per ledger, and that transaction CASes each protocol's cursor row — the same row this wipe holds. So the wipe stalls all ingestion, not just the protocol being rebuilt.

TRUNCATE costs the same for 10 rows or 10 million, which bounds the stall. DELETE did not.

Checked before switching: nothing foreign-keys these tables (sep41 points at `contract_tokens`, not the reverse), and none are hypertables.

internal/services/protocol_migrate_history_rebuild.go
aditya1702: Added in `78d3fe31`. `wipe` now refuses when `latest < oldest`.

You were right that it loops: `latest - start` underflows on uint32, so `end` never equals `latest` and it runs until the process is killed. Test case added.

aditya1702:
Docs done in `61fba538` — `running-a-data-migration.md` has a new "Rebuilding a Protocol" section: the two commands, what each wipe clears, which tables are never touched and why, the locking, and why the two rebuilds order the wipe differently against the cursor reset.

Also dropped the `--oldest-ledger-cursor-name` table from Step 4a — that flag is removed in this PR, so the documented command would have failed.

internal/services/protocol_migrate_current_state_rebuild.go
```diff
@@ -0,0 +1,127 @@
...
+// (ErrCASCursorMissing).
+//
+// Live ingestion writes every protocol in one transaction per ledger, so
+// holding that row lock stalls ingestion for all protocols, not just this one.
+// The stall is bounded because WipeCurrentState truncates: its cost does not
+// grow with the number of rows being discarded.
+func (s *protocolCurrentStateRebuildService) wipe(ctx context.Context, protocolID string) error {
+	processor := s.engine.processors[protocolID]
+	cursorName := s.engine.strategy.CursorName(protocolID)
+	if txErr := db.RunInTransaction(ctx, s.engine.db, func(dbTx pgx.Tx) error {
```
aristidesstaffieri, line 125: one more thing here, should we also set a lock timeout like maybe `SET LOCAL lock_timeout = '5s'`?
truncate needs an exclusive lock and an slow API scan can still block ingestion here.

internal/services/protocol_migrate_history_rebuild.go
```diff
@@ -0,0 +1,236 @@
...
+	// writers after the reset fail their CAS and write nothing. Reading this
+	// before the reset would instead leave rows above the bound undeleted, for
+	// the re-migration to re-derive at the same state_change_ids.
+	latest, latestErr := s.engine.ingestStore.Get(ctx, data.LatestLedgerCursorName)
+	if latestErr != nil {
+		return fmt.Errorf("reading latest ingest ledger: %w", latestErr)
+	}
+	// Guard the slice loop below: uint32 arithmetic on an inverted window never
+	// reaches the upper bound, so it would run until the process is killed.
+	if latest < oldest {
```
aristidesstaffieri: since the reset transaction happens before this, this error path still leaves the cursor at `oldest−1, status not_started`. Should we also check `latest` in `oldestRetained`?

docs/data-migrations/running-a-data-migration.md
```diff
@@ -103,6 +97,42 @@ What it does:
...
+go run main.go protocol-migrate history       --protocol-id <PROTOCOL_ID> --rebuild
+```
+
+Both commands are **destructive** and run the same sequence: validate, take the lock, wipe, then run the normal migration. Live ingestion keeps running throughout; the protocol serves empty or partial data until the rebuild reaches the live frontier and hands off.
+
+### What each wipe clears
+
+| Command | Wipes | Cursor reset to |
+|---------|-------|-----------------|
+| `current-state --rebuild` | every `sep41_*` / `blend_*` current-state table | `--start-ledger` − 1 |
```
aristidesstaffieri: this is pretty nit but `every `sep41_*` / `blend_*` current-state table` could be confused to mean that this specifically wipes those tables when really it is per protocol.

internal/services/protocol_migrate_current_state_rebuild.go
aditya1702: Added in `6ab867ac` — `SET LOCAL lock_timeout = '5s'` at the top of the wipe transaction.

You are right, and it is worse than just waiting: Postgres queues later lock requests behind the waiting TRUNCATE, so new readers pile up too — all while we hold the ingest cursor row.

Verified it fires: with a reader holding the table for 20s, the TRUNCATE aborts after ~5s with `canceling statement due to lock timeout` instead of waiting it out.

internal/services/protocol_migrate_history_rebuild.go
aditya1702: Good catch — fixed in `ccc2803c`. The check moved up into `oldestRetained`, so it fails before any protocol is touched.

You are right about the state it left: cursor at the retention floor and status `not_started` stops live writing that protocol's history until a migration runs, which is worse than where the operator started.

Each wipe still reads its own upper bound after its reset — that read is the authoritative one. The up-front check is only there to fail before mutating.

Also made the slice loop bound itself (`start <= latest`) so it cannot spin no matter what it is handed, rather than depending on a check further up.

docs/data-migrations/running-a-data-migration.md
aditya1702: Fair, that was misleading. Reworded in `a3eab1b6` — the table now says "for that protocol only", with a line above it spelling out that rebuilding SEP41 leaves every `blend_*` table alone.

aristidesstaffieri (approved):

PR #353, opened 2025-10-21
[FEATURE] [internal ]adds metrics for rpc queries

aristidesstaffieri:
Closes #338 

### What
Adds metric collection for rpc interactions

### Why
To support initial deployment and monitoring

### Known limitations
N/a

### Issue that this PR addresses
https://github.com/stellar/wallet-backend/issues/338

#### PR Structure

- [x] It is not possible to break this PR down into smaller PRs.
- [x] This PR does not mix refactoring changes with feature changes.
- [x] This PR's title starts with name of package that is most changed in the PR, or `all` if the changes are broad or impact many packages.

#### Thoroughness

- [x] This PR adds tests for the new functionality or fixes.
- [x] All updated queries have been tested (refer to [this](https://stackoverflow.com/a/29163465) check if the data set returned by the updated query is expected to be same as the original one).

#### Release

- [x] This is not a breaking change.
- [x] This is ready to be tested in development.
- [x] The new functionality is gated with a feature flag if this is not ready for production.

PR #694, opened 2026-08-21
feat(checkpoint): read hot-archive entries so archived wasm hashes are stored

aditya1702:
**TL;DR: checkpoint load now also reads the protocol-23 hot archive. Evicted wasm hashes and contract instances land in `protocol_wasms` / `protocol_contracts` / `contract_tokens` instead of vanishing.**

## The bug

1. Checkpoint load read only the live bucket list.
2. Contract code evicted to the hot archive → hash missing from `protocol_wasms`.
3. Missing hash → every instance pointing at it dropped from `protocol_contracts`.
4. Dropped instance → SEP-41 contract never classified → wrong balances after restore.

## The fix

One new pass in the same load transaction, right before `finalize`. It streams `ingest.NewHotArchiveIterator`.

| Archived entry | Result |
|---|---|
| Contract code | hash → `protocol_wasms` |
| Wasm contract instance | `contract_tokens` (UNKNOWN) + `protocol_contracts` mapping |
| Balance entry | skipped — restore/live ingest folds it, no double count |
| SAC instance | skipped — no wasm to map; row recreated on restore |

Entries restored before the checkpoint never appear: the SDK's hot-archive tombstones suppress them.

## Why the code looks the way it does

- **No `WithFilter`.** A key filter breaks tombstone suppression and resurrects restored entries.
- **The iterator loop's only early exit is a yielded error.** Any other break deadlocks on the SDK iterator's internal WaitGroup.
- **Pre-protocol-23 checkpoints** (HAS version < 2) skip the pass instead of crashing on the missing bucket list.

## Classification: zero changes needed

stellar-rpc (protocol 23+) `getLedgerEntries` returns archived entries **with bytecode** (verified in v28 source, `internal/ledgerentries/main.go`). The existing RPC classify pass resolves archived hashes as-is.

## Testing

- 8 new tests. The key one: live instance + archived code → `protocol_contracts` mapping present (silently dropped before this PR).
- Mutation-checked: emptying the hot-archive iterator fails the tests on the exact expected inserts.
- Full `internal/services` suite green. `make check` green.

No schema changes. No migrations (wallet-backend is pre-release).

### DIFF
--- a/internal/services/checkpoint.go
+++ b/internal/services/checkpoint.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"iter"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


internal/services/checkpoint.go
```diff
@@ -548,6 +588,73 @@ func (p *checkpointProcessor) processContractCode(_ context.Context, wasmHash xd
...
+			// This error return must stay the loop's ONLY early exit. Any other break
+			// blocks forever: the SDK iterator's producer goroutine only stops via its
+			// context, and the iterator's internal WaitGroup unwinds before any
+			// deferred cancel here could fire.
+			return fmt.Errorf("reading hot archive entries: %w", iterErr)
+		}
+		//exhaustive:ignore
+		switch entry.Data.Type {
+		case xdr.LedgerEntryTypeContractCode:
+			contractCodeEntry := entry.Data.MustContractCode()
```
aristidesstaffieri, line 610: Does RPC serve archived contract code entries?I thought it didn't and we would have to use it from the hot archive.

aditya1702: Yes, it does — checked the v28 source.

RPC doesn't read the hot archive itself. Captive core does. `getLedgerEntries` proxies to core's query server, and core loads the live bucket list first, then checks the hot archive for any Soroban keys it didn't find live.

RPC then drops only `not-found`. `archived` passes straight through with the full entry XDR — `cmd/stellar-rpc/internal/ledgerentries/main.go:70-75`:

```go
if entry.State == coreProto.LedgerEntryStateNotFound {
    continue
}
```

And core returns the complete `LedgerEntry` for archived entries, wasm bytes included — `QueryServer.cpp:453-470`, `entry["entry"] = toOpaqueBase64(le)`.

Direct test for it in core: `QueryServerTests.cpp:279-283` seeds CONTRACT_CODE into the hot archive and asserts the response is byte-identical XDR.

Shipped in stellar-rpc v23.0.0 (fell out of #353, which moved `getLedgerEntries` onto core's HTTP server). We pin 24.0.0 in integration tests, so we're above that.

One caveat that doesn't bite us: there's no `archived` flag in the response, the only marker is `liveUntilLedgerSeq == 0`. Classification just reads `code`, so it doesn't care.

aristidesstaffieri (approved):

PR #695, opened 2026-08-21
fix(ingest): verify SAC before recording contract balances

JiahuiWho:
### What

Record a SAC balance only when its contract is confirmed as a SAC via its instance entry (i.e. it exists in `contract_tokens` with `type='SAC'`).

- **Live** (`token_ingestion.go` / `sac_balances.go`): `BatchUpsert`'s insert arm now joins `contract_tokens` on `type='SAC'`, so the check happens as part of the write and unverified balance-shaped entries are silently dropped. 
- **Checkpoint** (`checkpoint.go`): stop creating a `contract_tokens` SAC row from balance shape. SAC-shaped balance entries seen during the scan are buffered in memory instead of being streamed to the DB; `finalize` keeps only those whose contract has a confirmed SAC instance entry and copies them in.

In both paths every retained balance has a `contract_tokens` parent, so the deferred `fk_contract_token` constraint holds at COMMIT.

### Why

SAC balances were recorded from the shape of a contract-data entry alone, without confirming that the writing contract is actually a Stellar Asset Contract. Any contract can write that shape into its own storage, so balances got attached to arbitrary contracts.

Two consequences:

- **Live ingestion (availability):** the balance row references a `contract_id` with no `contract_tokens` parent, so the deferred `fk_contract_token` constraint fires at `COMMIT` (SQLSTATE 23503). That error is classified as permanent (non-retryable) and shares the transaction with the ingestion cursor update, so the ledger rolls back without advancing the cursor and re-fails identically on restart.
- **Checkpoint bootstrap (integrity):** the balance path additionally created a `contract_tokens` row with `type='SAC'` from the shape, classifying an arbitrary contract as a SAC.

### Known limitations

A balance cannot be authenticated from its own entry, so a balance whose SAC instance entry isn't present(i.e. a dormant SAC whose instance was state-archived at bootstrap time) is now dropped rather than trusted. This is bounded and self-healing: such a SAC is inactive, and the next time it's used its instance must be restored, which live ingestion re-ingests and gates the balances back in.

https://hackerone.com/reports/3923230

#### PR Structure

#### Thoroughness

#### Release

### DIFF
--- a/internal/data/sac_balances.go
+++ b/internal/data/sac_balances.go
@@ -111,7 +111,11 @@ func (m *SACBalanceModel) GetByAccount(ctx context.Context, accountAddress strin
 }
 
 // BatchUpsert performs upserts and deletes for SAC balances using UNNEST for efficiency.
-// For upserts (ADD/UPDATE): inserts or updates balance with authorization flags.
+// For upserts (ADD/UPDATE): inserts or updates balance with authorization flags. The
+// insert arm joins contract_tokens on type='SAC', so a balance is written only when its
+// contract is confirmed as a SAC via its instance entry; balance-shaped entries from
+// arbitrary contracts are silently dropped. This also keeps every sac_balances row
+// backed by a contract_tokens parent, so the deferred fk_contract_token holds at COMMIT.
 // For deletes (REMOVE): removes the balance row.
 func (m *SACBalanceModel) BatchUpsert(ctx context.Context, dbTx pgx.Tx, upserts []SACBalance, deletes []SACBalance) error {
 	if len(upserts) == 0 && len(deletes) == 0 {
@@ -152,13 +156,14 @@ func (m *SACBalanceModel) BatchUpsert(ctx context.Context, dbTx pgx.Tx, upserts 
 			SELECT u.account_id, u.contract_id, u.balance::numeric, u.is_authorized, u.is_clawback_enabled, u.last_modified_ledger
 			FROM UNNEST($1::bytea[], $2::uuid[], $3::text[], $4::boolean[], $5::boolean[], $6::int[])
 				AS u(account_id, contract_id, balance, is_authorized, is_clawback_enabled, last_modified_ledger)
+			JOIN contract_tokens ct ON ct.id = u.contract_id AND ct.type = $7
 			ON CONFLICT (account_id, contract_id) DO UPDATE SET
 				balance = EXCLUDED.balance,
 				is_authorized = EXCLUDED.is_authorized,
 				is_clawback_enabled = EXCLUDED.is_clawback_enabled,
 				last_modified_ledger = EXCLUDED.last_modified_ledger`
 
-		if _, err := dbTx.Exec(ctx, upsertQuery, accountIDs, contractIDs, balances, isAuthorized, isClawbackEnabled, ledgerNumbers); err != nil {
+		if _, err := dbTx.Exec(ctx, upsertQuery, accountIDs, contractIDs, balances, isAuthorized, isClawbackEnabled, ledgerNumbers, string(types.ContractTypeSAC)); err != nil {
 			m.Metrics.QueryDuration.WithLabelValues("BatchUpsert", "sac_balances").Observe(time.Since(start).Seconds())
 			m.Metrics.QueriesTotal.WithLabelValues("BatchUpsert", "sac_balances").Inc()
 			m.Metrics.QueryErrors.WithLabelValues("BatchUpsert", "sac_balances", utils.GetDBErrorType(err)).Inc()
--- a/internal/data/sac_balances.go
+++ b/internal/data/sac_balances.go
@@ -47,13 +47,6 @@ type SACBalanceModelInterface interface {
 
 	// Batch operations (for initial population)
 	BatchCopy(ctx context.Context, dbTx pgx.Tx, balances []SACBalance) error
-
-	// DeleteUnverified removes every sac_balances row whose contract is not present in
-	// contract_tokens as a SAC (type='SAC'), returning the number deleted. Checkpoint
-	// population streams balances in from their shape and calls this in finalize, once
-	// contract_tokens is fully populated, to drop rows for contracts not confirmed as
-	// SACs — before the deferred fk_contract_token is checked at COMMIT.
-	DeleteUnverified(ctx context.Context, dbTx pgx.Tx) (int64, error)
 }
 
 // SACBalanceModel implements SACBalanceModelInterface.
@@ -207,26 +200,6 @@ func (m *SACBalanceModel) BatchUpsert(ctx context.Context, dbTx pgx.Tx, upserts 
 	m.Metrics.QueryDuration.WithLabelValues("BatchUpsert", "sac_balances").Observe(time.Since(start).Seconds())
 	m.Metrics.QueriesTotal.WithLabelValues("BatchUpsert", "sac_balances").Inc()
 	return nil
-}
-
-// DeleteUnverified removes sac_balances rows whose contract is not a SAC in contract_tokens.
-func (m *SACBalanceModel) DeleteUnverified(ctx context.Context, dbTx pgx.Tx) (int64, error) {
-	const query = `
-		DELETE FROM sac_balances sb
-		WHERE NOT EXISTS (
-			SELECT 1 FROM contract_tokens ct
-			WHERE ct.id = sb.contract_id AND ct.type = $1
-		)`
-
-	start := time.Now()
-	tag, err := dbTx.Exec(ctx, query, string(types.ContractTypeSAC))
-	m.Metrics.QueryDuration.WithLabelValues("DeleteUnverified", "sac_balances").Observe(time.Since(start).Seconds())
-	m.Metrics.QueriesTotal.WithLabelValues("DeleteUnverified", "sac_balances").Inc()
-	if err != nil {
-		m.Metrics.QueryErrors.WithLabelValues("DeleteUnverified", "sac_balances", utils.GetDBErrorType(err)).Inc()
-		return 0, fmt.Errorf("deleting unverified SAC balances: %w", err)
-	}
-	return tag.RowsAffected(), nil
 }
 
 // BatchCopy performs bulk insert using COPY protocol for speed during checkpoint population.
--- a/internal/data/sac_balances.go
+++ b/internal/data/sac_balances.go
@@ -47,6 +47,13 @@ type SACBalanceModelInterface interface {
 
 	// Batch operations (for initial population)
 	BatchCopy(ctx context.Context, dbTx pgx.Tx, balances []SACBalance) error
+
+	// DeleteUnverified removes every sac_balances row whose contract is not present in
+	// contract_tokens as a SAC (type='SAC'), returning the number deleted. Checkpoint
+	// population streams balances in from their shape and calls this in finalize, once
+	// contract_tokens is fully populated, to drop rows for contracts not confirmed as
+	// SACs — before the deferred fk_contract_token is checked at COMMIT.
+	DeleteUnverified(ctx context.Context, dbTx pgx.Tx) (int64, error)
 }
 
 // SACBalanceModel implements SACBalanceModelInterface.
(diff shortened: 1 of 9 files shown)


internal/services/checkpoint.go
```diff
@@ -263,6 +267,11 @@ type checkpointProcessor struct {
 	// call; PopulateFromCheckpoint fetches metadata for these IDs in a short
 	// follow-up transaction after the load commits.
 	pendingSACMetadata []string
+	// sawSACBalance records whether any SAC-shaped balance entry was streamed to the
+	// batch during the scan. Such entries are inserted from their shape alone, so if any
+	// were seen, finalize runs a cleanup that deletes the rows whose contract is not a
+	// confirmed SAC before COMMIT (see deleteUnverifiedSACBalances).
+	sawSACBalance bool
```
aditya1702: Drop this flag.

`DeleteUnverified` is one unscoped whole-table statement that runs once per bootstrap. The flag only skips it when the checkpoint contained zero SAC balances — and in that case `sac_balances` is empty, so you're skipping a scan of an empty table.

Call `DeleteUnverified` unconditionally in `finalize`. The `deleted > 0` guard already keeps the log quiet on a no-op.

Separately, line 273 says `see deleteUnverifiedSACBalances` — no such function exists. Should be `SACBalanceModel.DeleteUnverified`.

internal/services/token_ingestion.go
```diff
@@ -155,10 +159,39 @@ func (s *tokenIngestionService) processSACBalanceChanges(ctx context.Context, db
...
+	// contract_tokens parent, so the fk_contract_token constraint holds at COMMIT.
+	contractIDSet := make(map[uuid.UUID]struct{}, len(changesByKey))
+	for _, change := range changesByKey {
+		contractIDSet[wbdata.DeterministicContractID(change.ContractID)] = struct{}{}
+	}
+	contractIDs := make([]uuid.UUID, 0, len(contractIDSet))
+	for id := range contractIDSet {
+		contractIDs = append(contractIDs, id)
+	}
+	verifiedIDs, err := s.contractModel.GetExistingSACByID(ctx, dbTx, contractIDs)
```
aditya1702: This whole block can be one SQL clause.

`BatchUpsert`'s insert arm (`sac_balances.go:155`) is already `INSERT ... SELECT ... FROM UNNEST(...) ON CONFLICT DO UPDATE`. Add:

```sql
JOIN contract_tokens ct ON ct.id = u.contract_id AND ct.type = 'SAC'
```

That removes lines 162-190 and 209-211, `GetExistingSACByID` + its godoc, the mock, and the `ContractModel` field on the config/struct/constructor plus `ingest.go:236`. Roughly 75 Go lines for one SQL line — and the check becomes atomic with the write instead of a separate round trip.

The delete arm needs no gate: a REMOVE for an unverified contract matches zero rows anyway, which is exactly what the Go filter achieves by skipping it.

Trade-off: the invariant moves into SQL and you lose the `skipped` count. Your call.

internal/services/token_ingestion.go
```diff
@@ -173,6 +206,9 @@ func (s *tokenIngestionService) processSACBalanceChanges(ctx context.Context, db
 			upserts = append(upserts, sacBal)
 		}
 	}
+	if skipped > 0 {
+		log.Ctx(ctx).Warnf("skipped %d SAC balance change(s) for contracts not verified as SAC", skipped)
```
aditya1702: Consider dropping this log.

It fires once per ledger for any contract writing Balance-shaped entries into its own storage — which per the HackerOne report anyone can do, for free, at any rate. So the volume here is set by an untrusted party.

These entries simply aren't SAC balances. There's nothing to report.

internal/services/token_ingestion.go
aditya1702: This whole block can be one SQL clause.

`BatchUpsert`'s insert arm (`sac_balances.go:155`) is already `INSERT ... SELECT ... FROM UNNEST(...) ON CONFLICT DO UPDATE`. Add:

```sql
JOIN contract_tokens ct ON ct.id = u.contract_id AND ct.type = 'SAC'
```

That removes lines 162-190 and 209-211, `GetExistingSACByID` + its godoc, the mock, and the `ContractModel` field on the config/struct/constructor plus `ingest.go:236`. Roughly 75 Go lines for one SQL line — and the check becomes atomic with the write instead of a separate round trip.

The delete arm needs no gate: a REMOVE for an unverified contract matches zero rows anyway, which is exactly what the Go filter achieves by skipping it.

Trade-off: the invariant moves into SQL and you lose the `skipped` count. Your call.

**If you'd rather keep the Go-side gate:** drop the `contractIDSet` map on lines 169-176. `= ANY($1::uuid[])` handles duplicate UUIDs natively, so the map plus two loops only buys a slightly shorter array param. Build `contractIDs` directly in one loop.

internal/services/token_ingestion.go
aditya1702: Consider dropping this log.

It fires once per ledger for any contract writing Balance-shaped entries into its own storage — which per the HackerOne report anyone can do, for free, at any rate. So the volume here is set by an untrusted party.

These entries simply aren't SAC balances. There's nothing to report.

internal/services/checkpoint.go
aditya1702: This flag goes away if we revert commit 2 (see my summary comment).

Even keeping the current design it doesn't really buy anything. It only skips the DELETE when the checkpoint had zero SAC balances, and in that case the table is empty anyway.

Also, line 273 mentions `deleteUnverifiedSACBalances`, which doesn't exist. Should be `SACBalanceModel.DeleteUnverified`.

internal/services/checkpoint.go
```diff
@@ -594,6 +600,20 @@ func (p *checkpointProcessor) finalize(ctx context.Context, dbTx pgx.Tx) error {
 		return fmt.Errorf("storing tokens in postgres: %w", err)
 	}
 
+	// SAC balances were streamed in from their shape alone. Now that contract_tokens is
+	// fully populated (above), delete any SAC balance whose contract is not a confirmed
+	// SAC. This runs before COMMIT, so every remaining balance has a verified
+	// contract_tokens parent and the deferred fk_contract_token holds.
+	if p.sawSACBalance {
```
aditya1702: This is the bit I'd drop. Commit 1 filtered in Go before the COPY, which meant no delete method, no whole-table DELETE, and no relying on the FK being DEFERRABLE to hold unparented rows partway through the load.

It got replaced over a memory concern that turns out to be about 8 MB: 57,974 rows on a full mainnet checkpoint at ~135 bytes each, against `GOMEMLIMIT=14GiB`. Numbers are in my summary comment.

If we do keep it: this DELETE is unscoped, and it's only safe because `PopulateFromCheckpoint` runs when the cursor is 0 (`ingest_live.go:302`). Nothing here says that.

internal/services/token_ingestion.go
aditya1702: Separate from the checkpoint question, this could just be a JOIN.

`BatchUpsert`'s insert arm (`sac_balances.go:155`) is already `INSERT ... SELECT ... FROM UNNEST(...)`. Adding

```sql
JOIN contract_tokens ct ON ct.id = u.contract_id AND ct.type = 'SAC'
```

drops lines 162-190 and 209-211, `GetExistingSACByID`, the mock, and the `ContractModel` wiring in the config, struct, constructor and `ingest.go:236`. Roughly 75 lines of Go for one line of SQL, and the check happens as part of the write instead of a separate query.

The delete arm doesn't need it, since a REMOVE for an unverified contract matches nothing anyway.

Downside is the rule now lives in SQL and you lose the `skipped` count, so up to you.

If you'd rather keep it in Go, at least drop the `contractIDSet` map on 169-176. `= ANY` already handles duplicates, so you can build `contractIDs` in one loop.

internal/services/token_ingestion.go
aditya1702: I'd drop this log. Any contract can write Balance-shaped entries into its own storage (that's the whole point of the report), so how often this fires is up to whoever feels like triggering it.

They're just not SAC balances. Nothing worth reporting.

internal/data/sac_balances.go
```diff
@@ -202,6 +209,26 @@ func (m *SACBalanceModel) BatchUpsert(ctx context.Context, dbTx pgx.Tx, upserts
 	return nil
 }
 
+// DeleteUnverified removes sac_balances rows whose contract is not a SAC in contract_tokens.
+func (m *SACBalanceModel) DeleteUnverified(ctx context.Context, dbTx pgx.Tx) (int64, error) {
```
aditya1702: This method goes too, along with the interface entry on 50-56, the mock, and `TestSACBalanceModel_DeleteUnverified`.

Worth saying separately from the revert: it's an unscoped whole-table DELETE sitting on the shared model interface, where anything can call it. It's only safe because `PopulateFromCheckpoint` runs at cursor 0 (`ingest_live.go:302`), and nothing in the signature or the godoc says so. Filtering before the COPY makes that question disappear.

aditya1702:
Fix looks right, gating on the instance entry is the correct call. Two bigger things, plus some inline comments.

### 1. I'd revert commit 2

I checked the numbers on the dev mainnet DB first:

- `sac_balances` is 57,974 rows / 12 MB, across 260 distinct contracts
- `oldest_ingest_ledger` is 63708223, which is the checkpoint ledger, so that's a full mainnet checkpoint bootstrap's worth of balances, i.e. exactly what commit 1 would have buffered

`SACBalance` is about 135 bytes live (`AddressBytea` is a string, so a 56-char strkey, a balance of at most 23 chars, and a 16-byte UUID). That's ~7.8 MB, maybe 20 MB peak with append doubling. Ingest runs `GOMEMLIMIT=14GiB` on a 16Gi pod, so it's around 0.06% of the budget. It'd have to grow 1000x before buffering is a problem.

Worth noting that DB was bootstrapped with the pre-fix code, which recorded balances from shape alone, so 57,974 is an upper bound rather than an undercount.

Reverting gets rid of `DeleteUnverified` and its interface entry, the mock, `TestSACBalanceModel_DeleteUnverified`, the `sawSACBalance` plumbing, an unscoped whole-table DELETE, and the dependency on `fk_contract_token` being DEFERRABLE to hold unparented rows mid-load.

Commit 1 was just "only insert verified balances": filter in Go, then COPY. Simpler, and the numbers say it costs ~8 MB.

### 2. This PR makes the SAC RPC enrichment path dead code

You removed the last thing that could create a SAC row with missing metadata. The only writers of `type='SAC'` left are `sac_instances.go:75` and `checkpoint.go:704`, and both always set Code/Issuer/Name/Symbol. So `checkpoint.go:593`'s `Code == nil` and the enrichment query's `name IS NULL` can never match again, and `BatchInsert` is `ON CONFLICT DO NOTHING` so it can't null them out either.

That leaves all of this unreachable, ~270 lines plus ~310 of tests:

- `enrichSACMetadata`, `enrichSACMetadataWithRetry`, `EnrichStaleSACMetadata` and its startup call at `ingest_live.go:330`
- `GetSACContractsMissingMetadata`
- `FetchSACMetadata` and `fetchSACMetadataForContract`
- `maxSACEnrichmentRetries`, `sacEnrichmentRetries`, `sacEnrichmentBackoff`
- the `contractMetadataService` field on `checkpointService`, whose only reader is `enrichSACMetadata`

One thing to check before deleting: is `FetchSACMetadata` meant to cover the archived-instance case you mention under Known Limitations? If so, keep the fetch and just drop the checkpoint wiring.

Minor: `contract_tokens.go:39-43`, `checkpoint.go:585-591` and `checkpoint.go:264-268` all still describe balance-derived SAC rows, which this PR removes.

internal/services/checkpoint.go
JiahuiWho: Gone with the revert.

internal/services/checkpoint.go
JiahuiWho: Reverted commit 2

internal/services/token_ingestion.go
JiahuiWho: Took the JOIN. GetExistingSACByID, its mock, and the ContractModel wiring are removed.

internal/data/sac_balances.go
JiahuiWho: Removed with the revert, along with the interface entry, mock, and test.

JiahuiWho:
1. Reverted commit 2, also fixed the three stale comments. Thanks for actually measuring it.
2. Confirmed the enrichment path is dead. On your open question: `FetchSACMetadata` doesn't help the archived-instance case either, since an archived instance means no `contract_tokens` row exists at all, so there's nothing to enrich. Delete in https://github.com/stellar/wallet-backend/pull/723 to keep this diff reviewable.

internal/data/sac_balances.go
```diff
@@ -152,13 +156,14 @@ func (m *SACBalanceModel) BatchUpsert(ctx context.Context, dbTx pgx.Tx, upserts
 			SELECT u.account_id, u.contract_id, u.balance::numeric, u.is_authorized, u.is_clawback_enabled, u.last_modified_ledger
 			FROM UNNEST($1::bytea[], $2::uuid[], $3::text[], $4::boolean[], $5::boolean[], $6::int[])
 				AS u(account_id, contract_id, balance, is_authorized, is_clawback_enabled, last_modified_ledger)
+			JOIN contract_tokens ct ON ct.id = u.contract_id AND ct.type = $7
```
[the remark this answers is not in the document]
JiahuiWho: There are no deployed instances upgrading in place

aditya1702 (approved):

PR #697, opened 2026-08-22
perf(data): COPY account-link rows in primary-key order

aditya1702:
(Recreates #696, closed by a head-branch rename.)

### What
Sort the transactions_accounts / operations_accounts COPY batches by (account_id, id) before shipping them.

### Why
The rows were built from nested map iteration, so each COPY paid a randomized btree descent per row on the account_id-led primary key. The balance upserts already sort by PK for the same reason.

### Testing
- `go test ./internal/data/` green; COPY order is correctness-neutral.
- A/B on the phase-3 loadtest rig: neutral there (the rig's ~26k-account population keeps the index cache-resident); the change is motivated by mainnet-scale indexes where random probes miss cache. Branch also hosts the upcoming persist-phase optimization batch (see handoff spec).

### DIFF
--- a/internal/data/accounts.go
+++ b/internal/data/accounts.go
@@ -5,8 +5,11 @@
 package data
 
 import (
+	"bytes"
+	"cmp"
 	"context"
 	"fmt"
+	"slices"
 	"time"
 
 	"github.com/jackc/pgx/v5"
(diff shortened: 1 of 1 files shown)


aditya1702:
Duplicate of #696, which survived the branch rename.

aditya1702:
Supersedes #696 (closed permanently by the head-branch rename to live-ingest-persist-opts).

aditya1702:
Holding off — this branch will get one PR together with the persist-phase optimization batch.

PR #698, opened 2026-08-22
fix(graphql): memoize fragment depth to bound depth-limit walk

JiahuiWho:
### What

Cache each fragment's computed depth by name in `selectionSetDepth` so a fragment is expanded at most once per operation, instead of being re-expanded on every path that reaches it.

### Why

`selectionSetDepth` re-expanded a fragment every time it was spread. Since a fragment spread adds no level of its own, that work is redundant, and when fragments fan out into one another (each spreading the next more than once), it compounds into work that grows exponentially with the fragment count, even for a small document with shallow actual depth. Since the reported depth stays low, neither the depth nor complexity limit rejects it.

Memoizing each fragment's depth by name bounds the walk to O(fragments + selections) and returns the identical depth value.

### Known limitations

N/A

https://hackerone.com/reports/3959812

#### PR Structure

#### Thoroughness

#### Release


aristidesstaffieri:
@JiahuiWho this fixes the depth walk but not the growth rate, we should have something that rejects the request when it exceeds that cap on total selections/fragments also.

PR #704, opened 2026-08-26
Fix the state_changes indexes and add a foreign key on pool-share balances

aditya1702:
**Start here:** `internal/db/migrations/2026-08-21.0-statechanges-account-id-index.sql` — the only file that changes an existing database.

Two schema changes. Each ships with the query change that needs it.

### `state_changes` lookups by operation now use the primary key

`BatchGetByOperationID` and `BatchGetByOperationIDs` only filtered on `operation_id` and `ledger_created_at`. That isn't enough to use the primary key `(ledger_created_at, to_id, operation_id, state_change_id)`, so both queries needed a separate index on `operation_id`.

We can work out `to_id` from the operation ID for free, because of how TOIDs are encoded: `tx_to_id = operation_id &^ 0xFFF`. Adding it to the WHERE clause matches the first three columns of the primary key, so both queries use the PK directly.

Two indexes now have nothing using them, and are dropped:
- `idx_state_changes_operation_id` — the primary key covers it.
- `idx_state_changes_account_category` — replaced by the smaller `idx_state_changes_account_id`.

`sql-migrate` tracks migrations by filename, so editing `2025-06-10.4` in place only affects brand-new databases. `2026-08-21.0` brings existing ones in line and does nothing on new ones.

### Pool-share balances get a foreign key to their pool

Ingestion writes the pool row first, in the same transaction, so a share row without a pool is always a bug. The foreign key makes it fail loudly on write instead of quietly producing a broken join on read. Same idea as the existing `trustline_balances → trustline_assets` constraint.

It's deferrable, and added as `NOT VALID` then validated outside a transaction, so ingestion keeps writing while it runs.

---
**PR 1 of 6** replacing #684 · schema → fixes → pipeline → parallel writes → table groups → final fixes


internal/db/migrations/2026-08-21.0-statechanges-account-id-index.sql
```diff
@@ -0,0 +1,26 @@
+-- +migrate Up
```
aristidesstaffieri, line 1: why not `notransaction` here too?

aristidesstaffieri (approved):

PR #705, opened 2026-08-26
Three small fixes: checkpoint pool shares, worker pool sizing, metric buckets

aditya1702:
Three unrelated fixes. Read them in any order.

**1. Checkpoint drops share rows for pools it skipped** — `internal/services/checkpoint.go`

The checkpoint skips pool types `liquidity_pools` can't store, but it was saving their share rows anyway. With the new foreign key from #704, a share row with no matching pool now fails the whole checkpoint load. Share rows are held back until every pool has been read, then filtered against the skipped ones.

**2. Worker pool sizes off GOMAXPROCS instead of NumCPU** — `internal/services/ingest.go`

`runtime.NumCPU()` reports the *node's* cores, not the container's limit. A 6-core pod on a 36-core node was running 72 workers on 6 threads. `GOMAXPROCS(0)` respects the limit.

**3. Metric buckets line up with the ledger close time** — `internal/metrics/ingestion.go`

The histogram jumped 0.5 → 1 → 2 seconds, so a stage taking 0.9s and one taking 1.9s landed in the same p99 bucket. We need to know whether the slowest stage stays under the ledger close time, and the old buckets couldn't tell us. Boundaries are now 0.6 and 1.0, with 0.5 and 0.75 just below.

---
**PR 2 of 6** replacing #684 · schema → **fixes** → pipeline → parallel writes → table groups → final fixes


internal/services/ingest.go
```diff
@@ -162,10 +162,10 @@ type ingestService struct {
...
 	// Create worker pool for the ledger indexer (parallel transaction processing within a
 	// ledger). This is CPU-bound XDR decode/processing work, not RPC-bound, so it's sized off
-	// NumCPU rather than an RPC batch size; 2x gives headroom for goroutines blocked on the
-	// occasional DB lookup without letting the pool grow unbounded (pond.NewPool(0) is
-	// unbounded).
-	ledgerIndexerPool := pond.NewPool(2 * runtime.NumCPU())
+	// GOMAXPROCS — which honors a container CPU limit, where NumCPU reports the node's cores —
+	// with 2x headroom for goroutines blocked on the occasional DB lookup, without letting the
+	// pool grow unbounded (pond.NewPool(0) is unbounded).
+	ledgerIndexerPool := pond.NewPool(2 * runtime.GOMAXPROCS(0))
```
aristidesstaffieri, line 156: should we also update the same thing at `internal/services/ingest.go:175`?

internal/services/ingest.go
JiahuiWho, line 163: The backfill pool still defaults to `runtime.NumCPU()`. Same reasoning applies here, should it also use `GOMAXPROCS(0)`?

internal/services/ingest.go
```diff
@@ -150,10 +150,10 @@ type ingestService struct {
 func NewIngestService(cfg IngestServiceConfig) (*ingestService, error) {
 	// Create worker pool for the ledger indexer (parallel transaction processing within a
 	// ledger). This is CPU-bound XDR decode/processing work, not RPC-bound, so it's sized off
-	// NumCPU rather than an RPC batch size; 2x gives headroom for goroutines blocked on the
-	// occasional DB lookup without letting the pool grow unbounded (pond.NewPool(0) is
-	// unbounded).
-	ledgerIndexerPool := pond.NewPool(2 * runtime.NumCPU())
+	// GOMAXPROCS — which honors a container CPU limit, where NumCPU reports the node's cores —
```
JiahuiWho, line 153: nit: dash

internal/metrics/ingestion.go
```diff
@@ -90,9 +90,14 @@ func newIngestionMetrics(reg prometheus.Registerer) *IngestionMetrics {
 			Buckets: []float64{0.05, 0.1, 0.15, 0.25, 0.5, 0.75, 1, 1.5, 2, 3, 5, 7, 10},
 		}),
 		PhaseDuration: prometheus.NewHistogramVec(prometheus.HistogramOpts{
-			Name:    "wallet_ingestion_phase_duration_seconds",
-			Help:    "Duration of each ingestion phase.",
-			Buckets: []float64{0.01, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 30, 60},
+			Name: "wallet_ingestion_phase_duration_seconds",
+			Help: "Duration of each ingestion phase.",
+			// The 0.6 and 1 boundaries are the grading bars: the pipeline's
```
JiahuiWho: Comment here says 0.6s and 1s bracket the ledger close time, but the comment on line 39 says close cadence is 5 to 6s 🤔

internal/services/ingest.go
aditya1702: Done in 65e56e3f: the backfill pool now defaults to `runtime.GOMAXPROCS(0)` as well, and the `BackfillWorkers` doc comment says so.

internal/metrics/ingestion_test.go
```diff
@@ -99,7 +99,7 @@ func TestIngestionMetrics_PhaseDuration_Buckets(t *testing.T) {
 	for _, f := range families {
 		if f.GetName() == "wallet_ingestion_phase_duration_seconds" {
 			h := f.GetMetric()[0].GetHistogram()
-			assert.Len(t, h.GetBucket(), 11) // 11 custom boundaries
+			assert.Len(t, h.GetBucket(), 13) // 13 custom boundaries
```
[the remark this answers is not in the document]
aditya1702: Done in 096f038a: the test asserts each of the 13 upper bounds against the literal boundary slice, so swapping 0.6 or 0.75 for another value fails it.

internal/services/ingest.go
aditya1702: Done in 65e56e3f: the backfill pool defaults to `GOMAXPROCS(0)` too.

internal/services/ingest.go
aditya1702: Dropped the em-dashes in 65e56e3f.

internal/metrics/ingestion.go
```diff
@@ -90,9 +90,14 @@ func newIngestionMetrics(reg prometheus.Registerer) *IngestionMetrics {
...
+			// contract is that the slowest stage's p99 stays under the ledger
+			// close time, and those two bracket the close times this service
+			// ingests at. 0.5 and 0.75 sit just below them so movement between
+			// runs is visible rather than rounded away.
+			Buckets: []float64{0.01, 0.05, 0.1, 0.25, 0.5, 0.6, 0.75, 1, 1.5, 2, 3, 5, 10},
```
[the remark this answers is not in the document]
aditya1702: Good catch: with 10s as the last finite bound, `histogram_quantile` would report a 60s stall as 10s. Restored the 30s and 60s tail buckets in 97f2c323 (15 boundaries; the test pins all of them).

PR #707, opened 2026-08-26
Stream bulk COPYs on sibling connections and coalesce persist backlogs

aditya1702:
**Start with:** the doc comment on `persistLedgerData` in `internal/services/ingest_live.go`. Its commit-ordering rules are the contract for everything here.

### 1. Bulk COPYs stream on sibling connections

The three COPY families ran sequentially in one transaction on one connection, so the insert phase cost their sum and one Postgres backend did all the index maintenance.

They now stream concurrently on three sibling connections, each in its own transaction, while a coordinating transaction stages everything else. Table sets are disjoint with no FKs between them.

**Commits are the visibility point.** Siblings commit first, the coordinating transaction — whose cursor decides which ledgers exist — strictly last.

- Fail *before* the first commit → everything rolls back, retryable as before.
- Fail *after* → wraps `ErrPartialPersist`, fatal. COPY has no `ON CONFLICT`, so a retry would collide on PKs.

The only crash state this produces is orphaned rows above the cursor, which `DeleteRowsAboveLedger` clears at startup.

### 2. Backlogs coalesce into batched commits

When process outruns persist, up to `--live-persist-max-batch-size` consecutive ledgers fold into one commit set. At normal cadence every batch is size 1 and nothing changes.

### 3 and 4

Siblings commit with `synchronous_commit = off` (durability unchanged — the coordinator's synchronous commit covers their WAL). And batches only cut on *unseen* classification inputs, not any observation — the old rule cut every batch to size 1 and disabled batching entirely.

No performance numbers yet — review on design and correctness.

---
4 of 6 splitting #684. schema → fixes → pipeline → **seams** → write families → hardening

### DIFF
--- a/cmd/ingest.go
+++ b/cmd/ingest.go
@@ -69,6 +69,14 @@ func (c *ingestCmd) Command() *cobra.Command {
 			Required:    false,
 		},
 		{
+			Name:        "live-persist-max-batch-size",
+			Usage:       "Maximum consecutive ledgers coalesced into one persist commit when live ingestion falls behind. 1 persists every ledger in its own commit; larger values amortize inserts across a backlog at the cost of coarser crash recovery and visibility latency under load.",
+			OptType:     types.Int,
+			ConfigKey:   &cfg.LivePersistMaxBatchSize,
+			FlagDefault: 5,
+			Required:    false,
+		},
+		{
 			Name:        "archive-url",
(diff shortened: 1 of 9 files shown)


aditya1702:
Reopened as #711 — the branch was renamed to `persist/4-parallel-writes` and GitHub closes a PR when its source branch is renamed. Same commits, same code.

PR #708, opened 2026-08-26
Give each write family its own persist sibling

aditya1702:
**Start with:** the sibling table in `persistLedgerData`. This PR is that table growing from 3 entries to 7.

**One rule throughout:** each write family rides the transaction that stages its own FK parents.

### Siblings: 3 → 7

| New sibling | Why it can move |
|---|---|
| `transactions_accounts`, `operations_accounts` | Largest index payloads; were serialized behind their parents |
| native balances + liquidity pools | No foreign keys at all |
| trustline assets + balances | Their FK parents move with them |

`SAC balances stay on the coordinator` — their parent `contract_tokens` is also written by the classification path there, and two concurrent same-key inserts could deadlock at the barrier.

**Protocol history also moves onto the `state_changes` sibling.** It was writing `state_changes` rows from the coordinating transaction, so two of our own transactions hit the same hypertable concurrently. At a chunk boundary TimescaleDB serializes chunk creation — and since the coordinator only commits after every sibling returns, a waiting sibling hangs the pipeline forever, invisible to Postgres's deadlock detector.

### Correctness fixes this surfaced

- **Event membership reads the persist transaction, not the pool** — mid-batch ledgers couldn't see `protocol_contracts` rows staged earlier in the same batch, silently dropping events for SEP-41.
- **Buffer rotation needs `2*cap` buffers, not `cap+1`** — process starved on `freeBuffers` and batches settled below the configured cap at every load.
- **Transactions key by ToID, not hash** — one envelope at two tx-set positions made the transaction and participant maps disagree, dropping rows and COPYing zero timestamps.

Plus data-layer hygiene: sorted + `IS DISTINCT FROM`-guarded balance upserts, close-time-bounded startup reconciliation, and batching now defaults to 1 (commit per ledger).

No performance numbers yet — review on design and correctness.

---
5 of 6 splitting #684. schema → fixes → pipeline → seams → **write families** → hardening

### DIFF
--- a/cmd/ingest.go
+++ b/cmd/ingest.go
@@ -70,10 +70,10 @@ func (c *ingestCmd) Command() *cobra.Command {
 		},
 		{
 			Name:        "live-persist-max-batch-size",
-			Usage:       "Maximum consecutive ledgers coalesced into one persist commit when live ingestion falls behind. 1 persists every ledger in its own commit; larger values amortize inserts across a backlog at the cost of coarser crash recovery and visibility latency under load.",
+			Usage:       "Maximum consecutive ledgers coalesced into one persist commit when live ingestion falls behind. The default 1 persists every ledger in its own commit — right for networks whose close time comfortably exceeds the persist time; raise it on high-TPL/short-block deployments to amortize inserts across a backlog at the cost of coarser crash recovery and visibility latency under load.",
 			OptType:     types.Int,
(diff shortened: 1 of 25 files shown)


aditya1702:
Reopened as #712 — the branch was renamed to `persist/5-table-groups` and GitHub closes a PR when its source branch is renamed. Same commits, same code.

PR #715, opened 2026-08-26
build(deps): upgrade to protocol 28 (go-stellar-sdk v0.7.3, core/rpc 28)

aditya1702:
**Merge before 2026-08-27 17:00 UTC.** That's the testnet p28 vote. Our SDK is p27 and *errors* on p28 ledgers, so testnet ingestion stops at the vote.

## The change

| | |
|---|---|
| SDK | `v0.7.3` (p28 XDR) |
| core / rpc / protocol | `:28` / `:28.0.0` / `28` |
| CAP-0085 contracts | logged + counted instead of silently dropped |
| `checkpoint.go` panic | fixed |

Zero source edits needed to compile. p28 is purely additive.

## Warning: CI proves nothing here

`go build` passes untouched, and `exhaustive` is blind (`default-signifies-exhaustive: true`, and the two enums p28 extends were only checked with `if`).

Green CI = "build isn't broken", not "p28 works". Hence hand-written fixtures that round-trip each new arm through binary encoding.

## Bug fixed: `checkpoint.go` could panic ingest

Used `Val.MustInstance()` while the caller constrains only `Key.Type`. Those are independent XDR unions. Now `GetInstance()`.

## Not building CAP-0085 resolution yet

External-ref contracts can't be classified, so a SEP-41 token or Blend pool deployed that way won't show in `contract_tokens` / `blend_pools`.

Zero such contracts exist anywhere. And the owner can re-point the tag at any ledger **emitting no event**, so doing it right needs a migration + flip detection — built against nothing. Made visible instead:

- `wallet_ingestion_external_ref_contracts_total` (live **and** checkpoint paths)
- warn log with owner + tag

## Heads up

v0.7.2 changed trustline-revocation LP-leg comparison from pointer to value. **New revocation state changes may pick a different leg than historical rows.** Upstream fix, not repaired here.

## Review round

| Comment | Outcome |
|---|---|
| Copilot: "still-valid address" wording is wrong | Correct. Verified `DecodeAny` → `invalid payload length 32, expected 40`. My earlier probe ran against the old SDK. Moot now — guard reverted. |
| @aristidesstaffieri: guard causes a crash loop, 3 paths pass non-G addresses | Correct, all 3 verified. Guard reverted in `55c246c8`. Filed as #716. |
| @aristidesstaffieri: wire metrics into `CheckpointServiceConfig` | Done in `b1d2b847`. |

## Testing

- `make check` — pass
- `make unit-test` — pass, 32 packages, no races
- `make integration-test` — pass, 194.9s on core `:28` + rpc `:28.0.0` (gated off it returns in 0.9s, so this was real)

## You need to do

Bump `stellar-rpc` → `28.0.0` in `stellar/kube` for dev/stg/prd. Not touched here.


internal/indexer/types/types.go
```diff
@@ -81,6 +81,14 @@ func (a AddressBytea) Value() (driver.Value, error) {
 	if err != nil {
 		return nil, fmt.Errorf("decoding stellar address %s: %w", a, err)
 	}
+	// The column holds one version byte plus exactly 32 payload bytes. A longer
+	// payload (a muxed account carries 40) would be truncated by the copy below
+	// and would then re-encode as a different, still-valid address, losing the
+	// muxed ID with no error anywhere. Muxed addresses belong in the separate
+	// address_muxed columns; refuse anything that does not fit.
+	if len(rawBytes) != 32 {
```
aristidesstaffieri: it looks like when we hit this case, ingestion  will go into a crash loop and there are 3 cases that don't pass G addresses.
```
- Soroban op participants — internal/indexer/processors/contract_operations.go:156, :256, :300, :344 seed participants with op.SourceAccount().Address(). SourceAccount() returns *xdr.MuxedAccount, so a muxed source renders as M... (40-byte payload). The classic path at transaction_operation_wrapper.go:515-526 uses ToAccountId(). This is the most reachable producer.
- SEP-41 muxed ScAddress — sep41/events.go:302-312 calls ScAddress.String(), which emits M... for ScAddressTypeMuxedAccount; sep41/processor.go:173-243 writes it into AddressBytea and WithAccount.
- Signed-payload signers — effects_horizon.go:434-483 → effects.go:556-591 → state_change_builder.go:96 writes the P... strkey (32+4+N bytes) into SignerAccountID.
```

internal/services/checkpoint.go
```diff
@@ -709,6 +719,17 @@ func (s *checkpointService) processContractInstanceChange(
...
 		}
+
+	case xdr.ContractExecutableTypeContractExecutableStellarAsset:
+		// Handled by the SAC branch above; reaching here means Extract failed.
+
+	case xdr.ContractExecutableTypeContractExecutableExternalRef:
+		// CAP-0085: an (owner, tag) pair naming an entry in another contract's
+		// storage, so there is no WASM hash and no classification. Live
+		// ingestion counts these; the checkpoint pass only logs, because
+		// checkpointService carries no metrics registry.
```
aristidesstaffieri: this can probably wait until we add the CAP85 resolution but the `CheckpointServiceConfig` just omits the metrics registry, it can be added there in order to increment `ExternalRefContractsTotal` here.

internal/indexer/types/types.go
```diff
@@ -81,6 +81,14 @@ func (a AddressBytea) Value() (driver.Value, error) {
 	if err != nil {
 		return nil, fmt.Errorf("decoding stellar address %s: %w", a, err)
 	}
+	// The column holds one version byte plus exactly 32 payload bytes. A longer
+	// payload (a muxed account carries 40) would be truncated by the copy below
+	// and would then re-encode as a different, still-valid address, losing the
+	// muxed ID with no error anywhere. Muxed addresses belong in the separate
+	// address_muxed columns; refuse anything that does not fit.
```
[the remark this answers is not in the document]
aditya1702: You're right, and thanks — I had this backwards.

Verified on v0.7.3:
```
DecodeAny("MA7QY...UAPG") -> invalid payload length 32, expected 40
Decode(M, "MA7QY...UAPG") -> invalid payload length 32, expected 40
```
So the truncated value is undecodable, not "still-valid". My earlier probe accidentally ran against the previous SDK, where that check didn't exist.

The wording is moot here now — @aristidesstaffieri found the guard causes an ingestion crash loop, so it's reverted in 55c246c8. The accurate failure mode you describe is written up in #716.

internal/indexer/types/types.go
aditya1702: Confirmed all three, and you're right that it's a crash loop — reverted the guard in 55c246c8.

Traced each one:
- `contract_operations.go:156,:256,:300,:344` — `op.SourceAccount().Address()`, muxed renders `M...` (40 bytes). `transaction_operation_wrapper.go:515` uses `.ToAccountId()`, so the classic and Soroban paths genuinely disagree.
- `sep41/events.go:302` — `ScAddress.String()` emits `M...`, written into `AddressBytea` by `processor.go`.
- `state_change_builder.go:96` — `P...` into `SignerAccountID`.

All land in `AddressBytea`/`NullAddressBytea`, so a Valuer error becomes an insert error.

Worth flagging what this exposes: those paths are already writing rows that can't be decoded back. `Scan` re-encodes the truncated 32 bytes fine, but any later `DecodeAny` fails with `invalid payload length 32, expected 40`. So it's a live bug, not just a latent one.

Filed as #716 with the three producers and three options (unmux at the producer / store muxed separately / widen the column). Keeping it out of this PR since a version bump shouldn't carry that decision.

internal/services/checkpoint.go
aditya1702: Good catch — did it now rather than waiting, in b1d2b847. It was ~6 lines and the inconsistency bothered me: I'd written in the commit message that checkpoint only logs "because checkpointService carries no metrics registry", which was a self-inflicted limitation.

`CheckpointServiceConfig` now takes `MetricsService`, wired from the registry already built at `internal/ingest/ingest.go:179`. The counter now covers both paths, so contracts predating the ingest window aren't invisible to the metric that exists to make them visible.

Test asserts the delta, and asserts zero on the non-external-ref arms so a stray increment is caught too. Verified it bites by removing the `.Inc()`.

aristidesstaffieri (approved):

Issue #689, opened 2026-08-18
Parallel backfill batches publish an oldest_ingest_ledger that claims un-ingested ledgers


aristidesstaffieri:
### Summary

Backfill advances `oldest_ingest_ledger` once per batch, from inside that batch's own transaction, while batches run in parallel. A batch that succeeds above a batch that failed still lowers the cursor. The published cursor then claims contiguous history across ranges that were never written.

### Cause

1. Each batch's final flush passes its own start ledger as the cursor value — `internal/services/ingest_backfill.go:367`, applied at `:283` (and `:373`/`:384` when the ledger count is an exact multiple of the insert batch size).
2. `UpdateMin` is monotonically decreasing — `internal/data/ingest_store.go:198-203`:

   ```sql
   UPDATE ingest_store SET value = LEAST(value::bigint, $2)::text WHERE key = $1
   ```

3. Batches execute concurrently, so the final cursor is the minimum start ledger among *successful* batches. Failed batches return before their flush and contribute nothing. The result has no relationship to which ranges actually landed.
4. `reconcile_oldest_cursor` cannot repair this. It compares the stored value against `MIN(ledger_number)` and returns early unless `actual_min > stored` (`internal/ingest/timescaledb.go:223`). A gap above the minimum is invisible to it, and it only moves the cursor forward.

### Proof

The repo's real `UpdateMin` statement and the verbatim `reconcile_oldest_cursor` function, run against Postgres 16. A `[1000-1749]` backfill as three parallel batches, middle batch succeeds:

```
batch A [1000-1249] FAILS   batch B [1250-1499] OK   batch C [1500-1749] FAILS

cursor after backfill "succeeds" (exit 0, see #688):  1250
actual data:                    min=1250  max=1999  rows=500
gap inside the claimed range:   1500-1749  (250 ledgers missing)
after reconcile_oldest_cursor:  1250   (unchanged)
```

Lowest batch succeeds instead, middle fails:

```
batch A [1000-1249] OK   batch B [1250-1499] FAILS   batch C [1500-1749] OK

cursor:                         1000
gap inside the claimed range:   1250-1499  (250 ledgers missing)
after reconcile_oldest_cursor:  1000   (unchanged)
```

Both cases also satisfy the integration assertion `oldest <= expectedOldestLedger` (`internal/integrationtests/infrastructure/backfill_helpers.go:226`), so CI passes.

### Consequence

[...]

Issue #688, opened 2026-08-18
Backfill exits 0 when batches fail, so a partial backfill reports success


aristidesstaffieri:
### Summary

`startBackfilling` computes how many batches failed and then discards the result. It returns `nil` regardless, so the process exits 0. A backfill that fetched only part of its range is indistinguishable from one that completed.

### Impact

A Kubernetes backfill Job that loses batches is marked Complete. It is not retried, and no alert fires. The ledger ranges from the failed batches stay absent from `transactions`, `operations`, `state_changes`, and the accounts/balances tables.

The failures are logged, so the evidence exists in the container logs. It is absent from the exit status, which is what the Job controller and the on-call operator act on.

### Code path

1. `analyzeBatchResults` returns the failed-batch count — `internal/services/ingest_backfill.go:37-49`
2. `startBackfilling` calls it and drops the return value — `internal/services/ingest_backfill.go:89`
3. The same function returns `nil` unconditionally — `internal/services/ingest_backfill.go:97`
4. `Ingest` propagates that `nil` — `internal/ingest/ingest.go:118`
5. `RunE: return c.Run(cfg)` gives cobra a `nil` error, so the exit code is 0 — `cmd/ingest.go:187-189`

```go
// internal/services/ingest_backfill.go:89
analyzeBatchResults(ctx, results)   // return value discarded
...
// internal/services/ingest_backfill.go:97
return nil
```

### Why CI does not catch this

`WaitForBackfillCompletion` gates on `oldest <= expectedOldestLedger` (`internal/integrationtests/infrastructure/backfill_helpers.go:226`). The oldest cursor is advanced per batch, so this passes as soon as the single lowest batch commits, whether or not the batches above it succeeded.

The same helper already treats a non-zero exit as a failure (`backfill_helpers.go:235`, "backfill container exited with code %d"). The harness therefore assumes the contract this issue reports as missing: a failed backfill should exit non-zero.

### Reproduction

1. Start a backfill over a range wide enough to produce several batches.
2. Make one batch fail — for example, remove or corrupt one object in the datastore so its ledger fetch exhausts its retries.
3. Observe `Backfilling completed: 1/N batches failed` in the logs.
4. Observe the process exit code is 0, and the Job status is Complete.

### Suggested fix

Return an error when any batch failed, after the recompressor has drained. Compression is safe to complete: already-compressed chunks hold valid data and `compress_chunk` is idempotent.

```go
numFailed := analyzeBatchResults(ctx, results)

[...]

PR #722, opened 2026-08-31
fix: account for cancelled queued backfill batches

### LINKED ISSUE #721: SIGTERM can leave queued backfill batches unrecorded while the one-shot job exits cleanly
### Summary

Historical backfill uses a context-bound Pond task group, but queued batches that are cancelled before their closure starts never write a `BackfillResult`. Because `results` is pre-sized with zero values, those unstarted batches are later indistinguishable from successful batches (`Error == nil`). The task-group cancellation is logged and discarded, and the top-level ingest command turns a SIGINT/SIGTERM cancellation into a clean `nil` return.

The result is a one-shot backfill that can be interrupted with ledger ranges never attempted while still presenting the interruption as a clean exit. This is adjacent to #688 and #689, but it is a separate cancellation-accounting bug; fixing #688 by checking only `analyzeBatchResults` is not sufficient because the cancelled queued batches are currently invisible to that count.

### Current code path

`internal/services/ingest_backfill.go`:

```go
results := make([]BackfillResult, len(batches))
group := m.backfillPool.NewGroupContext(ctx)

for i, batch := range batches {
    group.Submit(func() {
        results[i] = m.processSingleBatch(ctx, batch, i, len(batches))
        if recompressor != nil && results[i].Error == nil {
            recompressor.MarkDone(i, results[i].StartTime, results[i].EndTime)
        }
    })
}

if err := group.Wait(); err != nil {
    log.Ctx(ctx).Warnf("Backfill batch group wait returned error: %v", err)
}

return results
```

Pond v2 checks the task-group context immediately before invoking queued user code. If the group context is already cancelled, it resolves the task with a cancellation error **without invoking the submitted closure**. `wallet-backend` currently uses `github.com/alitto/pond/v2 v2.5.0`.

That means the affected `results[i]` remains the Go zero value:

```go
BackfillResult{
    Batch: BackfillBatch{StartLedger: 0, EndLedger: 0},
    Error: nil,
}
```

`analyzeBatchResults` counts only `Error != nil`, so the unstarted batch is counted as success. `group.Wait()` does know about the cancellation, but `processBackfillBatchesParallel` only logs that error.

At the command boundary, `internal/ingest/ingest.go` intentionally maps root-context cancellation to a clean exit:

```go
if runErr != nil {
    if isShutdownRequested(ctx, runErr) {
        return nil
    }
    ...
}
```

That clean-exit policy is sensible for resumable live ingestion, but a bounded one-shot backfill has a different success contract: every requested gap should either complete or be reported incomplete.

### Why this matters with #688

A direct #688 fix along the lines of:

```go
numFailed := analyzeBatchResults(ctx, results)
if numFailed > 0 { return err }
```

still undercounts any batch cancelled before its closure starts. In the edge case where running work finishes before observing cancellation and only queued work is dropped, `numFailed` can remain zero even though requested ranges were never attempted.

### Reproduction / regression test shape

A deterministic unit test does not need a real ledger backend:

1. Create a Pond pool with concurrency 1.
2. Submit 3 synthetic backfill batches through the same group pattern.
3. Block batch 0 after it starts.
4. Cancel the parent context before batches 1 and 2 begin.
5. Release batch 0 and wait for the group.
6. Assert batches 1 and 2 retain their real `BackfillBatch` identities and are explicitly marked cancelled/not-started.
7. Assert the aggregate backfill result is non-success/incomplete.

The current implementation fails steps 6–7 because those slots remain zero-valued.

### Suggested direction

Make cancellation/incompletion explicit in the result model instead of inferring completion from an untouched zero value. For example:

- initialize every slot before submission with its real `Batch` plus a sentinel error such as `errBatchNotStarted`;
- overwrite the slot only when the closure actually starts/completes;
- return the task-group wait error alongside the results instead of only logging it;
- after all running work has settled, treat any still-sentinel/cancelled batch as an incomplete backfill;
- distinguish backfill interruption from live-ingest shutdown at the command boundary so a bounded backfill cannot be reported as successfully complete unless all requested ranges were accounted for.

A result-task group is another option, but the key invariant is: **every requested `BackfillBatch` must end in an explicit terminal state (`success`, `failed`, or `cancelled/not started`); an untouched zero value must never mean success.**

### Related

- #688 — partial batch failures currently still exit 0.
- #689 — successful out-of-order batches can move `oldest_ingest_ledger` across missing ranges.

Together, these three cases point to the same higher-level invariant worth enforcing in tests: a historical backfill should not publish or report completion beyond the longest contiguous range for which every requested batch reached a verified terminal success state.

ke7haav:
## Summary

Fixes #721.

Pond can cancel queued task-group work without invoking the submitted closure. Those batches previously retained zero-value `BackfillResult` entries, making unattempted ledger ranges appear successful.

## Changes

- Initialize every result with its real batch and an explicit not-started error.
- Preserve the Pond task-group cancellation cause on queued, unstarted results.
- Propagate the task-group wait error after running work and recompression drain.
- Keep clean signal shutdown behavior for live ingestion only.
- Add a deterministic concurrency-1 cancellation regression test.

## Scope

This PR is focused on #721.

It does not change ordinary processed-batch failure handling tracked by #688 or oldest-ledger cursor publication tracked by #689.

## Testing

- Focused regression and shutdown tests
- Focused race tests
- 100-run determinism check
- `make fmt`
- `go vet ./...`
- `golangci-lint v2.1.2`
- Broad database-free race suite
- Repository-wide compile checks

The complete database-backed suite could not run locally because PostgreSQL on `127.0.0.1:5432` rejected the repository’s passwordless test DSN.


internal/services/ingest_backfill.go
```diff
@@ -197,11 +213,17 @@ func (m *ingestService) processBackfillBatchesParallel(ctx context.Context, batc
...
 	}
 
-	if err := group.Wait(); err != nil {
-		log.Ctx(ctx).Warnf("Backfill batch group wait returned error: %v", err)
+	groupErr := group.Wait()
+	if groupErr != nil {
+		groupErr = fmt.Errorf("waiting for backfill task group: %w", groupErr)
+		for i := range results {
+			if errors.Is(results[i].Error, errBackfillBatchNotStarted) {
+				results[i].Error = fmt.Errorf("%w: %w", errBackfillBatchNotStarted, groupErr)
```
aristidesstaffieri, line 221: a SIGTERM will mark every batch as failed and prints an error per batch. Can we consider turning this into a summary log with the count and ledger range? We can keep the per batch error logs for batches that did run.

ke7haav: Addressed. Batches cancelled before starting are now aggregated into a single error log containing the batch count and overall ledger range. Batches that started and failed still produce individual error logs. Cancelled batches continue to count as failures, so the one-shot backfill cannot report success. I also added regression coverage for mixed started/unstarted results and unordered ledger ranges.

aristidesstaffieri (approved):

aristidesstaffieri:
thanks @ke7haav I will go ahead and merge this.

PR #727, opened 2026-09-01
feat(services): 3/n - preview custom SEP-41 token state changes in simulateStateChanges

JiahuiWho:
### What

Extends `simulateStateChanges` to cover custom SEP-41 tokens by mirroring what live ingestion does after the indexer pass, with persistence removed:

1. **`ProtocolProcessor.StagedStateChanges()`** (new interface method):  reads the processor's computed state changes without persisting them to the DB.
2. **Simulation service**: after the main indexer runs, we take the contract events it collected, ask the database which of those contracts are known SEP-41 tokens, run the SEP-41 processor over the events, and add its state changes to the preview.
3. **Wiring**: the serve binary now imports `sep41` so the processor registers itself at startup, and the simulation service gets `Models` so it can do the classification lookup.

### Why

Previously the simulation only ran the main indexer, skips non-SAC token events deliberately because `sep41.Processor` owns them. 

### Known limitations

N/A

https://github.com/stellar/wallet-backend/issues/618

#### PR Structure

#### Thoroughness

#### Release


internal/services/transaction_simulation.go
```diff
@@ -392,17 +401,79 @@ func isSorobanTransaction(envelope xdr.TransactionEnvelope) bool {
...
+// skipped, the same ingestion-lag caveat as SAC metadata enrichment.
+func (s *transactionSimulationService) protocolStateChanges(ctx context.Context, tx ingest.LedgerTransaction, contractEvents map[indexer.ContractEventKey][]xdr.ContractEvent) ([]types.StateChange, error) {
+	if s.models == nil || len(contractEvents) == 0 {
+		return nil, nil
+	}
+	eventContractIDs := distinctEventContractIDs(contractEvents)
+	if len(eventContractIDs) == 0 {
+		return nil, nil
+	}
+	committedByProtocol, err := s.models.ProtocolContracts.BatchGetByContractIDs(ctx, eventContractIDs)
```
leighmcculloch: `SimulateStateChanges` was previously a pure RPC-and-in-memory operation with no DB dependency. Now it hard-fails the entire preview (including the already-computed SAC/native token changes from `buffer.GetStateChanges()`) if this `BatchGetByContractIDs` DB lookup errors.

Scenario: a user previews an ordinary SAC/native token transfer (SAC contracts emit SEP-41-shaped events too, so `eventContractIDs` is non-empty). If the DB pool is briefly exhausted or the connection blips here, `protocolStateChanges` returns an error, `stateChangesForTransaction` discards the already-computed state changes and returns `nil`, and the whole `simulateStateChanges` GraphQL call fails with an internal error — even though the transaction has nothing to do with SEP-41 and the core indexer had already produced a valid preview.

Consider treating a failure here as "skip protocol enrichment" (log + continue with the indexer's changes) rather than failing the whole preview, mirroring the "unclassified contract is silently skipped" behavior already documented for the classification-miss case.

---

internal/services/sep41/processor.go
```diff
@@ -288,6 +288,13 @@ func (p *processor) applyBalanceDelta(account types.AddressBytea, contractStr st
 
 // Reset clears the staged sets for the next window. ledgerNumber is intentionally
 // left untouched — ProcessLedger sets it each ledger.
```
leighmcculloch: The `// Reset clears the staged sets for the next window...` doc comment was left in place when `StagedStateChanges()` was inserted directly below it, so it now documents the wrong function — it sits right above `func (p *processor) StagedStateChanges()` instead of `func (p *processor) Reset()`.

A future reader sees this comment directly above `StagedStateChanges` and could reasonably (but incorrectly) conclude that calling it clears staged state; meanwhile `Reset()` itself is left with no comment at its declaration. Worth moving this comment back down above `Reset()`.

---

internal/services/transaction_simulation.go
```diff
@@ -392,17 +401,79 @@ func isSorobanTransaction(envelope xdr.TransactionEnvelope) bool {
...
+	}
+
+	var stateChanges []types.StateChange
+	for _, protocolProcessor := range protocolProcessors {
+		contracts := committedByProtocol[protocolProcessor.ProtocolID()]
+		if len(contracts) == 0 {
+			continue
+		}
+		input := ProtocolProcessorInput{
+			LedgerSequence:    tx.Ledger.LedgerSequence(),
```
leighmcculloch: `tx.Ledger.LedgerCloseTime()` feeds the new SEP-41 state-change rows here, but the synthesized ledger built for simulation never sets `LedgerHeader.Header.ScpValue.CloseTime`, so every simulated SEP-41 balance/allowance change ends up stamped with `LedgerCreatedAt = 1970-01-01T00:00:00Z` (via `StateChangeBuilder`'s `time.Unix(ledgerCloseTime, 0)`).

The same gap already existed for the SAC/native path via `token_transfer.go`, but this PR is what makes SEP-41 rows carry a bogus epoch timestamp for the first time. Worth setting a real close time on the synthesized ledger (e.g. `time.Now()`) if `LedgerCreatedAt` is ever surfaced to or relied on by callers of this preview.

---

internal/services/transaction_simulation.go
```diff
@@ -392,17 +401,79 @@ func isSorobanTransaction(envelope xdr.TransactionEnvelope) bool {
...
+		return nil, nil
+	}
+
+	// Fresh processor instances per request: protocol processors fold state
+	// across ProcessLedger calls by design, so an instance must never be shared
+	// between simulations.
+	protocolProcessors, err := BuildProcessors(ProtocolDeps{
+		NetworkPassphrase: s.networkPassphrase,
+		Models:            s.models,
+		RPCService:        s.rpcService,
```
leighmcculloch: This builds a full processor instance for every registered protocol (`GetAllProcessorIDs()`), even though `committedByProtocol` (just computed above) already tells you exactly which protocol IDs are actually involved in this transaction.

As more protocol processors get registered over time, every `simulateStateChanges` call pays the allocation/construction cost of building all of them, even when zero or one actually has a matching contract — needless per-request work on a latency-sensitive path. Passing `keys(committedByProtocol)` instead of `GetAllProcessorIDs()` would build only the processors that are actually going to run.

---

JiahuiWho:
@copilot resolve the merge conflicts in this pull request

internal/services/transaction_simulation.go
```diff
@@ -379,17 +388,114 @@ func isSorobanTransaction(envelope xdr.TransactionEnvelope) bool {
...
+	if err != nil {
+		return nil, fmt.Errorf("resolving protocol contracts: %w", err)
+	}
+
+	// A binding changed by the simulated transaction itself must shadow the
+	// committed row: a contract upgraded away from a protocol emits no rows, and
+	// a contract bound to an already-classified wasm emits rows even without a
+	// committed protocol_contracts entry. Classification for the buffered wasm
+	// hashes comes from committed protocol_wasms.
+	var classification map[types.HashBytea]string
```
aditya1702, line 438: Why are we doing classification when generating the simulated state changes? Shouldn't that happen only in live ingestion?

JiahuiWho: Because ingestion will classify this contract when the transaction lands, the preview must classify it too, or it would miss rows history will show. It's in-memory and read-only, nothing is persisted.

PR #728, opened 2026-09-02
sep41: parse map-form burn, clawback, and approve events

### LINKED ISSUE #726: SEP-41 event parsers reject OpenZeppelin #[contractevent] map-format events (burns/clawbacks/approvals silently dropped)
## What

The SEP-41 event parsers in `internal/services/sep41/events.go` only accept the event-data shapes the built-in Stellar Asset Contract emits (bare `i128` / positional `ScVec`). Tokens built with OpenZeppelin's Stellar library emit the same events through the `#[contractevent]` macro, which encodes the data payload as an **ScMap with named fields** (e.g. `{amount: i128}`). Those events fail to parse and are silently skipped.

Affected parsers (line refs at 4711f6ca):

- `ParseBurnEvent` — calls `extractI128(event.Body.V0.Data)` directly (`events.go:133`). OZ burn data is a map → error → event dropped → **burned tokens remain on the holder's tracked balance (balances overstated)**.
- `ParseClawbackEvent` — same direct `extractI128` pattern → OZ clawbacks dropped.
- `ParseApproveEvent` — requires data to be a 2-element `ScVec` `[amount, live_until_ledger]` (`events.go:169-171`). OZ approve data is a map → **allowances from OZ tokens never recorded**.
- Transfer credits appear affected as well: in a controlled loadtest replaying OZ-only traffic, each transfer produced only ~1 balance state change (debit side) instead of the expected debit+credit pair.

## Why it matters

Any real OZ-based token on mainnet hits this. The failure mode is silent — no errors surface, the events just don't become state changes — so balances and allowances for OZ tokens are quietly wrong.

## Evidence

- First observed 2026-08-07 while ingesting `stellar-core apply-load` OZ (`custom_token`) corpus data through the dev loadtest rig.
- Reconfirmed 2026-09-01 on build `4711f6ca` during the per-family ingestion capacity campaign: OZ-only traffic produced ~1 state change per transfer (all `BALANCE/DEBIT`) vs ~3/tx for equivalent SAC traffic.

## Fix direction

Teach the parsers to accept both encodings: try the SAC bare-value/positional shape first, fall back to reading the ScMap by field name (`amount`, `live_until_ledger`). Test fixtures can be generated with apply-load's custom_token mode or the OZ token library directly.

ke7haav:
### What

- Accept SEP-41 Symbol-keyed map data for burn, clawback, and approve events.
- Require amount as i128 and approve live_until_ledger as u32.
- Reject missing, incorrectly typed, nil, non-Symbol-keyed, and duplicate fields while tolerating additional valid Symbol-keyed fields.
- Preserve existing SAC, legacy Soroban, positional, and CAP-67 encodings.

### Why

OpenZeppelin #[contractevent] uses map-form event data. The existing parsers only accepted a bare i128 for burn and clawback and a two-element ScVec for approve, so otherwise-valid events failed parsing and their balance, supply, or allowance changes were not recorded.

### Known limitations

The transfer observation from the issue was not reproduced. Transfer already accepts bare-i128 and map-form amounts, and existing processor tests verify that debit and credit state changes are staged together, so transfer processing is intentionally unchanged.

### Testing

- go test ./internal/services/sep41 -count=1
- go test ./internal/services/sep41 -run 'TestParseTransferEvent|TestProcessor_ProcessEvent' -count=1
- go vet ./internal/services/sep41
- make fmt
- git diff --check

Fixes https://github.com/stellar/wallet-backend/issues/726

Priority: not assigned on the issue.

#### PR Structure

- [x] This PR's title starts with name of package that is most changed in the PR, or all if the changes are broad or impact many packages.

#### Thoroughness

- [x] All updated queries have been tested (no queries were modified).

#### Release

- [x] The new functionality is gated with a feature flag if this is not ready for production (not applicable; this fix is ready for production).

### DIFF
--- a/internal/services/sep41/events.go
+++ b/internal/services/sep41/events.go
@@ -120,7 +120,8 @@ func ParseMintEvent(event xdr.ContractEvent) (*MintEvent, error) {
 	return &MintEvent{To: to, Amount: amt, ToMuxedID: muxedID}, nil
 }
 
-// ParseBurnEvent decodes a SEP-41 burn event: [sym("burn"), from: Address], data = i128.
+// ParseBurnEvent decodes a SEP-41 burn event: [sym("burn"), from: Address],
+// data = i128 or { amount: i128 }.
 func ParseBurnEvent(event xdr.ContractEvent) (*BurnEvent, error) {
 	topics, err := eventTopics(event, EventBurn, 2)
 	if err != nil {
@@ -130,7 +131,7 @@ func ParseBurnEvent(event xdr.ContractEvent) (*BurnEvent, error) {
 	if err != nil {
 		return nil, fmt.Errorf("decoding burn.from: %w", err)
 	}
-	amt, err := extractI128(event.Body.V0.Data)
+	amt, err := extractAmount(event.Body.V0.Data)
 	if err != nil {
 		return nil, fmt.Errorf("decoding burn amount: %w", err)
 	}
@@ -159,7 +160,7 @@ func ParseClawbackEvent(event xdr.ContractEvent) (*ClawbackEvent, error) {
 	if err != nil {
 		return nil, fmt.Errorf("decoding clawback.from: %w", err)
 	}
-	amt, err := extractI128(event.Body.V0.Data)
+	amt, err := extractAmount(event.Body.V0.Data)
 	if err != nil {
 		return nil, fmt.Errorf("decoding clawback amount: %w", err)
 	}
@@ -168,7 +169,8 @@ func ParseClawbackEvent(event xdr.ContractEvent) (*ClawbackEvent, error) {
 
 // ParseApproveEvent decodes a SEP-41 approve event:
 // topics: [sym("approve"), from: Address, spender: Address]
-// data:   [ i128 amount, u32 live_until_ledger ]  (ScVec of 2 elements)
+// data:   [i128 amount, u32 live_until_ledger] or
+// { amount: i128, live_until_ledger: u32 }.
 func ParseApproveEvent(event xdr.ContractEvent) (*ApproveEvent, error) {
 	topics, err := eventTopics(event, EventApprove, 3)
 	if err != nil {
@@ -183,22 +185,34 @@ func ParseApproveEvent(event xdr.ContractEvent) (*ApproveEvent, error) {
 		return nil, fmt.Errorf("decoding approve.spender: %w", err)
 	}
 
+	var amountVal, liveUntilVal xdr.ScVal
 	data := event.Body.V0.Data
-	if data.Type != xdr.ScValTypeScvVec {
-		return nil, fmt.Errorf("approve data must be ScVec, got %v", data.Type)
-	}
-	vec, ok := data.GetVec()
-	if !ok || vec == nil || len(*vec) != 2 {
-		return nil, fmt.Errorf("approve data must be a 2-element ScVec")
+	switch data.Type {
+	case xdr.ScValTypeScvVec:
+		vec, ok := data.GetVec()
+		if !ok || vec == nil || len(*vec) != 2 {
+			return nil, fmt.Errorf("approve data must be a 2-element ScVec")
+		}
+		amountVal = (*vec)[0]
+		liveUntilVal = (*vec)[1]
+	case xdr.ScValTypeScvMap:
+		fields, err := extractRequiredMapFields(data, "amount", "live_until_ledger")
+		if err != nil {
+			return nil, fmt.Errorf("decoding approve data: %w", err)
+		}
+		amountVal = fields["amount"]
+		liveUntilVal = fields["live_until_ledger"]
+	default:
+		return nil, fmt.Errorf("approve data must be ScVec or ScMap, got %v", data.Type)
 	}
 
-	amt, err := extractI128((*vec)[0])
+	amt, err := extractI128(amountVal)
 	if err != nil {
 		return nil, fmt.Errorf("decoding approve amount: %w", err)
 	}
-	liveUntil, ok := (*vec)[1].GetU32()
+	liveUntil, ok := liveUntilVal.GetU32()
 	if !ok {
-		return nil, fmt.Errorf("approve live_until_ledger must be u32, got %v", (*vec)[1].Type)
+		return nil, fmt.Errorf("approve live_until_ledger must be u32, got %v", liveUntilVal.Type)
 	}
 
 	return &ApproveEvent{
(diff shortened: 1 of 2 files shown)


ke7haav:
The integration test failed during PostgreSQL container setup with unexpected EOF, before test assertions ran. @JiahuiWho  please re-run the failed integration-test job? All other checks passed.

internal/services/sep41/events.go
```diff
@@ -253,6 +267,72 @@ func extractI128(val xdr.ScVal) (*big.Int, error) {
...
+// extractRequiredMapFields validates a SEP-41 map data value and returns the
+// requested fields. SEP-41 permits additional fields, but requires every map
+// key to be a Symbol.
+func extractRequiredMapFields(val xdr.ScVal, requiredFields ...string) (map[string]xdr.ScVal, error) {
+	if val.Type != xdr.ScValTypeScvMap {
+		return nil, fmt.Errorf("event data must be ScMap, got %v", val.Type)
+	}
+	if val.Map == nil {
+		return nil, fmt.Errorf("event data map was nil")
+	}
```
JiahuiWho: This is redundant with the `GetMap()` check right below it.

internal/services/sep41/events.go
```diff
@@ -253,6 +267,72 @@ func extractI128(val xdr.ScVal) (*big.Int, error) {
...
+	seen := make(map[string]struct{}, len(*m))
+	for _, entry := range *m {
+		keySym, ok := entry.Key.GetSym()
+		if !ok {
+			return nil, fmt.Errorf("event data map key must be Symbol, got %v", entry.Key.Type)
+		}
+		key := string(keySym)
+		if _, ok := seen[key]; ok {
+			return nil, fmt.Errorf("event data map contains duplicate key %q", key)
+		}
```
JiahuiWho: This rejects non-Symbol keys and duplicate keys, while `extractAmountAndMuxedID` silently skips non-Symbol keys and tolerates duplicates, and that gives us two different strictness policies. I'd align them one way or the other.

internal/services/sep41/events.go
```diff
@@ -253,6 +267,72 @@ func extractI128(val xdr.ScVal) (*big.Int, error) {
 	return bi, nil
 }
 
+// extractAmount decodes the two SEP-41 amount data formats used by burn and
+// clawback events: a single i128 value or a Symbol-keyed map containing amount.
+func extractAmount(val xdr.ScVal) (*big.Int, error) {
```
JiahuiWho, line 272: Quick question: would you mind sharing a reference for the map-form clawback event? 

I ask because when I looked through the OpenZeppelin repo I couldn't find a clawback event defined anywhere. Their clawback helpers call into the built-in SAC, and the SAC emits the plain i128 form we already parse.

internal/services/sep41/events.go
ke7haav: Good point about the apparent redundancy. In go-stellar-sdk v0.7.3, GetMap() dereferences the map union pointer when the discriminator is ScvMap, so it can panic for a malformed value with a nil union arm. I updated this to validate val.Map == nil || *val.Map == nil and then use the checked pointer directly, removing the redundant-looking GetMap() call while preserving panic safety.

internal/services/sep41/events.go
ke7haav: Agreed. I updated extractAmountAndMuxedID to reuse extractRequiredMapFields, so burn, clawback, approve, transfer, and mint now follow the same policy: non-Symbol and duplicate keys are rejected, while additional valid Symbol-keyed fields remain accepted. I also added CAP-67 regression tests for the malformed cases. Valid CAP-67 behavior is unchanged.

JiahuiWho:
The burn and approve fixes look good. I verified the event shapes against the OpenZeppelin source and everything checks out.

On the transfer observation from #726, I have a different theory about what happened, since the parser looks innocent: OZ transfers use `data_format = "single-value"`, so their data is a plain i128 just like SAC, and even a parse failure would drop the whole event rather than half of it. The debit and credit are also staged together in one append in the processor, so a parsed transfer can't produce only a debit.

What does match the observation is a classification failure. If the OZ contract was never classified as SEP-41, both processors skip all of its events, and the only state change left per transaction is the fee, which is recorded as exactly one `BALANCE`/`DEBIT` row. That is the same signature as "one state change per transfer, all debits".

@aditya1702 do you think this makes sense? If so, there are 2 things worth checking in the loadtest data:
1. whether the OZ contract from the campaign has a contract_tokens row with type sep41, or is missing/unknown.
2. which token the observed debit rows are attached to. If they're native XLM fee rows, that would confirm the transfer events never reached the parsers at all.

internal/services/sep41/events.go
ke7haav: Good catch. I could not find an OpenZeppelin-defined clawback event either. The current OpenZeppelin SAC administration helpers delegate clawback to the built-in SAC, which emits the bare i128 form. The reason for retaining map-form clawback support is SEP-41 itself, which specifies clawback data as either a bare i128 or a Symbol-keyed map containing amount: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0041.md#clawback-event. I renamed the test and updated the PR title/wording so map-form clawback is no longer attributed specifically to OpenZeppelin.

ke7haav:
> The burn and approve fixes look good. I verified the event shapes against the OpenZeppelin source and everything checks out.
> 
> On the transfer observation from #726, I have a different theory about what happened, since the parser looks innocent: OZ transfers use `data_format = "single-value"`, so their data is a plain i128 just like SAC, and even a parse failure would drop the whole event rather than half of it. The debit and credit are also staged together in one append in the processor, so a parsed transfer can't produce only a debit.
> 
> What does match the observation is a classification failure. If the OZ contract was never classified as SEP-41, both processors skip all of its events, and the only state change left per transaction is the fee, which is recorded as exactly one `BALANCE`/`DEBIT` row. That is the same signature as "one state change per transfer, all debits".
> 
> @aditya1702 do you think this makes sense? If so, there are 2 things worth checking in the loadtest data:
> 
> 1. whether the OZ contract from the campaign has a contract_tokens row with type sep41, or is missing/unknown.
> 2. which token the observed debit rows are attached to. If they're native XLM fee rows, that would confirm the transfer events never reached the parsers at all.

That theory matches what I found as well. An unclassified contract is skipped before event parsing, while a successfully parsed transfer stages both debit and credit together. I couldn’t reproduce a transfer-parser path that emits only a debit, which is why transfer processing was left unchanged in this PR. Checking the contract_tokens classification and whether the remaining debit rows are native XLM fees should distinguish the classification failure from a parser failure.

JiahuiWho (approved):

PR #730, opened 2026-09-03
Re-land the three non-Blend commits from main-blend onto main

aditya1702:
**To review this:** run the three `range-diff`s below. Check each commit matches what was already merged, minus the Blend parts.

**Merge with rebase-merge, not squash** — reason at the bottom.

## Why this PR exists

Three PRs were reviewed and merged into `main-blend` and never reached `main`. Blend is paused, the persist stack is moving to `main`, and it depends on all three. None of them is a Blend feature.

| Commit | Original | Size | vs the merged original |
|---|---|---|---|
| `86193aca` | #657 | 18 files, +1309/−182 | verbatim |
| `f7dd79d1` | #694 | 2 files, +432/−33 | 2 dead lines dropped |
| `84def4fd` | #692 | 28 files, +1835/−330 | Blend arm removed |

```
git range-diff 5b814edf~1..5b814edf 86193aca~1..86193aca   # #657
git range-diff 08b23c2d~1..08b23c2d f7dd79d1~1..f7dd79d1   # #694
git range-diff af524d52~1..af524d52 84def4fd~1..84def4fd   # #692
```

#657 comes back `=`. #694 and #692 come back `!` — their real differences are below, and the rest is context drift from `main`'s #695 and #723.

## What I dropped

**#694 — 2 lines, both dead on `main`:** an `ingest/sac` import `main` already carries, and an assertion that `FetchSACMetadata` is not called (#723 deleted that service). Its 8 hot-archive tests are unchanged and pass.

**#692 — 4 Blend-only paths:**

```
internal/data/blend/wipe.go
internal/data/blend/wipe_test.go
internal/integrationtests/blend_test.go
internal/services/blend/processor.go   (10-line hunk)
```

No call site moves: `--rebuild` reaches each protocol through the `WipeCurrentState` interface method rather than a registry, so deleting the Blend package deletes its implementation and nothing else refers to it. `sep41` stays as the live implementer.

**#657 is not Blend code**, despite its title. It adds the shared framework — `RequiresContractData()` on `ProtocolProcessor`, `ContractDataChanges` on `ProtocolProcessorInput`, `sep41` implementing it as `false` — and touches nothing under `blend/`. Every protocol on `main` returns `false`, so ContractData extraction stays off.

## Nine commits are new — review these normally

Review on this PR found four bugs, one stale doc, and three duplications, fixed on top so the range-diffs still compare cleanly.

1. **`291aedc7`** — an invalid contract id printed `%!w(<nil>)` instead of its byte count: `%w` wrapped a nil error.
2. **`2b527820`** — `DeleteNamespaceLedgerRange` packs its ledger bounds into a `toid`, which holds the ledger in the high 32 bits of a signed int64. A bound above `MaxInt32` wrapped negative and matched no rows, so the wipe reported success having deleted nothing.
3. **`5c727e6c`** + **`c40bfff1`** — the migrate advisory locks were taken once and never re-checked. A CNPG failover releases them server-side while pgxpool keeps serving healthy connections, so the run writes on while a second run takes the same locks. The rebuild wipes (cursor reset, truncate, every history delete slice) now begin their transactions on the lock-holding connection: a transaction on that session cannot commit once the session, and with it the lock, is gone. Only the wipes need it: the fold loop commits through the cursor CAS, so a second run there loses its CAS and hands off. `TestAcquireMigrateLocks` terminates the lock backend server-side and asserts the transaction fails.
4. **`ada255f5`** — the hot-archive pass skipped archived SAC instances, so `finalize` dropped every live SAC balance whose instance was archived at the checkpoint. The SAC extends its instance TTL by 7 days and balances by 30, so quiet assets hit this routinely, and restoring the instance emits no balance change to recover from it. Archived SAC instances now register like live ones.
5. **`84aa0215`** — `adding-a-protocol.md` documents `RequiresContractData`, `WipeCurrentState` and `ContractDataChanges`.

6. **`d607a401`** — `generateAdvisoryLockID` and `migrateAdvisoryLockID` hash through one `advisoryLockID(key)`; keys unchanged.
7. **`941406b0`** — both rebuild services shared one validate body; it lives on the engine and reads the status column through the strategy.
8. **`93a89987`** — the SEP-41 processor wrapped its wipe error with the same text the data layer already used.
9. **`f4c66ef2`** — the history rebuild takes `ProtocolMigrateHistoryConfig` plus the state-changes model instead of a copied config struct.

## Merge method

**Rebase-merge, not squash.** #679 and the persist stack sit on this branch and shed their own copies of these three commits by patch-id once the commits reach `main`. A squashed commit matches none of the three patch-ids, so the stack would replay all three onto a tree that already contains them.


internal/services/protocol_migrate.go
```diff
@@ -552,6 +687,43 @@ func allHandedOff(trackers []*protocolTracker) bool {
...
+	tracked := map[xdr.ContractId]struct{}{}
+	for _, t := range trackers {
+		if t.handedOff || t.cursorValue+t.pending >= seq || !t.processor.RequiresContractData() {
+			continue
+		}
+		for _, c := range contractsByProtocol[t.protocolID] {
+			idBytes, err := hex.DecodeString(string(c.ContractID))
+			if err != nil || len(idBytes) != len(xdr.ContractId{}) {
+				return nil, fmt.Errorf("protocol %s contract id %q is not a 32-byte hex hash: %w", t.protocolID, c.ContractID, err)
+			}
```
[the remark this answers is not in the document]
aditya1702: Fixed in 9002b545. Decode failures and wrong lengths are separate branches now, and the length message reports the byte count it actually got instead of rendering `%!w(<nil>)`.

internal/services/protocol_migrate_lock.go
```diff
@@ -0,0 +1,95 @@
...
+func migrateAdvisoryLockID(scope, protocolID string) int {
+	h := fnv.New64a()
+	h.Write([]byte("wallet-backend-" + scope + "-" + protocolID))
+	return int(h.Sum64())
+}
+
+// acquireMigrateLocks try-locks every protocol's lock for the given scope on
+// one dedicated connection and returns the release func to defer. A held lock
+// means another run of the same strategy owns that protocol — do not proceed.
+func acquireMigrateLocks(ctx context.Context, pool *pgxpool.Pool, scope string, protocolIDs []string) (func(), error) {
```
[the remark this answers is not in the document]
aditya1702: Fixed in 25f92cdf, with the probe scoped to the wipes.

`acquireMigrateLocks` now returns the held set with a `checkSession` probe alongside `release` — the same `SELECT 1` on the lock-holding connection that live ingestion runs per ledger. The two rebuild wipes call it: current-state before its truncate transaction, history before the cursor reset and again before every delete slice.

Not the fold loop, though. `flushWindow` commits each window through `CompareAndSwap(cursor, winStart-1 -> winEnd)`, so if two runs fold the same protocol one CAS wins and the other discards its window and hands off — a lost lock there is already safe. The wipes are the writes with nothing CAS-gated behind them, and that is where a lost lock costs data: run A deleting rows run B has already re-derived, with A's cursor reporting the range done.

internal/services/checkpoint.go
```diff
@@ -442,6 +482,73 @@ func (p *checkpointProcessor) processContractCode(_ context.Context, wasmHash xd
...
+
+	// Post-side wrapping mirrors the shape the live checkpoint reader produces, so
+	// processContractInstanceChange works identically on archived entries.
+	change := ingest.Change{
+		Type:       entry.Data.Type,
+		ChangeType: xdr.LedgerEntryChangeTypeLedgerEntryCreated,
+		Post:       &entry,
+	}
+	result := p.service.processContractInstanceChange(change, contractAddressStr, contractDataEntry)
+	if result.Skip || result.IsSAC {
```
aristidesstaffieri: If a SAC instance is archived at this time, will we just miss the balance history for it until it gets restored?

internal/services/protocol_processor.go
```diff
@@ -69,6 +70,13 @@ type ProtocolProcessor interface {
 	// migrated with --window-size=1.
 	ProcessLedger(ctx context.Context, input ProtocolProcessorInput) error
 
+	// RequiresContractData reports whether ProcessLedger needs
+	// ProtocolProcessorInput.ContractDataChanges populated. The migration
+	// engine and live ingestion run the (heavier) ContractData extraction
+	// only when a selected processor returns true, so event-only protocols
+	// pay nothing for it.
+	RequiresContractData() bool
```
aristidesstaffieri, line 78: we should up date the docs at `docs/data-migrations/adding-a-protocol.md` to reflect the new interface methods introduced here.

internal/services/checkpoint.go
aditya1702: No history is lost, but this did expose a current-state gap. Fixed in 9c159d97.

**History:** a SAC emits balance events only when invoked, and invoking an archived SAC restores the instance in the same transaction (CAP-62). That `Restored` change has `Post` set and `Pre` nil, and `SACInstanceProcessor` only checks `change.Post != nil`, so the `contract_tokens` row lands in the same ledger as the first post-restore transfer. Every event after that indexes normally.

**Current state (the gap):** this pass skipped archived SAC instances, and `finalize` keeps a live SAC balance only when its contract is in `uniqueContractTokens`. So a live balance under an archived instance was dropped at load, and restoring the instance emits no balance change that could recreate it. This is common, not an edge case: the SAC extends its instance TTL by 7 days and balances by 30 (`stellar_asset_contract/storage_types.rs`), so a quiet asset's instance archives while its holders' balances stay live.

**Fix:** register archived SAC instances the same way as live ones. No wasm, so no `protocol_contracts` row. The skip existed to avoid queuing an RPC metadata fetch for a non-live contract; #723 removed that fetch. Regression test: `TestCheckpointService_PopulateFromCheckpoint_HotArchiveSACInstanceKeepsLiveBalance` (fails without the fix).

internal/services/protocol_processor.go
aditya1702: Done in 3a8b4332. `adding-a-protocol.md` now covers `RequiresContractData`, `WipeCurrentState` and `ContractDataChanges` in the interface listing, the responsibilities table, the example processor, and a Key Patterns bullet on wipe scope.

internal/services/protocol_migrate_current_state_rebuild.go
```diff
@@ -0,0 +1,150 @@
...
+// CAS, so a second run that acquired a silently-released lock loses its CAS and
+// hands off — but two runs truncating and re-deriving under each other lose
+// rows outright.
+func (s *protocolCurrentStateRebuildService) wipe(ctx context.Context, protocolID string, checkLockSession func(context.Context) error) error {
+	if probeErr := checkLockSession(ctx); probeErr != nil {
+		return fmt.Errorf("advisory lock session is no longer alive, the lock may have been lost: %w", probeErr)
+	}
+	processor := s.engine.processors[protocolID]
+	cursorName := s.engine.strategy.CursorName(protocolID)
+	if txErr := db.RunInTransaction(ctx, s.engine.db, func(dbTx pgx.Tx) error {
```
JiahuiWho, line 134: Just curious: the probe on line 129 runs on the dedicated lock connection, but this transaction runs through the pool, so it goes out on a different connection. Why not begin the wipe transaction on the lock connection itself?

 (Same shape in the history rebuild probe at 207 / write at 211, and per slice at 239 / 246)

internal/services/protocol_migrate_history_rebuild.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	// Statuses are not_started after the wipe, so this is a normal migration
+	// run: fold from the retention floor, hand off to live at the frontier.
+	return s.engine.Run(ctx, protocolIDs)
+}
+
+// validate requires each protocol to exist, be classified, and not be marked
+// in_progress (dead-run residue — investigate, don't wipe under it).
+func (s *protocolHistoryRebuildService) validate(ctx context.Context, protocolIDs []string) error {
```
JiahuiWho, line 135: This validate is identical to `protocolCurrentStateRebuildService.validate` except for `p.HistoryMigrationStatus` vs `p.CurrentStateMigrationStatus`, and `engine.strategy.MigrationStatusField` already abstracts that. Is it possible to extract to one shared function?

internal/services/protocol_migrate_lock.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// release unlocks every key and returns the connection to the pool.
+	release func()
+}
+
+// acquireMigrateLocks try-locks every protocol's lock for the given scope on
+// one dedicated connection and returns the held set, whose release is the func
+// to defer. A held lock means another run of the same strategy owns that
+// protocol — do not proceed.
+func acquireMigrateLocks(ctx context.Context, pool *pgxpool.Pool, scope string, protocolIDs []string) (*migrateLocks, error) {
```
JiahuiWho, line 67: This reimplements the lock lifecycle from `ingestService.startLiveIngestion`. `mgrateAdvisoryLockID` is also a copy of `generateAdvisoryLockID` in ingest.go. The failover handling here is subtle and a fix to one copy will not reach the other. Could both call a shared helper that returns `{checkSession, release}`?

internal/services/sep41/processor.go
```diff
@@ -366,6 +370,15 @@ func (p *processor) PersistHistory(ctx context.Context, dbTx pgx.Tx) error {
 	return nil
 }
 
+// WipeCurrentState deletes every SEP-41 current-state row (balances and
+// allowances) in the caller's transaction. See ProtocolProcessor.
+func (p *processor) WipeCurrentState(ctx context.Context, dbTx pgx.Tx) error {
+	if err := sep41data.WipeCurrentState(ctx, dbTx); err != nil {
```
JiahuiWho, line 376: `sep41data.WipeCurrentState` already wraps with "wiping SEP-41 current state", so this produces "wiping SEP-41 current state: wiping SEP-41 current state: ..." and the rebuild adds a third prefix. 

Returning `sep41data.WipeCurrentState(ctx, dbTx)` directly keeps the message to one fact.

internal/services/protocol_migrate_history_rebuild.go
```diff
@@ -0,0 +1,255 @@
...
+	ProtocolsModel         data.ProtocolsModelInterface
+	ProtocolContractsModel data.ProtocolContractsModelInterface
+	IngestStore            *data.IngestStoreModel
+	StateChanges           *data.StateChangeModel
+	NetworkPassphrase      string
+	Processors             []ProtocolProcessor
+	WindowSize             uint32
+	Metrics                *metrics.MigrationMetrics
+	TipProvider            func() (uint32, error)
+}
```
JiahuiWho, line 66: Could this embed `ProtocolMigrateHistoryConfig`, or take `(cfg ProtocolMigrateHistoryConfig, stateChanges *data.StateChangeModel)` like the current-state side?

internal/services/protocol_migrate_lock.go
```diff
@@ -0,0 +1,118 @@
...
+			return nil, fmt.Errorf("%s lock for protocol %q is held: another migration or rebuild is running", scope, pid)
+		}
+		lockIDs = append(lockIDs, lockID)
+	}
+
+	return &migrateLocks{
+		checkSession: func(probeCtx context.Context) error {
+			var one int
+			return conn.QueryRow(probeCtx, "SELECT 1").Scan(&one)
+		},
```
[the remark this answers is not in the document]
aditya1702: Fixed in 9649a162: every write the rebuilds do outside the cursor CAS (the cursor reset, the truncate, and each history delete slice) now begins its transaction on the lock-holding connection. A transaction on that session cannot commit once the session, and with it the lock, is gone, so a failover between two steps fails the write instead of letting it land under another run's lock.

The session probe is removed with the window it guarded. `TestAcquireMigrateLocks` gains a case that terminates the lock backend server-side and asserts a transaction on that connection fails and the lock is free for the next run.

internal/services/protocol_migrate_current_state_rebuild.go
aditya1702: Done that way in 9649a162. `migrateLocks` exposes the lock connection and both wipes take it as their `db.TxBeginner`: the cursor reset and truncate (current state), and the reset plus every delete slice (history), begin on it. `DeleteNamespaceLedgerRange` takes the beginner so its own `SET LOCAL` + `DELETE` transaction runs there too. The probe is gone; the transaction failing is the check.

internal/services/protocol_migrate_history_rebuild.go
aditya1702: Done in 0b4a8ea3: one `validateRebuild` on the engine, reading the status through `strategy.MigrationStatusField` and naming the strategy in the in_progress message through `strategy.Label`. Both per-service copies are gone.

internal/services/protocol_migrate_lock.go
aditya1702: Partly. f9557653 shares the hash: `generateAdvisoryLockID` and `migrateAdvisoryLockID` both call one `advisoryLockID(key)` now, keys unchanged.

I left the two acquire/release lifecycles separate on purpose. They diverge right after acquire: live holds one lock and fences its per-ledger writes with the cursor CAS plus a session probe, while the rebuilds hold one lock per protocol and, as of 9649a162, fence their writes by running them on the lock connection, so they have no probe at all. A shared `{conn, release}` helper would be ~30 lines saved and would touch `startLiveIngestion` in a reland PR. If you want it, I would do it as a follow-up on `main` once this lands.

internal/services/sep41/processor.go
aditya1702: Done in f30e6f76: the processor returns `sep41data.WipeCurrentState` directly (with a `//nolint:wrapcheck`, since the linter flags any bare error crossing a package boundary). The chain now reads "wiping current state for SEP41: wiping SEP-41 current state: <pg error>".

internal/services/protocol_migrate_history_rebuild.go
aditya1702: Done in a54ef219: `NewProtocolHistoryRebuildService(cfg ProtocolMigrateHistoryConfig, stateChanges *data.StateChangeModel)`, matching the current-state side. `ProtocolHistoryRebuildConfig` is gone.

JiahuiWho (approved):

PR #734, opened 2026-09-09
fix: record SAC-created trustline state changes

### LINKED ISSUE #731: state_changes: SAC-created trustlines produce no trustline state change (CAP-0073)
## Summary

CAP-0073 adds `trust(address)` to the Stellar Asset Contract. It creates an unlimited (`i64::MAX`) trustline for a G-address and emits no contract event. Our balance tables handle this. `state_changes` does not.

A trustline created by `trust()` writes no `TRUSTLINE` / ADD row and no `BALANCE_AUTHORIZATION` row. `trustline_balances` stays correct, so the gap is silent: the balance appears with no history entry that explains it.

## Cause

Effects are gated on operation type. `Effects()` sends `InvokeHostFunction` to `default:` and returns no effects (`internal/indexer/processors/effects_horizon.go:120-136`). `EffectTrustlineCreated` is produced only inside `addChangeTrustEffects()` (`internal/indexer/processors/effects_horizon.go:489,512`), which runs only for `OperationTypeChangeTrust`. The consumer in `internal/indexer/processors/effects.go:202-221` therefore never fires.

## What should be added

Emit trustline state changes for trustline ledger-entry changes produced by Soroban operations. Do this in `SACEventsProcessor` (`internal/indexer/processors/contracts/sac.go`) rather than in the Horizon effects copy, which should stay a faithful copy. That processor already handles `InvokeHostFunction` only (`sac.go:66`) and already reads `tx.GetOperationChanges(opWrapper.Index)` (`sac.go:88`).

1. Add a pass over the operation changes that selects `xdr.LedgerEntryTypeTrustline` entries with `Pre == nil` and `Post != nil`. Skip `xdr.AssetTypeAssetTypePoolShare`, matching `internal/indexer/processors/trustlines.go:99`.
2. Build a `StateChangeCategoryTrustline` state change with the ADD reason. Reuse the existing builder and the asset-contract-ID helper the trustline-flag path in that file already uses.
3. Emit the matching `StateChangeCategoryBalanceAuthorization` state change from the new trustline's flags. Mirror `EffectsProcessor.generateBalanceAuthorizationForNewTrustline` (`internal/indexer/processors/effects.go:214`). Extract that flag-reading helper to a shared location if the `contracts` package cannot call it.
4. `extractTrustlineFlagChanges` already meets this case and rejects it with `errNoPreviousTrustlineFlagChangesFound` (`sac.go:331`). The new pass is the positive handling of the same case.
5. A REMOVE branch is optional. A SAC cannot delete a trustline under CAP-0073. Add it only if it is free, and record the decision in a comment.
6. No new ordinal namespace is needed. `SACEventsProcessor.StateChangeSubBase()` returns `types.StateChangeSubBaseSACEvents` (`sac.go:53`).

## Already covered — no work needed

- `trustline_balances`: `TrustlinesProcessor` reads operation changes with no operation-type gate, so a Soroban-created trustline lands as `TrustlineOpAdd` (`internal/indexer/processors/trustlines.go:47,55,80`).
- `trustline_assets`: ids are deterministic UUIDv5 over `CODE:ISSUER`, so a new asset needs no pre-registration (`internal/db/migrations/2025-12-14.0-trustline_assets.sql:5`).
- Cold start: the history-archive bootstrap is entry-based (`internal/services/checkpoint.go:489`).
- Participants: `InvokeContractOpProcessor.Participants()` unions the Soroban auth entries (`internal/indexer/processors/contract_operations.go:356`), and `trust()` calls `require_auth` on the trustor.
- `sac_balances`: G-address holders are rejected by `sac.ContractBalanceFromContractData` (`internal/indexer/processors/sac_balances.go:101`), so a G-account trustline needs no handling there.
- The CAP's second change — XLM SAC `transfer` creating an `AccountEntry` — needs no work. `AccountsProcessor` records it as `AccountOpCreate` (`internal/indexer/processors/accounts.go:88,170`), and `EffectAccountCreated` maps to no state change today, so classic `CreateAccount` is untracked as well. No regression.
- SEP-41 is unaffected. SACs have no WASM and never pass the signature check (`internal/services/sep41/validator.go:292`). `wasmspec.Match` ignores extra functions, so a WASM token that also exposes `trust()` still classifies (`internal/services/wasmspec/wasmspec.go:66`). The SEP-41 processor folds events only (`internal/services/sep41/processor.go:98`) and `trust()` emits none.

## Verification

- Add a case to `internal/indexer/processors/contracts/sac_test.go`. Feed an `InvokeHostFunction` operation whose meta carries a created `TrustLineEntry` with limit `i64::MAX` and no contract event. Assert one TRUSTLINE ADD state change and one BALANCE_AUTHORIZATION state change.
- `go test ./internal/indexer/... ./internal/services/sep41/...`
- Live integration needs protocol 26. If the pinned quickstart image does not serve it, build the meta fixture by hand following `internal/indexer/processors/contracts/test_utils.go` and state the limit in the PR.

## References

- CAP-0073: https://github.com/stellar/stellar-protocol/blob/master/core/cap-0073.md

ke7haav:
## Summary

Fixes #731.

CAP-0073 introduces `trust(address)` for Stellar Asset Contracts. For classic accounts, this invocation can create a trustline without emitting a corresponding contract event. Although the resulting balance was captured in `trustline_balances`, the state-change processor did not record the trustline creation or its initial authorization state.

## Changes

- Detect newly created `TrustLineEntry` records in SAC operation ledger-entry changes.
- Emit a `TRUSTLINE / ADD` state change containing the account, SAC contract ID, and trustline limit.
- Emit the corresponding `BALANCE_AUTHORIZATION / SET` state change based on the trustline flags.
- Ignore liquidity-pool-share trustlines, consistent with existing trustline processing.
- Reuse the existing trustline flag mapping used by the classic effects processor.
- Avoid duplicate authorization state changes when trustline creation is accompanied by a `set_authorized` event.
- Preserve the existing SAC state-change ordinal namespace.

## Tests

Added deterministic coverage for:

- CAP-73 trustline creation without contract events.
- Account and SAC contract identity.
- Unlimited trustline limit conversion.
- Initial authorization and clawback flags.
- Ignoring pool-share trustlines.
- Avoiding duplicate authorization changes during trustline creation.

Validation performed:

- `go test ./internal/indexer/processors/contracts/... -count=1`
- `go test ./internal/indexer/... ./internal/services/sep41/... -count=1`
- Regression test repeated 20 times.
- Race-enabled contracts package tests.
- Full CI-style unit suite: 2,530 tests passed, 3 intentionally skipped.
- Coverage: 76.1%, above the 65% CI threshold.
- `go build ./...`
- `go vet ./...`
- `golangci-lint`
- `deadcode`
- `gofmt`
- `goimports`
- Module verification and diff checks.

The unrelated SEP-41 current-state migration integration assertion also fails identically when run against an independently built, unmodified `upstream/main` image; it is not introduced by this change.

### DIFF
--- a/internal/indexer/processors/contracts/sac.go
+++ b/internal/indexer/processors/contracts/sac.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/stellar/go-stellar-sdk/amount"
 	"github.com/stellar/go-stellar-sdk/strkey"
 	"github.com/stellar/go-stellar-sdk/support/log"
 	"github.com/stellar/go-stellar-sdk/xdr"
@@ -90,8 +91,11 @@ func (p *SACEventsProcessor) ProcessOperation(_ context.Context, opWrapper *proc
 		return nil, fmt.Errorf("getting operation changes for operation %d: %w", opWrapper.ID(), err)
 	}
 
-	stateChanges := make([]types.StateChange, 0)
 	builder := processors.NewStateChangeBuilder(ledgerNumber, ledgerCloseTime, txID, p.metricsService).WithOperationID(opWrapper.ID())
+	stateChanges, err := p.processCreatedTrustlines(changes, builder)
+	if err != nil {
+		return nil, fmt.Errorf("processing created trustlines for operation %d: %w", opWrapper.ID(), err)
+	}
 	for _, event := range contractEvents {
 		// Validate basic contract contractEvent structure
 		if event.Type != xdr.ContractEventTypeContract || event.ContractId == nil || event.Body.V != 0 {
@@ -178,7 +182,11 @@ func (p *SACEventsProcessor) ProcessOperation(_ context.Context, opWrapper *proc
 			} else {
 				// For classic account addresses, check trustline flag changes
 				wasAuthorized, wasMaintainLiabilities, err = p.extractTrustlineFlagChanges(changes, accountToAuthorize, contractID)
-				if err != nil && !errors.Is(err, errNoPreviousTrustlineFlagChangesFound) {
+				if errors.Is(err, errNoPreviousTrustlineFlagChangesFound) {
+					// The creation pass already emitted the trustline's initial authorization state.
+					continue
+				}
+				if err != nil {
 					// We dont want to log errors for no trustline change found. We can simply skip processing.
 					if !errors.Is(err, errNoTrustlineChangeFound) {
 						log.Debugf("processor: %s: extracting trustline flag changes: txHash=%s opID=%d contractId=%s accountAddress=%s error=%v",
@@ -249,6 +257,43 @@ func (p *SACEventsProcessor) ProcessOperation(_ context.Context, opWrapper *proc
 		default:
 			continue
 		}
+	}
+	return stateChanges, nil
+}
+
+func (p *SACEventsProcessor) processCreatedTrustlines(changes []ingest.Change, builder *processors.StateChangeBuilder) ([]types.StateChange, error) {
+	stateChanges := make([]types.StateChange, 0)
+	for _, change := range changes {
+		if change.Type != xdr.LedgerEntryTypeTrustline || change.Pre != nil || change.Post == nil {
+			continue
+		}
+
+		trustline := change.Post.Data.MustTrustLine()
+		if trustline.Asset.Type == xdr.AssetTypeAssetTypePoolShare {
+			continue
+		}
+
+		asset, err := trustLineAssetToAsset(trustline.Asset)
+		if err != nil {
+			return nil, fmt.Errorf("converting trustline asset: %w", err)
+		}
+		assetContractID, err := asset.ContractID(p.networkPassphrase)
+		if err != nil {
+			return nil, fmt.Errorf("getting trustline asset contract ID: %w", err)
+		}
+
+		contractID := strkey.MustEncode(strkey.VersionByteContract, assetContractID[:])
+		account := trustline.AccountId.Address()
+		baseBuilder := builder.Clone().WithAccount(account).WithToken(contractID)
+		limit := amount.String(trustline.Limit)
+		stateChanges = append(stateChanges,
+			baseBuilder.Clone().
+				WithCategory(types.StateChangeCategoryTrustline).
+				WithReason(types.StateChangeReasonAdd).
+				WithTrustlineLimit(nil, &limit).
+				Build(),
+			processors.BuildBalanceAuthorizationForNewTrustline(baseBuilder, xdr.TrustLineFlags(trustline.Flags)),
+		)
 	}
 	return stateChanges, nil
 }
--- a/internal/indexer/processors/effects.go
+++ b/internal/indexer/processors/effects.go
@@ -315,7 +315,7 @@ func (p *EffectsProcessor) generateBalanceAuthorizationForNewTrustline(baseBuild
 		return types.StateChange{}, nil
 	}
 
-	var defaultFlags []string
+	var trustlineFlags xdr.TrustLineFlags
 	if assetType == "liquidity_pool_shares" {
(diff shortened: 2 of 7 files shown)


internal/indexer/processors/effects.go
```diff
@@ -330,21 +330,22 @@ func (p *EffectsProcessor) generateBalanceAuthorizationForNewTrustline(baseBuild
...
+}
+
+// BuildBalanceAuthorizationForNewTrustline builds the initial authorization state from a new trustline's flags.
+func BuildBalanceAuthorizationForNewTrustline(baseBuilder *StateChangeBuilder, flags xdr.TrustLineFlags) types.StateChange {
 	return baseBuilder.Clone().
 		WithCategory(types.StateChangeCategoryBalanceAuthorization).
 		WithReason(types.StateChangeReasonSet).
-		WithFlags(defaultFlags).
-		Build(), nil
+		WithFlags(mapTrustlineFlagsToStrings(flags)).
```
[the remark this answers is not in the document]
ke7haav: WithFlags previously treated an empty flag set as absent, causing zero-flag classic trustlines to be persisted with a null flags column. The builder now preserves an explicitly supplied empty set as a valid zero bitmask, and zero-bitmask trustline decoding returns a non-nil empty slice so GraphQL exposes []. SAC contract-holder authorization continues to return null. Added CAP-73, decoder, and GraphQL resolver regression coverage.

internal/indexer/processors/contracts/sac.go
```diff
@@ -253,6 +261,43 @@ func (p *SACEventsProcessor) ProcessOperation(_ context.Context, opWrapper *proc
...
+		}
+
+		trustline := change.Post.Data.MustTrustLine()
+		if trustline.Asset.Type == xdr.AssetTypeAssetTypePoolShare {
+			continue
+		}
+
+		asset, err := trustLineAssetToAsset(trustline.Asset)
+		if err != nil {
+			return nil, fmt.Errorf("converting trustline asset: %w", err)
```
aristidesstaffieri, line 278: it looks like this couldn't fail today but it would be more consistent to debug log and skip in case that a future protocol upgrade adds a new arm to `TrustLineAsset`, then ingestion can at least continue until it gets added.

internal/indexer/processors/state_change_builder.go
```diff
@@ -78,10 +78,8 @@ func (b *StateChangeBuilder) WithTrustlineLimit(oldValue, newValue *string) *Sta
 
 // WithFlags sets the flags as a bitmask from a slice of flag names
 func (b *StateChangeBuilder) WithFlags(flags []string) *StateChangeBuilder {
-	if len(flags) > 0 {
-		bitmask := types.EncodeFlagsToBitmask(flags)
-		b.base.Flags = sql.NullInt16{Int16: bitmask, Valid: true}
-	}
+	bitmask := types.EncodeFlagsToBitmask(flags)
```
aristidesstaffieri, line 81: this will change the API layer response from null to an empty array because `trustlineFlags` stays at it's 0 value for that type. This seems unintentional so we should maintain the existing behavior there.

internal/indexer/processors/contracts/sac.go
ke7haav: Updated. Unsupported TrustLineAsset variants are now debug-logged and skipped instead of failing operation processing, allowing ingestion to continue until support for a future asset arm is added. Added regression coverage using a synthetic unsupported asset type.

internal/indexer/processors/state_change_builder.go
ke7haav: Updated. The shared authorization helper now distinguishes absent flags from an explicit zero bitmask. Liquidity-pool trustlines pass no classic flags and retain the existing null API representation, while classic trustlines with a zero bitmask continue to produce an empty flag list. Added regression coverage for the liquidity-pool behavior.

aristidesstaffieri (approved):

PR #738, opened 2026-09-15
fix(serve): bound HTTP metric labels to route patterns and known methods

aditya1702:
### What

Bound the labels on the HTTP request metrics so a client cannot create new Prometheus series.

- `endpoint` is the chi route pattern (`/health`, `/graphql/query`). Unmatched paths become `unmatched`; a 404 under the GraphQL mount becomes `/graphql/*`.
- `method` is one of the seven standard verbs, anything else becomes `OTHER`.
- `wallet_http_request_duration_seconds` is now a histogram instead of a summary.
- `/api-metrics` allows 5 concurrent scrapes with a 10s timeout.

### Why

The middleware used the raw path and method as label values. Both are chosen by the client, and the middleware runs before authentication. Each new pair permanently allocated ~68 KB, so anyone could grow the heap until the pod died.

### Breaking change

The `quantile=` series are gone. Dashboards need `histogram_quantile(...)` over `wallet_http_request_duration_seconds_bucket`. Only the local Grafana JSON used the old series; it is updated.

Internal security report.

#### PR Structure

#### Thoroughness

#### Release


internal/metrics/http.go
```diff
@@ -14,10 +14,12 @@ func newHTTPMetrics(reg prometheus.Registerer) *HTTPMetrics {
...
-		RequestsDuration: prometheus.NewSummaryVec(prometheus.SummaryOpts{
-			Name:       "wallet_http_request_duration_seconds",
-			Help:       "Duration of HTTP requests in seconds.",
-			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		// A histogram costs ~200 B per label set against ~68 KB for a summary with
+		// quantile objectives, and its buckets aggregate across replicas.
+		RequestsDuration: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "wallet_http_request_duration_seconds",
+			Help:    "Duration of HTTP requests in seconds.",
+			Buckets: []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0},
```
aristidesstaffieri: The highest bucket is `10.0`, but `requestContextTimeout` is `30 * time.Second` so any request between 10s and 30s falls only into +Inf, the previous summary did include this range, should this still surface slow requests under the timeout?

aditya1702: Good catch. Added a `30.0` bucket, matching `requestContextTimeout`, so a 10–30s request lands in a finite bucket. Commit 9eabeb5c.

internal/serve/middleware/metrics_middleware_test.go
```diff
@@ -0,0 +1,74 @@
...
+	mux.Route("/graphql", func(r chi.Router) {
+		r.Handle("/query", ok)
+	})
+
+	do := func(method, path string) {
+		mux.ServeHTTP(httptest.NewRecorder(), httptest.NewRequest(method, path, nil))
+	}
+	for i := range 200 {
+		do(http.MethodGet, fmt.Sprintf("/does-not-exist/%d", i)) // 404, nothing matched
+		do(fmt.Sprintf("VERB%d", i), "/graphql/query")           // 405, unknown method token
```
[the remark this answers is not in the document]
aditya1702: chi decides 405 before routing: `routeHTTP` looks up `methodMap[r.Method]` and calls the MethodNotAllowed handler when the token is unknown, so `Handle` never sees the request. The test asserts exactly that series and passes (200 hits on `unmatched/OTHER/405`). No change.

aristidesstaffieri (approved):

PR #739, opened 2026-09-15
fix(indexer): derive Soroban participants from executed meta, not the auth tree

aditya1702:
### What

Stop reading Soroban participants from the transaction's auth entries. Read them from what the host actually executed.

- Addresses that authorised a call come from the nonce ledger entries the host writes after a signature verifies.
- Nested contracts come from the contract events they emitted.
- The auth-tree walk is deleted, including its quadratic `Union` merges.
- Failed transactions attribute only the tx source and fee-bump source. A failed payment no longer shows up in the destination's history.

### Why

The auth entries are written by whoever submits the transaction. The host only checks an entry when a `require_auth` call matches it, so an unmatched entry naming any account passes with no signature. That let anyone insert a transaction into any account's history for a fraction of a cent, and the same held for the entry's own credential address, not only the nested tree.

Checked on testnet: across 6,000 transactions, all 92 address-credential entries in successful transactions had a matching nonce entry, and none of the 2,256 source-account-only transactions had one.

### Known limitations

A nested call that emits no event and writes no state is no longer a participant. Its authoriser still is, through the nonce entry.

### Testing

Regression tests for the forged auth tree, nonce-derived authorisers, event-derived contracts, and failed-transaction attribution. Plus a replay of a real protocol-28 testnet transaction (V4 meta) in `testdata/`. All fail on the old code.

Internal security report.

#### PR Structure

#### Thoroughness

#### Release


internal/indexer/processors/contract_operations.go
```diff
@@ -48,317 +46,127 @@ func calculateContractID(networkPassphrase string, fromAddress xdr.ContractIdPre
...
+//   - The operation source account is always included.
+//   - InvokeContract: the invoked contract.
+//   - CreateContract(V1/V2): the deployer address and the derived contract ID (FromAddress),
+//     or the SAC ID derived from the asset (FromAsset). The host calls require_auth on the
+//     FromAddress, so a successful top-level deploy is authenticated.
+//   - Every address that authorised the invocation. After a signature verifies, the host
+//     writes a temporary ContractData entry with key type SCV_LEDGER_KEY_NONCE under the
+//     authorising address; only the host can write such a key. Source-account credentials
+//     consume no nonce, and that account is already included.
+//   - Every contract that emitted an event while executing.
```
JakeUrban, line 64: I think this is a new set of participants compared to the prior implementation right?

Can we state our rationale for why we consider each set mentioned above a participant?

IIUC, a "participant" is an account (note: not just an address) involved in the transaction/operation. We track participants so that we can serve transactions-for-account and similar queries. Is that correct, or is our definition of and use for a participant broader than that?

The reason I ask is because I'm wondering why we shouldn't simplify our logic by only including addresses that authorized the transaction. Thinking about our current set:

- op source: has to authorize the operation
- invoked contract: contracts can require their own authorization, but its not necessary in many cases. For example, a SEP-41 token contract doesn't require its own authorization, and should be considered an account.
- contract creator / created: I think we can use the rationale in the above point for the creator. If the creator required its own authorization, we can consider it an account. For the created address, its a bit trickier because they wouldn't need to authorize their own creation. Alternatively, we could only index the created contract if it had the special `__check_auth` function.
- every contract that emitted an event: I'm not sure about this one. Lots of non-account contracts can be involved in a transaction and we shouldn't necessarily believe they're an account.

💡 something to consider: there may be value in considering non-accounts as participants. For example, it could be value for clients if they could make a transactions-for-token query, or transactions-for-contract more generally. That would be an expansion of scope for the API, so we should consider it outside this PR.

aristidesstaffieri:
it looks like `contract_deploy.go` still walks InvokeHostFunctionOp.Auth so the same injection channel works through that state change path.

PR #741, opened 2026-09-15
fix(indexer): record a declared nested deploy only when the meta shows it happened

aditya1702:
### What

Record a nested contract deploy only when the operation meta contains the created contract instance entry. Stacked on #739.

Top-level deploys (the host function itself) are unchanged. The declared auth tree is still read for the deployer address, but each `CreateContract` node is kept only if the meta proves that contract was created.

### Why

A `CreateContract` nested in an auth entry is a claim by the submitter. The host does not check unmatched entries, so a transaction could record that any account created a contract that does not exist.

No node cap: the walk is linear with two hashes per node. The quadratic cost in the issue was the participant walk, deleted in #739.

Internal security report.

#### PR Structure

#### Thoroughness

#### Release


internal/indexer/processors/contract_deploy.go
```diff
@@ -56,43 +56,73 @@ func (p *ContractDeployProcessor) ProcessOperation(_ context.Context, op *Transa
...
 			return fmt.Errorf("deployer address to string: %w", err)
 		}
-
 		seen[contractID] = struct{}{}
 		stateChanges = append(stateChanges, builder.Clone().WithAccount(contractID).WithCreator(deployerAddr).Build())
 		return nil
 	}
 
+	// The top-level host function is authenticated: the host calls require_auth on the
+	// FromAddress, so a successful operation proves this deploy and its deployer.
```
aristidesstaffieri: it looks like this is true for a failed tx as well?

aditya1702: Yes, it was. State-change processors run for failed transactions too, so a failed top-level `CreateContract` emitted a creation record. `ProcessOperation` now returns early when the transaction failed, with a test. Commit fe3f0729.

aditya1702:
Addressed Copilot's suppressed note on `contract_deploy_test.go`: the fabricated-deploy regression now puts an unrelated created contract instance in the meta, so an implementation that accepted any created instance fails it (checked by weakening the lookup locally). Commit 3f9375b4.

aristidesstaffieri (approved):

PR #745, opened 2026-09-16
fix(indexer): skip contract deploys from unstorable ScAddress kinds

JakeUrban:
### What

Guard the create-contract deployer address against `ScAddress` kinds an account column cannot store.

- **`utils.go`:** new `deployerAddressString`, which converts the address and reports whether it is storable. The allowlist of `ACCOUNT`, `CONTRACT` and `MUXED_ACCOUNT` mirrors what `AddressBytea.Value()` accepts.
- **`contract_deploy.go`:** `processCreate` skips the deployment when the deployer is not storable. All four call sites (`CreateContract` / `CreateContractV2` at the root host function, and both create-contract arms of the auth tree walk) funnel through this one function.
- **`contract_operations.go`:** `contractIDsForPreimage` omits a non-storable deployer from the participant set. The contract ID derived from the preimage is still included, leaving the rest of the behaviour unchanged.

Both readers share the helper because they share a reason: the deployer address is the one place a CAP-0067 arm survives to ingestion, because the host keeps a create-contract preimage as XDR and compares it raw instead of converting it. Addresses the host does convert — auth credentials, a contract being invoked — trap at apply and are not guarded here.

### Why

Anyone could permanently halt ingestion with a single, cheap transaction.

Protocol 23 introduced two new kinds of address — claimable balances and liquidity pools — that may appear wherever an address is expected, including as the account deploying a contract. Neither is an account, and neither can ever actually deploy one. The network does not check this up front, so such a transaction is accepted, charged its fee and written into a ledger. There are two ways it then reaches ingestion.

**Through a failed transaction.** The deployment fails once the ledger is applied. The indexer derives state changes from every transaction in a ledger, failed ones included, and took the deploying address at face value — recording a claimable balance or a liquidity pool as the account that created a contract. The two kinds then behaved differently:

- **Claimable balance — availability.** The address does not fit a column that stores accounts, so writing the ledger failed. The failure was in the data rather than the connection, so retrying could not succeed: ingestion stopped at that ledger, never advanced past it, and failed identically after a restart. Rebuilding from history stopped at the same ledger. Since the transaction is a permanent part of ledger history, the service could not recover on its own.
- **Liquidity pool — integrity.** The address happens to be the right length, so it was stored without complaint, leaving a liquidity pool recorded as the creator of a contract.

**Through a successful one.** A transaction may declare more permissions than it ends up using, and a declared deployment that is never carried out is ignored when the ledger is applied — so the transaction succeeds normally. The indexer reads the accounts involved in an operation out of those declarations, used or not, and a claimable balance named there as a deploying account stops the write the same way. This route needs no failed transaction, so the indexer's existing habit of ignoring the accounts involved in a failed transaction does not help.

Triggering either needed no privileged access and no interaction from anyone else, and cost the sender only the transaction fee. Every deployment ingesting the affected ledger would halt at the same point, and each would go on serving data frozen at that ledger for as long as it stayed halted.

Ingestion already declines to treat these two kinds of address as accounts elsewhere; contract deployment was missing the same handling.

Reported via HackerOne [4025982](https://hackerone.com/reports/4025982).

#### PR Structure

#### Thoroughness

- [ ] All updated queries have been tested (refer to [this](https://stackoverflow.com/a/29163465) check if the data set returned by the updated query is expected to be same as the original one). — no queries changed.

#### Release

https://claude.ai/code/session_014rhuFdyQAj3wrHeU2QzsYv


internal/indexer/processors/contract_deploy.go
```diff
@@ -57,6 +57,25 @@ func (p *ContractDeployProcessor) ProcessOperation(_ context.Context, op *Transa
...
+		// that merely failed at apply.
+		//
+		// Skip the deployment entirely rather than emitting a state change with an unstorable
+		// creator. Per CAP-0067 the host refuses to convert a claimable-balance or
+		// liquidity-pool address to a host object, so a deploy from one can only ever fail:
+		// there is no contract to record. Emitting it anyway would put a B... strkey in
+		// creator_account_id, whose 33-byte payload fails AddressBytea.Value(); that failure is
+		// wrapped as data.ErrRowEncoding, which isPermanentPersistError treats as permanent, so
+		// it aborts the ledger's persist transaction without advancing the cursor and repeats
+		// identically on every retry and restart.
```
JakeUrban: This is way too much information. No need to distill the entire context for this filter.

internal/indexer/processors/utils.go
```diff
@@ -294,3 +294,30 @@ func ConvertOperation(
...
+// CAP-0067 (protocol 23) widened the ScAddress union with SC_ADDRESS_TYPE_CLAIMABLE_BALANCE and
+// SC_ADDRESS_TYPE_LIQUIDITY_POOL. Neither is an account, and they fail differently: a
+// claimable-balance strkey (B...) carries a 33-byte payload that AddressBytea.Value() rejects
+// outright, while a liquidity-pool strkey (L...) is 32 bytes and would encode silently into a
+// column meant to hold an account or contract. The classic path already refuses both (see
+// isClaimableBalance / isLiquidityPool above, "we dont track them as accounts").
+//
+// This is an allowlist rather than a claimable-balance/liquidity-pool denylist so that any arm a
+// later protocol adds to the union is excluded until it is handled explicitly: an address kind
+// these columns do not recognize is, by definition, one they cannot store.
```
JakeUrban: Remove the 2nd and 3rd components to this comment.

internal/indexer/processors/contract_deploy.go
JakeUrban: Cut down to two lines in c120041 — the background now lives in the commit message and PR description instead.

---

internal/indexer/processors/utils.go
JakeUrban: Dropped both paragraphs in c120041; only the first (what the function reports) remains.

---

internal/indexer/processors/contract_deploy.go
```diff
@@ -57,6 +57,12 @@ func (p *ContractDeployProcessor) ProcessOperation(_ context.Context, op *Transa
 	seen := map[string]struct{}{}
 
 	processCreate := func(fromAddr xdr.ContractIdPreimageFromAddress) error {
+		// The host rejects a deployer address it can't convert to a host object, so a deploy
+		// from one never creates a contract and there is nothing to record.
+		if !isStorableAccountAddress(fromAddr.Address) {
+			return nil
```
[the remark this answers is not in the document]
JakeUrban: Half right, and I've fixed that half in 7015b21.

**Correct:** `goodPreimage` was a child of the rejected create, not a sibling, so the test incidentally asserted that a rejected node's descendants still get emitted — which is exactly the open question, not something this PR should be pinning. The fixture now hangs both deployments off a common `ContractFn` parent as true siblings, so the test covers only what the guard itself decides. Verified it still fails without the guard.

**Not introduced here:** the guard can only *remove* emissions. Before it, a claimable-balance node emitted a poisoned state change *and* descended into its subinvocations; after it, that node emits nothing *and* descends. The descent is untouched — the whole diff to this file is the five added lines of the guard — so children behave identically either way, for every address kind.

The broader point stands on its own merits: this processor emits contract creations from auth declarations regardless of whether they executed, and regardless of whether the transaction succeeded at all. That predates this change, is independent of address kind, and affects every ledger — not just ones carrying a CAP-0067 address. Correlating auth-tree candidates against contract-instance entries in the operation metadata is a real improvement, but it's a behavioural redesign of the processor, and folding it into a minimal security fix would widen the blast radius of a change that wants to be easy to review and backport. Worth its own issue.

---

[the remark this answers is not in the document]
JakeUrban: Tracked separately in #746, which covers all three ways this happens — failed transactions, authorized-but-unperformed deployments, and deployments named underneath a rejected one — and proposes the fix you suggested: check each named deployment against the contracts the ledger says were actually created.

---

JiahuiWho (approved):

JiahuiWho:
The guard is right, but there's a second reader `contractIDsForPreimage` (`contract_operations.go:213`) of the same data. Could you add the same check [there](https://github.com/stellar/wallet-backend/blob/main/internal/indexer/processors/contract_operations.go#L213) ? We could use a single helper that converts an `ScAddress` to a storable string, so future call sites won't miss it. 

The confirmed chain:

1. A successful `CreateContractV2` with a normal `G...` deployer, plus one unused source-account auth entry declaring a `create_contract` from `B...`, closes as `txSUCCESS`. Because an auth entry nobody consumes is silently ignored, host never looks at it.
2. Since the transaction is succeeded, `participants.go:182` doesn't skip its Soroban participants.
3. `participantsForAuthEntries` walks the auth tree, reaches that `create_contract`, and line 213 adds the `B...` address to the set.
4. `operations.go:425` encodes each participant before the `COPY`, and `AddressBytea.Value()` rejects this one because its payload is 33 bytes instead of 32. That aborts the ledger insert, and it fails the same way on every retry, so ingestion stops there.

JakeUrban:
Confirmed and fixed in cef88f6. Thanks — the participant path was a real gap, and the successful-transaction route is the part I'd missed: I'd checked the failed-transaction gate on Soroban participant collection and wrongly concluded that path was covered. An auth entry nobody consumes gets there on `txSUCCESS`, so the gate never applies.

I reproduced your chain before fixing it, and it turned out to be wider than the one call site. Three separate routes each put a `B...` address into the participants set:

| Route | Source |
| --- | --- |
| create-contract preimage deployer | the one you flagged |
| address credentials | `SorobanAddressCredentials.Address` |
| auth tree contract function | `ContractFn.ContractAddress` |

All three failed with `participant "BAAAAAICAMCAKBQHBAEQUCYMBUHA6EARCIJRIFIWC4MBSGQ3DQOR4H2TOM" must encode for the operation-address COPY`. There's a fourth of the same shape on the invoke-contract host function's own contract address, which I guarded too.

Took your suggestion on the shape of the fix. `storableAddressString` pairs the allowlist with the strkey conversion and returns an `ok` flag, and every site that turns an envelope address into a stored one now goes through it — the four above plus the state-change path, which no longer calls the allowlist and the conversion separately. One helper for a future call site to reach for.

One behavioural note worth a look: when a deployer address is unstorable, its address is dropped from the participants set but the contract ID derived from the preimage is still included. That contract can never exist, so arguably it shouldn't be a participant either — but that's the phantom-record question from #746 rather than the encoding wedge, so I left it as it was rather than widening this PR.

Verified: with the check disabled, 10 subtests fail across the state-change and participant tests; with it, the full indexer suite passes.

---

JiahuiWho:
I tested the other two call sites and they won't get this far. A claimable-balance address in `ContractFn.ContractAddress` or in `SorobanAddressCredentials.Address` traps during apply, because the host converts those two addresses eagerly and rejects the arm. The transaction then fails, and we drop Soroban participants for failed transactions.

Still nice to have the guard on all three. Just worth splitting the test comment, which currently explains all three as riding along on a successful transaction and that's only true for the `create_contract` case.

JakeUrban:
You're right, and my earlier claim was wrong. Corrected in 8cf8b76.

I checked this against the host rather than just taking the correction, and it confirms your mechanism exactly. In `soroban-env-host`, `AccountAuthorizationTracker::from_authorization_entry` converts the credentials address with `host.add_host_object(address_creds.address)`, and `AuthorizedFunction::from_xdr` does the same for `xdr_contract_fn.contract_address` — both at tracker-build time, before any matching. The create-contract arm is the odd one out: it keeps `contract_id_preimage` as XDR and matches with `host.compare(c1, c2)` on the raw structure, so the declared deployer is never converted. That asymmetry is exactly why only that route survives to `txSUCCESS`.

Worth being clear about where I went wrong, since it matters for how the earlier comment reads. My test fixture *sets* the transaction successful and then walks the participant code — so it only ever proved "if this reached here, it wedges". I reported that as real-world reachability for all three routes, which it wasn't evidence for. The create-contract route is reachable because you'd verified it; the other two I asserted on the strength of a fixture I'd built myself.

Same conclusion applies to a fourth site I guarded, the invoke-contract host function's own contract address — the host has to convert that to dispatch the call, so it's in the same unreachable bucket. I confirmed all three unreachable ones feed only participant collection (`participantsForSorobanOp` has a single caller, behind the success gate) and that no state-change processor reads auth entries, so there's no ungated path I'd missed. Only `contractIDsForPreimage` is load-bearing.

Kept the guards, per your call — they're ~2 lines each, no behaviour change for storable addresses, and they mean a shift in what the host converts eagerly can't silently reopen this. Split the test comment so it says which route is real and which fixtures are counterfactual, and the success-precondition assertion now only runs for the route that depends on it.

Also updated the PR description, which still described only the failed-transaction route and predated the participant fix entirely.

---

JiahuiWho:
Looks good to me!
Feel free to ship once the two Copilot items are in.

JakeUrban:
Narrowed to what the current host actually makes reachable, in 8494a50.

Dropped the guards on the credentials address, the auth tree's contract-function address, and the invoke-contract host function's contract address. The host converts all three eagerly — two while building auth trackers, one to dispatch the call — so a CAP-0067 arm in any of them traps at apply and the transaction fails, and Soroban participants are dropped for failed transactions. Nothing ever reached the encoder through them. Their tests are gone too; they only asserted against fixtures that could not exist.

What's left is the create-contract preimage deployer, and its two readers share the helper because they share a reason rather than a shape: the host keeps the preimage as XDR and compares it raw, so it never converts the declared deployer, which is precisely why that address survives to ingestion — on a failed transaction through the contract-deploy state changes, and on a successful one through the participants. Same hazard, same field, one guard.

Also collapsed `isStorableAccountAddress` and `storableAddressString` into a single `deployerAddressString`. Two functions for one pair of callers was abstraction without a second use to justify it, and the name now says what it is for instead of describing a general capability the codebase doesn't need. Its comment is grounded in what the host does today rather than what it might do later.

Net production diff is three files: one helper and two call sites. With the allowlist disabled, 7 subtests fail across both paths, so both guards are load-bearing. Full indexer suite, build and lint are clean.

Updated the PR description to match — the defence-in-depth framing is gone from it.

---

internal/indexer/processors/utils.go
```diff
@@ -294,3 +294,28 @@ func ConvertOperation(
...
+// therefore ignored, and a CAP-0067 claimable-balance or liquidity-pool address declared as the
+// deployer reaches ingestion — on a failed transaction via the contract-deploy state changes, and
+// on a successful one via the operation's participants. The allowlist mirrors what
+// AddressBytea.Value() accepts: a version byte plus a 32-byte account or contract key, with a
+// muxed (M...) payload reduced to its base account.
+func deployerAddressString(addr xdr.ScAddress) (string, bool, error) {
+	switch addr.Type {
+	case xdr.ScAddressTypeScAddressTypeAccount,
+		xdr.ScAddressTypeScAddressTypeContract,
+		xdr.ScAddressTypeScAddressTypeMuxedAccount:
```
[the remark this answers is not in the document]
JakeUrban: Confirmed and fixed in c7bc886.

I reproduced it before changing anything, using the route this PR already guards — an unconsumed auth entry naming a muxed deployer over the operation's own source account, on a `txSUCCESS` transaction:

```
COLLISION: [MAUE24B36YYY3CXTXNFE3IFXU6EE4NUOS5L744IWGTNXVXZAXFGMOAAAAAAAAAAAFKWIE
            GAUE24B36YYY3CXTXNFE3IFXU6EE4NUOS5L744IWGTNXVXZAXFGMP6CC]
encode to identical operations_accounts rows
```

`operations_accounts` is keyed on `(account_id, ledger_created_at, operation_id)`, so those are two rows on one key — a class 23 failure, which `isPermanentPersistError` treats as permanent, so the insert aborts without advancing the cursor and fails identically on retry. Same wedge, different door.

One correction: the precedent isn't `extractAddressFromScVal`, which calls `addr.String()` and doesn't normalize either. The established idiom is in `internal/services/sep41/events.go` from #693, and I followed that — `strkey.Encode(VersionByteAccountID, muxed.Ed25519[:])`, with the same reasoning that the multiplexing id is off-chain routing metadata rather than account identity. `AddressBytea`'s own doc comment already states ingestion is expected to normalize before storage, so this helper was the place that was violating the contract.

Verified: with normalization removed, `muxed/reducedToBase` and `muxedOverSource/dedupesWithSource` both fail.

---

internal/indexer/processors/contract_operations_test.go
```diff
@@ -589,3 +590,86 @@ func Test_participantsForSorobanOp_invokeHostFunction_invokeContract(t *testing.
...
+// not apply, and the auth tree is walked. A CAP-0067 claimable-balance address declared there
+// would otherwise reach the participants set and fail to encode, aborting the ledger insert.
+func Test_participants_unstorableAddressInAuthEntry(t *testing.T) {
+	const deployer = "GCQIH6MRLCJREVE76LVTKKEZXRIT6KSX7KU65HPDDBYFKFYHIYSJE57R"
+
+	var h xdr.Hash
+	for i := range h {
+		h[i] = byte(i)
+	}
+	cbAddr := makeScClaimableBalance(h)
```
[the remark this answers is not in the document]
JakeUrban: Right, and fixed in c7bc886. The encode-only assertion was vacuous for liquidity pools exactly as you describe — a 32-byte payload encodes fine, so the test passed whether or not the filter dropped it.

The test now asserts the exact participant set, with cases for claimable balance and liquidity pool (both dropped), a plain account (kept), and a muxed deployer (reduced to its base account). It also checks that no two members of the set encode to the same row, which is what caught the muxed duplicate in the sibling thread.

Confirmed it now bites: with the filter disabled, `liquidityPool/dropped` fails alongside `claimableBalance/dropped`, where previously only the claimable-balance case would have.

The state-change tests were already asserting absence rather than encodability (`assert.Empty` on the emitted changes, for both address kinds), so the gap was specific to the participant test.

---

internal/indexer/processors/utils.go
```diff
@@ -294,3 +294,38 @@ func ConvertOperation(
...
+// deployerAddressString converts a create-contract preimage's deployer address to its strkey
+// form, reporting ok=false when it is a kind types.AddressBytea cannot store.
+//
+// The host keeps a create-contract auth invocation's ContractIdPreimage as XDR and compares it
+// raw, so it never converts the declared deployer address. An auth entry nobody consumes is
+// therefore ignored, and a CAP-0067 claimable-balance or liquidity-pool address declared as the
+// deployer reaches ingestion — on a failed transaction via the contract-deploy state changes, and
+// on a successful one via the operation's participants. The allowlist mirrors what
+// AddressBytea.Value() accepts: a version byte plus a 32-byte account or contract key, with a
+// muxed (M...) payload reduced to its base account.
```
JakeUrban: Remove this.

JakeUrban: Removed in 8d3679c — the doc comment is now just the two lines saying what the function reports.

---

